PR #8, opened 2016-07-20
Update directory structure

greg-hellings:
Created configure and provision directories
Moved existing Ansible work into provision directory
Created base files for configure directory
Updated README.md to fix some formatting problems
Updated README.md to reflect new structure


arilivigni:
1 - You probably want to rebase since master has changed
2 - library folder inside provision should be in the main linch-pin directory
3 - all documentation should be updated to reflect the changes not just README files

greg-hellings:
1. Done
2. Disagree. The library folder parallel to the top level playbook is automatically added to Ansible's search paths. There is very little likelihood of overlap between custom Ansible modules/plugins between provisioning and configuration, and if we do come across some in the future that need to be shared between them, we can create a top-level folder at that time and include it via Ansible configurations at that time.
3. Why do we have an output of the "tree" command inside of the docs? That seems both excessive, as the project grows, and incredibly tedious to update for something that could just be gathered by the user when they mutter "tree" at the command prompt.

samvarankashyap:
1. Regarding ansible module path: Unless we set the $ANSIBLE_LIBRARY environment variable , or library in ansible.cfg modules are not resolved.
2. Regarding tree command output it will give us a snapshot of whole directory structure and there are some comments explaining what purpose of the folders. We can cut down the depth of the tree if its too detailed .

greg-hellings:
1. Not anymore
2. The list of subfolders shouldn't need to go deeper than the top level. I'm going to update it so that is the case. Otherwise, we need to update that file with every single new file added or removed from the repository, which is not feasible.

jlflynn:
Greg, what do you mean by not anymore.  What is the alternative?

1. IMHO, users should be able to run the code in this repo without the requirement to set special variables in their environment or ansible config.  It should run out of the box.
2. There is a desire to keep the tree in the docs because it is in line with other projects.  SK, do  you feel the full depth of the tree should be maintained?  If so, in the interest of collaboration, we can can keep the tree updated.   I don't think it is up-to-date with the most recent changes, though.

arilivigni:
1. What does that mean?  sk tested this and it didn't work with what you suggested.  We are keeping the library in the top level directory and configure and provision are subdirs
2. We will keep the depth the way it is currently and not go deeper, but it should be updated and accurate in both the README and the docs.

greg-hellings:
1. I fixed the problem, it now works without any updates to configuration or environment variables that were previously referenced in the docs. The library directory had been improperly structured. Testing on my system, it works (after my changes) with the library alongside the site.yml file, which is what the Ansible docs state should happen. Please have SK test again with my branch without setting environment variables or updating his ansible.cfg file.

2. I pulled it up to one level in the docs. I just pushed another update that pulls it up to a single level in the README.

arilivigni:
1. But why does is need to be in the provisioner directory.  It makes more sense to the rest of us to be in the root directory of the project to have all libraries there and is that way in the document I wrote.  I see no concrete reason to change that.
2. ok I guess I don't know why we can't follow what we have.  You are the only one that wants something different.

greg-hellings:
1. If it's in the provisioner directory and another is placed in the configure directory parallel to its site.yml file when the need arises, then the user doesn't have to set environment variables or config file options. It just works. It's how Ansible documentation tells you to place custom modules with your project. "The directory i:file:./library, alongside your top level playbooks, is also automatically added as a search directory." If there are modules that are not going to be shared between provisioning and configuring, then there isn't a reason to place those into a library folder. If we find they do, in the future, then we can provide a shared location as well. This mechanism is well understood and well documented by Ansible's behavior and simplifies the needs of our users by providing a system that works without the extra configuration.

2. The data lives in two places in the docs and was already out of date and inconsistent between them. I didn't change what was there, I just limited it to a depth of 1 and updated the two locations. They are, in fact, still out of sync - one has theoretical locations that don't yet exist, while the other has only locations that do. Why would we want to have two places that maintain nothing more than the information that could be garnered by a single command on the developer's system? Sure, having a description of each top level folder makes sense, to begin directing people's attention. But trying to maintain a full inventory inside of the documentation of the entire tree of the project is just an open invitation for data to become out of date and a top notch location to create merge conflicts unnecessarily.

arilivigni:
1. Why make the change later we can just have them stored in one location now?  You are making a change that isn't necessary.  You contradict yourself by saying there are not going to be shared modules but if they are then we will change it.

greg-hellings:
1. I didn't say to change it then. I said we can add a shared location if it arises. But at this time, we have no shared modules that cross between the two Ansible directories and no concrete plans to add them.

Let me put this another way: why do you want to make things harder for our users? Your proposed way requires the user to either create an environment variable or create an Ansible config file that gets referenced by an environment variable. The method in this change simply works without any of that. All I did was drop the entire set of Ansible structures (library, group_vars, hosts, site.yml, roles) into the provisioner folder and restructure the library folder into the proper form.

PR #13, opened 2016-07-22
Restructure the library folder

greg-hellings:
By default, Ansible adds any library folder that is alongside
the top level site.yml file to the library search path. By
restructuring the library folder to be the proper shape for
Ansible to search, we can remove the need to set the search
path in ansible.cfg or environment variables. This version
of the structure should work stright out of the box.


samvarankashyap:
Will be trying out the changes you have made if everything goes as expected, we will be merging it.

greg-hellings:
In order to properly test, you'll need to remove the module search path changes in your ansible.cfg file (the ones that are mentioned in the README.md lines I deleted in this PR). All this change does is structure the library folder the way that Ansible expects it to look. For whatever reason it doesn't seem to like having subfolders in the local module path.

samvarankashyap:
Agreed , its not recognizing the sub folders in local library folder. 
Might be a bug in Ansible.
Merging your pull request.

PR #17, opened 2016-07-27
Restructure into provision folder

greg-hellings:
Move provisioning code into folder named provision
Update documentation to reflect new layout


herlo:
I'm trying to understand the logic of this move. Is it so there can be clearly defined lines of provisioning, teardown, etc?

jlflynn:
This change is proposed to match the directory structure laid out in the architecture doc.  This allows for clear separation of the roles for provisioning and configuration. This was discussed last week and it was agreed that this restructuring should be done.  The configure directory has already been created in a separate PR. This PR completes the agreed upon restructuring.

greg-hellings:
As Jaime noted, this is just establishing the prior agreed-to structure for the folder layout. It establishes a parallel structure between the provision and configure components to the project, as each is going to be its own set of Ansible playbook, roles, and supporting files.

PR #18, opened 2016-07-27
if resource_group_vars are not defined, it's okay to ignore them

herlo:
I think this is a definite improvement over the requirement of an empty resource_group_vars. What say you?


roles/common/tasks/main.yml
```diff
@@ -26,12 +26,18 @@
...
 
 - name: "filtering the output for duffy resource groups"
-  set_fact: 
+  set_fact:
     duffy_res_grps: "{{ output.output.data.resource_groups | selectattr('res_group_type', 'equalto', 'duffy') | list }}"
 
+#- name: "DEBUG:: checking registered output of os_res_grps"
+#  debug:
+#    msg: "{{ item }}"
+#  with_items: "{{ output.output.data }}"
```
greg-hellings: These commented lines shouldn't land as-is.

herlo: I don't disagree. However, these kinds of artifacts already existed in the code. It is a good idea to clean them up.

greg-hellings: These are new ones, though. So let's at least not add them in during this change.

PR #29, opened 2016-08-04
Revert "Added a port of Jenkins slave configuration from Taskrunner to Ansible"

arilivigni:
Reverts CentOS-PaaS-SIG/linch-pin#27


robled:
@arilivigni - Can you comment on the specific lines in this pull request that are problematic?  I know that some tasks such as installing particular packages might not be suitable for all users, and I propose that we break those tasks out into separate roles which are optional and broken out from the Jenkins slave task.

arilivigni:
@robled We already have a Jenkins slave playbook that we want to use which I told you a bit ago.  I will add that by next week as I indicated to @jlflynn.  You also didn't cover repos and pkgs thoroughly for Fedora and Centos.

PR #43, opened 2016-08-12
custom module fix for async_status module

samvarankashyap:
custom fix for ansible async_status module , until there is community bug fix for os_server module .

### DIFF
--- a/library/async_status.py
+++ b/library/async_status.py
@@ -1,0 +1,111 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2012, Michael DeHaan <michael.dehaan@gmail.com>, and others
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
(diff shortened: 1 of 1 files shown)


herlo:
So, you are taking the async_status module directly from ansible, and modifying it for use with linch-pin?

samvarankashyap:
Any module in ansible when used with async parameter will create file with jobid as name in ~/.ansible_async/jobid ( for example : ~/.ansible_async/27762078897.12739 )
Ansible creates a thread in background which updates the respective job file when task is finished.
Moreover, In order to get the status of the task and its output , Ansible provides a module named asyn_status which essentially checkout the file and loads it using json.loads 
Ideally the jobfile created should contain only json object . However , os_server module when used with async parameter writes some extra log statements including the actual output which is a actual bug in os_server module. please see [https://github.com/ansible/ansible-modules-core/issues/4393](url)
Due to the extra log statements, json.loads in async_status fails . 
### Changes made in current PR: 
In order to bypass the error , and extract required json object from job file the following changes 
 are made at lines : 60 and 85-87 in async_status module. 
Such that , when async_status module passed with extra argument module_name: "os_server"
instead of failing at json loads it extracts the required json object eliminating extra data written to job file.

herlo:
With that explanation, I'm happy to merge this request.

PR #46, opened 2016-08-12
Update duffy_inventory.py to ensure every line ends properly

herlo:
Add a new line to strings that might not have them to ensure the inventory prints out properly.


arilivigni:
LGTM.  I don't know it to the level of detail you guys do, but makes sense

PR #47, opened 2016-08-16
Cleanup of documentation, linchpin_config.yml

herlo:
Update to README.md with values now being used
Modified linchpin_config.yml to remove reference to inventory_layout_file
Delete output and inventory directories before recreating them to ensure files don't get reused.


provision/hosts
```diff
@@ -1,2 +0,0 @@
-localhost ansible_connection=local
```
arilivigni: Are you sure this isn't needed any longer?

herlo: It works without this value on slave01. I suspect we could leave it, however. Let me update that in my commits.

herlo:
Replaced the provision/hosts file as requested. Merging with comment from @arilivigni in irc

15:06 <@ari> herlo: I only make that comment because it was added back to support the way we provision but maybe it is no longer needed
15:07 < herlo> ari: yeah. I don't know if it will cause problems elsewhere. You made a good point about it.
15:07 <@ari> other than that LGTM

README.md
```diff
@@ -161,20 +188,71 @@ For GCE the absolute path of the service account json file should be provided
...
 
-### De-Provision a topology
 ```
-command: ansible-playbook -vvv provision/site.yml -e "{'data':'path_to_topolgy_file', 'schema':'path_to_schema_file', 'state':'absent', 'topology_output_file':'path_to_output_file'}"
+## CLI Examples
+### Provision a topology
 
-example: ansible-playbook -vvv provision/site.yml -e "{'data':'../ex_topo/ex_data.yml', 'schema':'../ex_schemas/schema_v2.json', 'state':'absent', 'topology_output_file':'/tmp/ex_data_output.yaml'}"
 ```
+$ ansible-playbook -vvv provision/site.yml -e "topology='/path/to/topology_file'" \
```
samvarankashyap: @herlo  : Why are there multiple -e args instead of single -e with multiple variables inside it ?

greg-hellings: @samvarankashyap This syntax is simpler for people to read an understand than trying to write and escape JSON on the command line.

herlo: For documentation, it's clearer to understand from a user perspective. I've been doing a bit of cleaning up to make things more simplistic to use. I'm hoping the update to the documentation has helped others understand how the provisioning works.

dygrys:
Hi I would like to create masurian version of Firefox, how can I do it?

PR #48, opened 2016-08-19
added async_provisioning for os_server resources

samvarankashyap:
Changes made in current pull request: 

1. Renamed async_status to async_status_custom to avoid conflicts with the Ansible core module
2. Made changes to provision_os_server to save the asynchronous job details into a list
3. Made changes to output writer to do following tasks:
    - wait and poll on all job ids of different groups until they are finished based on
       their resource group &  type. 
    - collect the output of each job and write it to output file.

> Note
planning to organize outputwriter role into multiple playbooks once asynchronous provisioning of all types of resources are implemented.


herlo:
LGTM. I only recommend we use teardown as terminology vs deprovision. Cleaner and more consistent with other tools we use.

PR #51, opened 2016-08-20
added feature to provision rackspace server 

samvarankashyap:
added async provisioning for os_object
added feature to provision rackspace server supports only async== false


herlo:
LTGM. Thanks sk!

PR #52, opened 2016-08-22
make_host_vars should assign proper variable

herlo:
This fixes host_vars getting assigned the same value across all hosts. Now the hosts_vars get assigned properly.


samvarankashyap:
LGTM Thanks.

PR #53, opened 2016-08-22
added async provisioning to os_volume resource

samvarankashyap:
os_volume module raised the same bug which was raised by os_server module ,
Moreover , it can be fixed with the async_status_custom module.
Added Async provisioning  for openstack heat templates 
fixed a bug with async teardown of os_server module

### DIFF
--- a/library/async_status_custom.py
+++ b/library/async_status_custom.py
@@ -57,7 +57,7 @@ def main():
     module = AnsibleModule(argument_spec=dict(
         jid=dict(required=True),
         mode=dict(default='status', choices=['status','cleanup']),
-        module_name=dict(default='', choices=['os_server','']),
+        module_name=dict(default='', choices=['os_server','os_volume','']),
     ))
 
     mode = module.params['mode']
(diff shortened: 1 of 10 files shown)


herlo:
Can we move this block (write inventory file from outputs - https://github.com/samvarankashyap/linch-pin/blob/2532223229951628c9e053f1be09828d7d0e0c89/provision/roles/output_writer/tasks/main.yml#L55) after generating async output data? What impact does this have?

herlo:
When updating the topology_outputs fact, it probably should to be updated in two files; roles/output_writer/tasks/main.yml and invfilter.yml (used for testing). Might be good to make a variable structure of some sort like I've done in roles/output_writer/tasks/duffy_inventory.yml.

samvarankashyap:
Yes , we can move down the block (write inventory file from outputs - https://github.com/samvarankashyap/linch-pin/blob/2532223229951628c9e053f1be09828d7d0e0c89/provision/roles/output_writer/tasks/main.yml#L55) 
However, the async output generated would be different from the output we get from async == false . 
More we might have to write a new filter to handle the output of async == true.
I will make a commit to move it down before I merge this PR

PR #55, opened 2016-08-24
async provisioning of resources and schema_v3 update

samvarankashyap:
added async provisioning to aws_cfn resources 
added missing files of aws_cfn async provisioning
added async provisioning to aws_cfn resources
added missing files of aws_cfn async provisioning
added gcloud_gce async provisioning
added output formatting for gcloud_gce
added schema_v3 for strict resource_grp validations
fixed resource_group_name conflict

### DIFF
--- a/ex_schemas/schema_v3.json
+++ b/ex_schemas/schema_v3.json
@@ -1,0 +1,496 @@
+{  "description": "Infra Topology for openstack",
+    "type": "object",
+    "properties": {
+        "topology_name": {
+            "description": "The unique identifier for topology",
+            "type": "string"
+        },
+        "date": {
+            "description": "Date of creation",
+            "type": "string"
+        },
+        "site": {
+            "description": "site of provisioning ",
+            "type": "string"
+        },
+        "credentials": {
+             "description": "contains various credential_vaults names",
+             "type": "array",
+             "items":{
+                "type":"string"
+             },
+             "minimum":1
+        },
+        "resource_groups": {
+             "description": "contains list of resources ",
+             "type": "array",
+             "items":{
+                       "oneOf" :      
+                                 [ 
+                                   {"$ref": "#/definitions/openstack"},
+                                   {"$ref": "#/definitions/aws"},
+                                   {"$ref": "#/definitions/gcloud"},
+                                   {"$ref": "#/definitions/rackspace"},
+                                   {"$ref": "#/definitions/duffy_grp"},
+                                   {"$ref": "#/definitions/libvirt_grp"}
+                                 ]
+                }
+        }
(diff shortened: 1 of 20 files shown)


provision/roles/output_writer/tasks/main.yml
```diff
@@ -85,6 +88,8 @@
       os_volume_res: "{{ async_outputs_os_volume }}"
       os_heat_res: "{{ async_outputs_os_heat }}"
       aws_ec2_res: "{{ async_outputs_aws_ec2 }}"
+      aws_cfn_res: "{{ async_outputs_aws_cfn }}"
```
herlo: Should these vars be added to invfilter.yml?

ex_schemas/schema_v3.json
```diff
@@ -0,0 +1,496 @@
...
+                        "description":"contains creds file associated to this resource",
+                        "type":"string"
+                    }
+                 },
+                 "required":["resource_group_name","res_group_type","res_defs","assoc_creds"],
+                 "additionalProperties": true
+         }
+    }
+}
+
```
herlo: How does this differ from schema_v2? I'd like to start using v3 for my stuff, is that likely going to work? It's hard to tell without a diff.

Additionally, having a default schema in the config seems like a good idea, unless it's been overridden.

herlo:
Made a couple of comments, otherwise it looks good to me.

Thanks!

provision/roles/output_writer/tasks/main.yml
samvarankashyap: Yes we need to add them.
Moreover , the async == true output is different from the async == false output . So have to write a custom filter to parse it.

ex_schemas/schema_v3.json
samvarankashyap: schema_v3 provides a strict division of resource group functionality , 
One bug which i found in schema_v2 was , when we declare a resource_group_type  = "openstack", it actually allows aws resources also to be defined inside that openstack group which kind of gives us a flexibility, but i felt inherently by having a resource_group_type its wrong. 
so if want to have a combination of multiple cloud resources , we can have a a new resource group type known as "hybrid" which accepts all the resource types.

samvarankashyap: having a default schema its a good idea , will add it in next commit.

provision/roles/output_writer/tasks/main.yml
herlo: Will you add a commit for this?

samvarankashyap: sure !

PR #57, opened 2016-09-06
added inventory gen role and multiple configurations options

samvarankashyap:
added inventory gen role and multiple configurations options for layout outputs and inventory files


provision/site.yml
```diff
@@ -19,3 +19,8 @@
   hosts: localhost
   roles:
     - { role: 'output_writer'}
+
+- name: "inventory generation role"
+  hosts: localhost
+  roles:
+    - { role: 'inventory_gen'}
```
herlo: Should this be also when state == "present"? Essentially, I didn't see anything that indicated the inventory_gen role should NOT run when state == "absent".

linchpin_config.yml
```diff
@@ -2,11 +2,13 @@
...
 inventory_playbooks:
   - duffy_inventory.yml
-
 async: false
 async_timeout : 1000
 no_output : false
 schema: "ex_schemas/schema_v2.json"
+inventory_layouts_path: "inventory_layouts"
+inventory_outputs_path: "inventory_outputs"
+inventory_type: "aws"
```
herlo: What if you want multiple inventory_type values? Say I want to generate inventories for both aws and openstack, how would one do this?

inventory_layouts/duffy_3node_inventory_layout.yml
```diff
@@ -0,0 +1,41 @@
...
+        ansible_sudo: False
+        ansible_ssh_user: root
+        openshift_override_hostname_check: True
+        openshift_set_hostname: True
+      children:
+        - masters
+        - nodes
+    masters: ""
+    nodes: ""
+    repo_host: ""
```
herlo: This file is poorly named. I realized this recently. It should be named openshift-3node-cluster.yml or something similar. The goal of this layout is not to spin up duffy instances, but to generate inventory file(s) that can be used to create/run an openshift cluster.

provision/filter_plugins/aws_inventory.py
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env python
+
+from ansible import errors
+try:
+    from configparser import ConfigParser
+except ImportError:
+    from ConfigParser import ConfigParser
+import StringIO
+import pprint 
+def pop_items(item_list,count):
```
herlo: Put a space here to delineate imports from code?

provision/filter_plugins/aws_inventory.py
```diff
@@ -0,0 +1,110 @@
...
+                   host_string += " "+ var +"="+item+" "
+            config.set(group,host_string)
+    return config
+    
+def aws_inventory(topo, layout):
+    inventory = ConfigParser(allow_no_value=True)
+    no_of_groups = len(topo['aws_ec2_res'])
+    layout_hosts = get_layout_hosts(layout)
+    inven_hosts = get_host_ips(topo)
+    if len(inven_hosts) != len(layout_hosts):
```
herlo: I'm not sure this is a problem you should worry about solving. I only checked to see if there were *enough* hosts spun up to cover the inventory needed. However, it seems that the best idea here is to move this logic to a generic inventory filter check. Something that validates the number of hosts to cover the number of hosts needed in the inv_layout. Thoughts?

provision/filter_plugins/gcloud_inventory.py
```diff
@@ -0,0 +1,110 @@
...
+                   host_string += " "+ var +"="+item+" "
+            config.set(group,host_string)
+    return config
+    
+def gcloud_inventory(topo, layout):
+    inventory = ConfigParser(allow_no_value=True)
+    no_of_groups = len(topo['gcloud_gce_res'])
+    layout_hosts = get_layout_hosts(layout)
+    inven_hosts = get_host_ips(topo)
+    if len(inven_hosts) != len(layout_hosts):
```
herlo: See https://github.com/CentOS-PaaS-SIG/linch-pin/pull/57/files#r77861369

provision/filter_plugins/openstack_inventory.py
```diff
@@ -0,0 +1,108 @@
...
+                   host_string += " "+ var +"="+item+" "
+            config.set(group,host_string)
+    return config
+    
+def openstack_inventory(topo, layout):
+    inventory = ConfigParser(allow_no_value=True)
+    no_of_groups = len(topo['os_server_res'])
+    layout_hosts = get_layout_hosts(layout)
+    inven_hosts = get_host_ips(topo)
+    if len(inven_hosts) != len(layout_hosts):
```
herlo: See https://github.com/CentOS-PaaS-SIG/linch-pin/pull/57/files#r77861369

provision/roles/inventory_gen/tasks/main.yml
```diff
@@ -0,0 +1,63 @@
...
+  template:
+    src: "templates/openstack_inventory_formatter.j2"
+    dest: "{{ inventory_output_file }}"
+  when: inventory_type == 'openstack'
+
+- name: "generate duffy_inventory"
+  template:
+    src: "templates/duffy_inventory_formatter.j2"
+    dest: "{{ inventory_output_file }}"
+  when: inventory_type == 'duffy'
```
herlo: I kind of dislike the approach here. Specifically, this could be single loop that goes through a list of inventories to generate, then includes the appropriate inventory generation playbook. You could also just do a loop here with the template task for each type of inventory. Generating just ONE inventory seems limiting, no?

provision/filter_plugins/gcloud_inventory.py
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env python
+
+from ansible import errors
+try:
+    from configparser import ConfigParser
+except ImportError:
+    from ConfigParser import ConfigParser
+import StringIO
+
+def pop_items(item_list,count):
```
herlo: I don't see what pop_items() does? Also, I am noting common code in each inventory layout filter. Maybe a library is in order here?

herlo:
Good work on this SK! Overall, I love the direction here. A new module was the right choice for sure. My only real complaint so far is that we're limiting to JUST ONE inventory layout, at least from what I could see.

herlo:
I like the way you used ConfigParser for inventory generation. I'm planning on updating the duffy inventory filter to do it this way.

provision/site.yml
samvarankashyap: Sure , I will update it .

linchpin_config.yml
samvarankashyap: I agree with you it doesnot work with multiple inventories , 
In that case we might have to loop it thorough a list of inventory types. 
Will update it in next commit .

inventory_layouts/duffy_3node_inventory_layout.yml
samvarankashyap: will rename it to openshift-3node-cluster.yml

provision/filter_plugins/aws_inventory.py
samvarankashyap: shall be updated.

provision/filter_plugins/aws_inventory.py
samvarankashyap: guess we it might have to be
len(inven_hosts) >= len(layout_hosts) 
because , inventory_host can be greater in number .

provision/filter_plugins/gcloud_inventory.py
samvarankashyap: agreed . pop_items needs to be deleted . Yes , There is common functions , i was skeptical about the imports inside a filter file . Will try to make a simple library file from which we can import .

linchpin_config.yml
herlo: This was the reason I had the item based loop in the output_writer role, which you removed.

- include: "{{ item }}"
  with_items:
    - "{{ inventory_playbooks }}"
  when: state == "present" and inventory_layout is defined

provision/filter_plugins/aws_inventory.py
herlo: Initially, the idea was that if we had >1 clouds involved, we wouldn't track the hosts vs inventory nodes here. Instead, we'd track it in a generic place. Essentially, running through the inventory layout until we run out of items is likely going to be the way forward. We only care that there are enough hosts to handle whatever is in the layout. 

For duffy, this was simple, since it'll never really be used with any of the other cloud providers. However, consider this scenario:

Let's you have say 3 openstack nodes, 1 aws node, and 1 gce node. 
For the inventory layout, you have a need of four (4) nodes to generate an inventory.

Based upon this, you'd want to have a loop that goes through the nodes in some order. I'd suspect based upon a variable (like I have above 'inventory_playbooks' loop), which then would process the inventories based upon the order of the list. If the list were as follows:

gce, openstack, aws

And get inventory output files like:

inventory/gce.inventory (representing one node)
inventory/openstack.inventory (representing 3 nodes)

Even though linch-pin spins up 5 nodes, the aws node is not used in the inventory going forward.

This is completely acceptable. However, if you had a need for 6 nodes, linch-pin should at least alert you (or fail, depending on what we decide here) that you don't have enough nodes to generate all of the inventories correctly.

Thoughts?

samvarankashyap: I completely agree with the above idea. 
I have concern with multiple inventories being generated being generated in the aforementioned 5 node scenario, 
Given the inventory ultimately consists of just ipaddresses/hostnames/dns and the vars why would there be two inventories , instead of single inventory  
inventory/topology_name.inventory

provision/roles/inventory_gen/tasks/main.yml
samvarankashyap: Agreed on will update the code accordingly.

provision/filter_plugins/aws_inventory.py
herlo: With ansible, you can pass a directory as an inventory (eg. ansible-playbook -i /path/to/inventories/ site.yml'). 

The idea here is that for each different cloud, a different inventory file is generated. If you don't do that, you would have to append / manage the previous inventory file. The example inventory layout only has a few items inside, this is true, however, a much more complex example may exist. Managing multiple inventory files seems simpler (considering the directory bit above) than trying to generate one global inventory file.

provision/roles/inventory_gen/tasks/main.yml
```diff
@@ -0,0 +1,54 @@
...
+
+# Generating inventory based on inventory type 
+# Note: inventory type is configured in linchpin_config.yml can be overridden extravar inventory_type
+- name: "Generate inventories"
+  template:
+    src: "templates/{{ item.0 }}_inventory_formatter.j2"
+    dest: "{{ item.1 }}.{{ item.0 }}"
+  with_nested:
+    - "{{ inventory_types }}"
+    - ["{{ inventory_output_file }}"]
```
herlo: Do we want this to be similar to line 43 above? Essentially, this would be blank if 'inventory_output_file' is not supplied, correct?

samvarankashyap: @herlo : if the inventory_outputfile is not supplied it defaults to topologyname.inventory mentioned 
according to 43

herlo: Oh, I see. I just misread that one. Thanks!

provision/roles/inventory_gen/tasks/main.yml
```diff
@@ -33,9 +33,9 @@
     msg: "{{ inventory_layout }}"
   when: state == "present" and inventory_layout is defined
 
-# include inventory generation files here : can be overridden by  extravar inventory_layout_file currently defaulted to duffy_3node_inventory_layout.yml
+# include inventory generation files here : can be overridden by  extravar inventory_layout_file currently defaulted to openshift-3node-cluster.yml
 - name: "include the layout file for inventory generation"
-  include_vars: "{{ inventory_layout_file | default( inventory_layouts_path+'/'+'duffy_3node_inventory_layout.yml') }}"
+  include_vars: "{{ inventory_layout_file | default( inventory_layouts_path+'/'+'openshift-3node-cluster.yml') }}"
```
herlo: I'm confused by this change. Why would we hardcode the inventory_layout_file? It's either set at runtime with --extra-vars, or provided in a config file (like linchpin_config.yml), right?

samvarankashyap: @herlo : As we are limited to only one layout now , if the layout file is not found we would be considering openshift-3node-cluster.yml for now . 
In later commit it can be removed , and we can have a layout that could just return only the ips/hostnames for the inventory.

PR #62, opened 2016-09-13
added generic inventory filter minor fixes to other inventory files 

samvarankashyap:


provision/InventoryFilters/GCloudInventory.py
```diff
@@ -19,7 +19,6 @@ def get_host_ips(self, topo):
 
     def get_inventory(self, topo, layout):
         inventory = ConfigParser(allow_no_value=True)
-        layout_hosts = self.get_layout_hosts(layout)
```
samvarankashyap: removed unused variable

provision/InventoryFilters/GenericInventory.py
```diff
@@ -0,0 +1,68 @@
...
+        # set children for each host group
+        inventory = self.set_children(inventory, layout)
+        # set vars for each host group
+        inventory = self.set_vars(inventory, layout)
+        # add ip addresses to each host
+        inventory = self.add_ips_to_groups(inventory, inven_hosts, layout)
+        inventory = self.add_common_vars(inventory, host_groups, layout)
+        output = StringIO.StringIO()
+        inventory.write(output)
+        return output.getvalue()
```
herlo: This looks similar to what you've done before. Were you not planning on making the config a class member of InventoryFilter? I'm okay with that being a future change if that's what you want, or we can discuss those options further.

samvarankashyap: @herlo : I will be changing in config obj as member of InventoryFilter in following commits of current PR.

herlo: @samvarankashyap okay, sounds good. I was not expecting more commits to come in on this PR. I'll wait for more commits. :)

library/schema_check.py
```diff
@@ -79,9 +79,9 @@ def check_file_paths(module, *args):
 
 def validate_grp_names(data):
     res_grps = data['resource_groups']
-    res_grp_vars = data['resource_group_vars']
+    res_grp_vars = data['resource_group_vars'] if 'resource_group_vars' in data.keys() else []
```
herlo: I'd highly recommend list comprehensions here. It's cleaner.

http://stackoverflow.com/questions/4406389/if-else-in-a-list-comprehension

library/schema_check.py
```diff
@@ -79,9 +79,9 @@ def check_file_paths(module, *args):
...
     res_grp_names = [ x['resource_group_name'] for x in res_grps ]
-    res_grp_vars = [ x['resource_group_name'] for x in res_grp_vars ]
+    res_grp_vars = [ x['resource_group_name'] for x in res_grp_vars ] if len(res_grp_vars) > 0 else []
```
herlo: I'd highly recommend list comprehensions here. It's cleaner.

http://stackoverflow.com/questions/4406389/if-else-in-a-list-comprehension

provision/InventoryFilters/GenericInventory.py
```diff
@@ -0,0 +1,67 @@
...
+from AWSInventory import AWSInventory
+from OpenstackInventory import OpenstackInventory
+from GCloudInventory import GCloudInventory
+from InventoryFilter import InventoryFilter
+
+
+class GenericInventory(InventoryFilter):
+    def __init__(self):
+        InventoryFilter.__init__(self)
+        self.filter_classes = {
```
herlo: I continue to question this list of classes. Could it be pulled from the linchpin_config.yml or elsewhere? I'd like to see this be more dynamic. Or possibly, it could be pulled from the classes that have <something>Inventory.py in the python path?

I have code that can do the instantiation of the classes from a config file or the like in a pythonic way, so that shouldn't be too troublesome to implement.

library/schema_check.py
samvarankashyap: @herlo In the above scenario , i am using if else statement as a ternery operator ,
http://book.pythontips.com/en/latest/ternary_operators.html
res_grp_vars are defaulted to empty list , if not mentioned.
For readability it can be changed to as follows : 
```
if len(res_grp_vars) > 0:
    # essentially all dicts of res_grp_vars into a list
    res_grp_vars = [ x['resource_group_name'] for x in res_grp_vars ]
else:
    res_grp_vars = []
```

provision/InventoryFilters/GenericInventory.py
samvarankashyap: Agreed , 
Whats you call on design which is similar to as follows : 
https://github.com/apache/libcloud/blob/trunk/libcloud/compute/providers.py

library/schema_check.py
herlo: res_grp_vars = [ x['resource_group_name'] if len(res_grp_vars) > 0 else [] for x in res_grp_vars] 

The above seems like it would handle this, no?

samvarankashyap: no it creates a nested list when res_grp_vars are empty 
when len(res_grp_vars) == 0 it results in following list 
[ [] ]

provision/InventoryFilters/GenericInventory.py
herlo: I'm good with this concept.

provision/InventoryFilters/InventoryFilter.py
```diff
@@ -10,70 +10,86 @@
...
+    
+    @abc.abstractmethod
+    def get_inventory(self, topo, layout):
+        pass
 
     def get_layout_hosts(self, inv):
-        return inv['host_groups'].keys()
+        count = 0
+        for host_group in inv['hosts']:
+            count += inv['hosts'][host_group]['count'] if 'count' in inv['hosts'][host_group] else 1
```
herlo: This could use a list comprehension here.

provision/InventoryFilters/InventoryFilter.py
```diff
@@ -10,70 +10,86 @@
...
-    def add_ips_to_groups(self, config, inven_hosts, layout):
+    def add_ips_to_groups(self, inven_hosts, layout):
         # create a ip to host mapping based on count
         ip_to_host = {}
+        host_grps = self.get_layout_host_groups(layout)
+        
         for host_name in layout['hosts']:
-            count = layout['hosts'][host_name]['count']
+            #if type(layout['hosts'][host_name]) is dict:
+            count = layout['hosts'][host_name]['count'] if 'count' in layout['hosts'][host_name] else 1
```
herlo: list comprehension?

PR #65, opened 2016-09-28
units tests and bugfixes

samvarankashyap:


provision/filter_plugins/aws_inv.py
```diff
@@ -8,7 +8,9 @@
     from configparser import ConfigParser
 except ImportError:
     from ConfigParser import ConfigParser
-sys.path.append(os.getcwd())
+filepath = os.path.realpath(__file__)
```
herlo: I still do not like this concept. Can we work on maybe putting in an easy_install (.pth) file or something? Eventually, I'd like to see an alternative way of doing this, but even in #ansible, they suggest something similar. 

Thinking that maybe packaging up both the modules and the filter classes into a package might work? Thoughts?

herlo:
Instead of having merge commits, you can do the following:

git stash
git pull --rebase upstream master
git pull --rebase origin master
git push origin master
git stash pop

I don't generally have any merge commits when I perform these tasks. I hope this helps.

tests/TestLinchPinCredentials.py
```diff
@@ -0,0 +1,152 @@
+import os
```
herlo: https://www.python.org/dev/peps/pep-0008/#package-and-module-names implies that the filenames here need to be test_linchpin_credentials.py. Can you fix this please?

tests/TestLinchPinInvocation.py
```diff
@@ -0,0 +1,56 @@
+import os
```
herlo: https://www.python.org/dev/peps/pep-0008/#package-and-module-names

tests/filter_plugins
```diff
@@ -0,0 +1 @@
+../provision/filter_plugins
```
herlo: Could we not add this to the sys.path in the test setUp method? I'm not clear the value of this symlink.

tests/TestLinchPinCredentials.py
```diff
@@ -12,65 +12,8 @@
 from nose.tools import raises
 from nose import with_setup 
 from collections import namedtuple
+from linchpin_utils import cred_utils
```
herlo: I like this idea. Makes sense to have this in general. :+1:

outputs/all_stack_outputs.testoutput
```diff
@@ -0,0 +1,935 @@
...
+                protocol: null
+                remote_ip_prefix: null
+                security_group_id: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
+        status: ACTIVE
+        tenant_id: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
+        updated: 'xxxxxxxxxxxxxxxxxxxx'
+        user_id: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
+        volumes: []
+os_volume_res: []
+rax_server_res: []
```
herlo: I wonder if all of this data could be mocked using https://docs.python.org/3/library/unittest.mock.html. Thoughts here?

tests/mockdata/test_input.json
```diff
@@ -0,0 +1 @@
..., "public_ip": "xxx.xxx.xxx.xxx", "private_ip": "xx.xxx.x.x", "machine_type": "n1-standard-1", "metadata": {}, "network": "default"}], "changed": true, "_ansible_no_log": false, "zone": "us-central1-a", "instance_names": ["testresourca12a0"], "_ansible_item_result": true, "instance": ["xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx", "n1-standard-1", "debian-8", "us-central1-a", "present", "xxxxxxxxxxxxx@xxxxxxxxxxxxxxxxxxxxx.iam.gserviceaccount.com", "xxxxxxxxx-xxxx-xxxxxx", "testresourca12", 0], "state": "present", "invocation": {"module_name": "gce", "module_args": {"disks": null, "network": "default", "zone": "us-central1-a", "ip_forward": false, "external_ip": "ephemeral", "service_account_email": "xxxxxxxxxxxxx@xxxxxxxxxxxxxxxxxxxxx.iam.gserviceaccount.com", "instance_names": "testresourca12a0", "service_account_permissions": null, "metadata": null, "persistent_boot_disk": false, "pem_file": null, "preemptible": null, "project_id": "effective-pipe-141600", "state": "present", "machine_type": "n1-standard-1", "credentials_file": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx", "disk_auto_delete": true, "image": "debian-8", "tags": null, "name": null}}}]}
```
herlo: https://docs.python.org/3/library/unittest.mock.html here as well?

tests/linchpin_utils/module_utils.py
```diff
@@ -0,0 +1,214 @@
...
+# this script is for testing modules without running through the
+# entire guts of ansible, and is very helpful for when developing
+# modules
+#
+# example:
+#    test-module -m ../library/commands/command -a "/bin/sleep 3"
+#    test-module -m ../library/system/service -a "name=httpd ensure=restarted"
+#    test-module -m ../library/system/service -a "name=httpd ensure=restarted" --debugger /usr/bin/pdb
+#    test-module -m ../library/file/lineinfile -a "dest=/etc/exports line='/srv/home hostname1(rw,sync)'" --check
+#    test-module -m ../library/commands/command -a "echo hello" -n -o "test_hello"
```
herlo: Why are we basically copy/pasta this code? Could we not install it as a dependency and use it as is?

requirements.txt
```diff
@@ -13,3 +13,4 @@ jsonschema==2.5.1
 ansible==2.1.0
 coverage
 nose
+python-openstackclient==0.8.6
```
herlo: Are nose and python-openstackclient really requirements? Or are they requirements for testing? If I'm running duffy or gce, do I really need either of those?

provision/filter_plugins/aws_inv.py
samvarankashyap: @herlo : I feel adding it to easy_install path is great idea. In that case we need to have a common path/folder for all the packages we write from here on. Something like linchpin_pkgs/packages. We can also have a parameter on linchpin_config.yml  
However , packaging both ansible modules and filter classes together seems not fair , because ansible packages have different purpose when compared to filter classes. I would suggest to maintain them separately.

tests/filter_plugins
samvarankashyap: while invoking ansible playbooks , ansible searches for filter_plugins folder in the immediate dir tree from point of invocation .  In order to make filter_plugins available for ansible api i had to create a symlink. As far as my experience , sys.path works for python script but not ansible. However, i will try it and update you on that.

requirements.txt
samvarankashyap: @herlo : These are the requirements for testing, However nose comes by default in most of the machines. I added it for just in case if we could not find. Further  os_hot_stack ansible module is an implementation of python-openstackclient. That's why we need them in requirements.txt

outputs/all_stack_outputs.testoutput
samvarankashyap: @herlo : yes . We can mockup the test data using unittest.mock. I found it tedious and time taking to write up mock data covering all cases. However , I am planning to use mock or MagicMock to write up mock and update the code with mock classes. Currently I used  sample json file which can pace up writing unit tests.

tests/linchpin_utils/module_utils.py
samvarankashyap: @herlo : Its not just copy paste of the current code , I have to remove certain parts of the code to make it work as a module_util .
I will clean up module_util file .

tests/filter_plugins
samvarankashyap: @herlo : tried adding it to sys.path . I am afraid , its not working because of the same reason i mentioned above . Apperently , ansible searches for filter_plugins from point of invocation . 
we either have to maintain a symlink or , add it to ansible.cfg .

provision/filter_plugins/aws_inv.py
herlo: Two separate packages is fine with me. something like linchpin-modules and linchpin-filters seems appropriate.

PR #67, opened 2016-10-04
bugfix: linchpin_config moved up to fix schema reference

samvarankashyap:


samvarankashyap:
@herlo : fixes the default schema reference

PR #72, opened 2016-10-05
Update requirements.txt: shade

samvarankashyap:


samvarankashyap:
@herlo : updated requirements.txt with shade dependency

Issue #73, opened 2016-10-05
Password not redacted in ansible-playbook output.


herlo:
"instance": ["http://10.8.188.11:5000/v2.0", "atomic-e2e-jenkins", "(redacted but actually displayed)", "atomic-e2e-jenkins-test", "present", "rhel-6.5_jeos", "ci-factory", "m1.small", "atomic-e2e-jenkins-test", "os", "ha_inst", 0]

PR #78, opened 2016-10-17
Redact passwords from outputs in general. 

herlo:
The example output file had a password in it. We've mitigated the password internally, but just in case this shouldn't be in the outputs. Removing it now.

There are many files listed due to a rebase which needed to happen. These files are essentially unchanged, only the last commit changed files related to this PR.

### DIFF
--- a/ex_topo/ex_topology_output.yaml
+++ b/ex_topo/ex_topology_output.yaml
@@ -102,15 +102,15 @@ os_server_res:
     changed: true
     id: f5832fcc-4e1b-442d-8be8-ba5e3783e7f2
     instance:
-    - http://dashboard.centralci.eng.rdu2.redhat.com:5000/v2.0/
-    - e2e-openstack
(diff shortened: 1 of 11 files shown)


ex_topo/ex_topology_output.yaml
```diff
@@ -102,13 +102,13 @@ os_server_res:
...
+    - <endpoint>
+    - <username>
+    - <password>
+    - <username>
     - present
     - rhel-6.5_jeos
-    - ci-factory
+    - <ssh-key-reference>
     - m1.small
     - e2e-openstack
```
arilivigni: <network-name>

herlo: @arilivigni so that change says that I removed the named reference and replaced it with <ssh-key-reference>. Is that not what you want?

herlo:
fixes #73

PR #86, opened 2016-10-20
Bugfixes inventoryfilters

samvarankashyap:
couple of bug fixes:
1. ignore hostgroups vars unless mentioned in inventory_layout
2. ignore omit_place holders in output file


herlo:
Alright, I dug a bit further and printed out the type of the topo['aws_ec2_res'] value and it turns out it's a 'unicode' type, rather than str. Not sure if that impacts this, but I'm going to dig further and see.

herlo:
Changing the line from    
```
if len(topo['aws_ec2_res']) == 0 or type(topo['aws_ec2_res']) is str:
```
to
```
if (len(topo['aws_ec2_res']) == 0 or
    type(topo['aws_ec2_res']) is str or
    type(topo['aws_ec2_res']) is unicode):
```

PR #88, opened 2016-10-31
Linchpincli and setup.py

samvarankashyap:
Summary:
Added setup.py for linchpin cli packaging
contains templates
contains basic implementation 
1. linchpin init
2. linchpin list
3. linchpin get


herlo:
Installing this in a virtualenv gives the following errors:

```
linch-pin-testing]$ python ~/Projects/linch-pin/setup.py install
running install
running bdist_egg
running egg_info
writing requirements to LinchpinCli.egg-info/requires.txt
writing LinchpinCli.egg-info/PKG-INFO
writing top-level names to LinchpinCli.egg-info/top_level.txt
writing dependency_links to LinchpinCli.egg-info/dependency_links.txt
writing entry points to LinchpinCli.egg-info/entry_points.txt
warning: manifest_maker: standard file 'setup.py' not found

file linchpin.py (for module linchpin) not found
error: package directory 'library' does not exist
```

samvarankashyap:
@herlo : could you try cd into the folder and try python setup.py install.

samvarankashyap:
@herlo : followup commits should resolve the issues in virtualenv . 
Could you please check it and let me know.

herlo:
```
(herlo_sandbox)[sig-paas@slave01 linch-pin]$ python setup.py install

..snip..

error: could not create '/etc/linchpin': Permission denied
```
This would be a problem in a virtualenv. One cannot assume /etc/linchpin to exist with the python package. Instead, I might suggest the path be altered based upon the virtualenv path.

samvarankashyap:
@herlo :+1:

herlo:
requirements.txt still has shade==1.9.0 listed twice. Additionally, most reqs should not be pinned (==), but rather set to a minimum version (>=). Can you adjust this?

setup.py
```diff
@@ -55,6 +55,6 @@ def list_all_files(root_dir):
           'templates': list_all_files('templates'),
     },
     data_files=[
-         ('/etc/linchpin', ['linchpin_config.yml']),
+         ('linchpin_config.yml', ['linchpin_config.yml']),
```
herlo: I don't believe this is the way you should do this. Instead, I'd suggest the following

```('', ['linchpin_config.yml'])

Additionally, the version should currently be 0.8.1, maybe? Definitely not 1.0. :)

File: requirements.txt [hunk 2]
```diff
@@ -1,18 +1,17 @@
...
-apache-libcloud==0.20.1
-sphinx==1.4.4
-jsonschema==2.5.1
-ansible==2.1.0
-shade==1.9.0
+boto>=2.40.0
+apache-libcloud>=0.20.1
+sphinx>=1.4.4
+jsonschema>=2.5.1
+ansible>=2.1.0
```
herlo: Pin this to ansible==2.1.2 per the irc conversation wtih @arilivigni

herlo: Also need to add functools32 to the requirements.txt

herlo: And what about jinja2??

herlo:
Moving the discussion from email now that my internet is working. Here's the replay

```
@herlo: What about the .lpf extension. Why not just call the fill Linchfile or somesuch?
@samvarankashyap: Unlike vagrant, I thought it would be great to have multiple files for linchpin CLI instead of one file per folder. We might need filenames with extensions  for the same. so I added a temporary extension.
@herlo: Well, with vagrant, you can actually set an environment variable or pass a flag with the filename. I don't think that's of major concern. 
@arilivigni: I would like to see this work  the same as vagrant since folks use
that pretty heavily.  Just my 2 cents.
@herlo: +1 that's the goal
```
Further thought here leads me down this path. Let's call the file 'Pinfile'. It allows us to detail which services being pinned, or using. One can have multiple configurations in a single file, or multiple files, each with a single configuration in a directory. 

The configurations could be defined simply to point to a topology file.  Each section would have a named cluster or configuration, such as 'across2clouds', providing semantics for the human reader. Additionally, an inventory layout, playbooks which could use the resulting inventory, and more could be provided, allowing configuration of the provisioned resources in some way. 

There may be other bits that would live in this file down the road. Time will tell. It seems linchpin init is going in the right path, especially if we document the Pinfile with good examples and such.

herlo:
linchpin init should add an additional line at the bottom of the file for completeness.

It appears linchpin list currently does nothing?

```
$ linchpin list
linchpin list called !
```

How does one configure linchpin get to pull from a repository of linchpin topologies or layouts? I like the idea here, but maybe we should create a simple github repository with topologies and layouts from which to pull? Kind of like the atomic/docker registry.

samvarankashyap:
@herlo : linchpin list shpuld be called with --topos or --layouts , 
As linchpin list does not return anything .  I should return a message ideally. I will update that.

samvarankashyap:
@herlo In order to get list from a remote we might have to have additional option --remote to fetch the list of layouts and topologies from, and it changes the how we fetch using linchpin get . I will add enhancement to the current functionality which will be addressed soon after the linchpin cli is merged successfully.

samvarankashyap:
@herlo I have raised , PR #92 for the above issue. Please feel free to edit .

PR #115, opened 2016-12-06
linch-pin miscellany

### LINKED ISSUE #114: Duplicate InventoryFilters paths cause confusion
It appears there are duplicate InventoryFilters folders at the root and provision under the linch-pin project.

It is suggested that we remove the provision/InventoryFilters and ensure any extra files there get moved to the InventoryFilters at the root of the tree.

herlo:
- Update libvirt_node schema to allow driver option
- symlinks for InventoryFilters and library to close #114
- fix aws_inv.py failing when attempting to import AWSInventory from
InventoryFilters


ex_schemas/schema_v3.json
```diff
@@ -295,6 +295,9 @@
                     "type":"string",
                     "pattern":"libvirt_network"
                 },
+                "driver": {
```
samvarankashyap: I guess main idea is to give users an option to chose their type drive which can be either "kvm" or qemu right ?
I that case , it would be great if we can limit their options in schema by 
 "driver": {
             "enum": ["kvm", "qemu"]
           }
}
Since we are planning to rewrite the schema validation , the existing also works.

samvarankashyap:
@herlo : You can go ahead and merge the your PR .Currently, The schema made does not effect the existing functionality.

samvarankashyap:
@herlo : I got the error

herlo:
@samvarankashyap alright, moved to libvirt_node object.

PR #121, opened 2016-12-13
Authdriver

samvarankashyap:
Summary:
- Implementation of Auth driver ,
- Updated credential references of openstack, google cloud , aws roles 
- Implementation of cloud_creds package
- Example playbook for authdriver


samvarankashyap:
centos-paas-test

samvarankashyap:
centos-paas-test

README.md
```diff
@@ -163,10 +197,10 @@ url_base: http://admin.ci.centos.org:8080
             res_name: "testvmgce"  # note gce resource names should not contain '_' characters 
             flavor: "n1-standard-1"
             res_type: "gcloud_gce"
-            region: "us-central1-a"
+            region: "us-central1-a" # name of credential_file
```
herlo: Should this example be assoc_creds? Or is the comment misleading?

greg-hellings:
Why are auth values not simply passed as a normal Ansible argument?​ Why do
they need to be passed as a file name that then gets included by way of an
include_vars directive?

Each type that requires authorization could define its own structure, and
then the assoc_creds could reference an entry in that structure. For
instance, openstack could define auth as coming from an Ansible variable
called "openstack_credentials" and the assoc_creds could be referenced in
the Ansible call as:

{{ openstack_credentials[ item['assoc_creds'] ] }}

And the credentials within that object could take whatever form they need
for OpenStack.

That way, these values could be encrypted, stored, etc by already existing
Ansible methods.

greg-hellings:
​It also eliminates the need for a user to create files inside of the
Ansible role structure the way they currently need to do for OpenStack
which then will be called through include_vars and thereby possibly
(accidentally) override existing variables.

herlo:
I think your points are well founded, @greg-hellings. I even wanted to allow for environment variables to be passed as an option in assoc_creds. Essentially, your suggestion is what I was wanting as well. Let the user put in authentication the way that the cloud provider does as standard, and override as necessary (which should be rare).

I don't really care how the data gets passed to the module, really, as long as it's simple and straightforward for the user.

herlo:
I think we can safely close this PR.

PR #122, opened 2016-12-15
Beaker module with proper teardown

herlo:
Thanks to @greg-hellings, we have a beaker module with useful provisioning. However, the teardown required manual intervention. The outputs have the data, so let's use it and teardown without manual adjustment.


herlo:
centos-paas-test

Issue #99, opened 2016-11-08
Tracking bug for openstack os_server issue


herlo:
The following bug affects floating IPs on linch-pin since it depends on shade.

https://github.com/ansible/ansible-modules-core/issues/1981

On our trello board here: https://trello.com/c/mfwC3TmE/107-5-ci-provisioner-linch-pin-look-into-floating-ip-bug

Now in code review: https://review.openstack.org/#/c/393884

herlo:
According to the review, the fix was merged Nov 8, 2016 https://review.openstack.org/#/c/393884. Waiting on verification from upstream ansible bug

johnbieren:
https://github.com/ansible/ansible/issues/19613

samvarankashyap:
#126  fixes the current issue

PR #126, opened 2017-01-17
Bugfix for floating ip issue #99

samvarankashyap:
Summary:
- user can provide the floating ip pools using fip_pool attribute in topology.
- schema update to incorporate floating ip pools
- added os_net for translating networks to nics

### DIFF
--- a/provision/roles/aws/templates/output_formatter.j2
+++ b/provision/roles/aws/templates/output_formatter.j2
@@ -1,1 +1,1 @@
-{{ topology_outputs | to_nice_yaml }}
+{{ topology_outputs }}
--- a/docs/source/intro_installation.rst
+++ b/docs/source/intro_installation.rst
@@ -45,4 +45,5 @@ Install from source.
 
     $ git clone https://github.com/CentOS-PaaS-SIG/linch-pin.git --recursive
     $ cd ./linch-pin
-    $ sudo pip install -r requirements.txt # installs the dependencies required for linch-pin
+    $ sudo pip install setuptools --upgrade  # upgrade setuptools if you are on RHEL 7.x 
+    $ python setup.py install
--- a/ex_schemas/schema_v3.json
+++ b/ex_schemas/schema_v3.json
@@ -58,7 +58,10 @@
                 },
                 "keypair": {
                     "type":"string"
-                }
+                },
+		"fip_pool": {
+		    "type":"string"
+		}
             },
             "required":["flavor","res_type","image","count","keypair"]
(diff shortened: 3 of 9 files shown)


herlo:
centos-paas-test

samvarankashyap:
@herlo @johnbieren could you please test the current PR and let us know if there are any issues.

johnbieren:
As of this PR, I have to do the following things which I did not do before (with current master):
- python setup.py install (this should be in the readthedocs installation guide if it is supposed to always be necessary)
- pip install setuptools --upgrade
- pip install python-keystoneclient
- modify library in /etc/ansible/ansible.cfg to point to the library dir in the linchpin dir

I also think it would be good to add fip_pools to ex_topo/ex_os_server_fp.yml, assuming fp = floating IP.  Hopefully people would think to look there.

samvarankashyap:
Thank @johnbieren for input.
Will be updating the read the docs with above to steps.
pip install keystone client is not necessary because its already specified in requirements.txt
As for the library path , Ansible should be able to resolve the library automatically will be looking into it.
I have already added ex_topo/ex_os_server_fp2.yml as a example for topology with floating ip pools.

arilivigni:
centos-paas-test

herlo:
@johnbieren you shouldn't have to perform the setuptools upgrade. But we are working on some of these requirements to make things simpler. I expect you'll be able to a simple pip install linch-pin and things will work soon.

@samvarankashyap why is python-keystoneclient now required? Isn't that a dep of python-openstackclient?

samvarankashyap:
@herlo I meant the same ,The keystone client  already included with openstackclient .

samvarankashyap:
centos-paas-test

samvarankashyap:
centos-paas-test

samvarankashyap:
centos-paas-test

samvarankashyap:
centos-paas-test

samvarankashyap:
centos-paas-test

samvarankashyap:
centos-paas-test

provision/roles/aws/templates/output_formatter.j2
```diff
@@ -1 +1 @@
-{{ topology_outputs | to_nice_yaml }}
+{{ topology_outputs }}
```
herlo: As discussed, this will need to be returned. I am aware of the other PR with ansible v2.2.0 in place to ensure inventories are properly generated. AIUI, the to_nice_yaml filter is also updated.

PR #137, opened 2017-01-24
bugfix: revert back to list_all_files in setup.py

samvarankashyap:
revert back to list_all_files because findall function in setuptools is skipping subpackages causing linchpin 0.8.3, 0.8.4 builds fail while installing through pip.


samvarankashyap:
centos-paas-test

PR #139, opened 2017-01-26
Add more default paths to look for linchpin_config

herlo:
If, for some reason, we can't find the linchpin_config.yml in the initial paths, use the default one in the package. It should be somewhere in the sys.path.


herlo:
centos-paas-test

samvarankashyap:
LGTM

herlo:
centos-paas-sig

Issue #167, opened 2017-03-03
AWS_EC2 fails when the res_grp_vars are not defined


samvarankashyap:
Example topology : 
```yaml 
---
    topology_name: "ex_aws_topo"
    site: "qeos"
    resource_groups: 
      - 
        resource_group_name: "testgroup1"
        res_group_type: "aws"
        res_defs:
          - 
            res_name: "ha_inst" 
            flavor: "t2.micro"
            res_type: "aws_ec2"
            region: "us-east-1"
            image: "ami-f4cc1de2"
            count: 1
            keypair: "sk_key"
        assoc_creds: "sk_aws_personal"
```
Error : 
```
TASK [aws : Provisioning AWS_EC2 Resource when async == false] *****************
fatal: [localhost]: FAILED! => {"failed": true, "msg": "the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'res_grp_vars' is undefined\n\nThe error appears to have been in '/home/srallaba/.virtualenvs/newenv/lib/python2.7/site-packages/linchpin-0.8.6-py2.7.egg/provision/roles/aws/tasks/provision_aws_ec2.yml': line 9, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: \"Provisioning AWS_EC2 Resource when async == false\"\n  ^ here\n"}
        to retry, use: --limit @/home/srallaba/.virtualenvs/newenv/lib/python2.7/site-packages/linchpin-0.8.6-py2.7.egg/provision/site.retry

PLAY RECAP *********************************************************************
localhost                  : ok=19   changed=0    unreachable=0    failed=1 
```

samvarankashyap:
#166 fixes this issue

PR #168, opened 2017-03-03
Hot fixes : linchpin_config resolution , aws_ec2 default instance tags #167

samvarankashyap:
Summary :
-   While initiating provisioning , it is observed that linchpin cli passes path of linchpin_config.yaml
file in linchpin_config extra var. However , its not being used while resolving the linchpin config file
due to a typo. This current commit resolves the config resolution issue of linchpin.
-   Currently , Provisioning of aws_ec2 instances are failing when res_grp_vars in undefined due to the  fact res_grp_vars are being used as instance tags while provisioning ec2 instance. In the current commit fixes the bug by setting resource_group_name and res_name as  default values of res_grp_vars for any instance when the res_grp_vars are undefined in topology.
- Resolves : issue #167

### DIFF
--- a/provision/roles/aws/tasks/provision_aws_ec2.yml
+++ b/provision/roles/aws/tasks/provision_aws_ec2.yml
@@ -1,3 +1,11 @@
+- name: "Set default res_grp_vars when not defined"
+  set_fact:
+    res_grp_vars:
+      -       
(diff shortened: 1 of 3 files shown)


herlo:
centos-paas-test

samvarankashyap:
@herlo tests passed . Merging the PR.

PR #195, opened 2017-03-28
Added Beaker topology example to docs

robled:


docs/source/topologies_beaker.rst
```diff
@@ -0,0 +1,36 @@
...
+            arch: x86_64
+            arches:
+              - X86_64
+              - X86_64
+            keyvalue:
+              - MEMORY>1000
+              - DISKSPACE>20000
+            hostrequire:
+              - arch=X86_64
+    count: 1
```
herlo: Is 'count' in the right place? It seems it's at the base level, but should it not be related to at least the resource_group or the recipeset?

robled: Good question.  This example topology was submitted a while ago, and I have used it as-is in a number of my recent test runs.  I'll try indenting the **count** key and I'll report my results.

robled: I think this was a copypasta error; the indentation is lost when copying from the GitHub web interface.  Nevertheless, I did run a test to verify that within the **recipesets** section is the correct place to put the **count** setting.  Here's a test run where I requested two Beaker servers and successfully provisioned and then dropped them:

https://gist.github.com/robled/23a1f844aeb811f1e5309906c5aa07a3

PR #215, opened 2017-04-24
Declare babel package with a minimum version.

rywillia:
Running into issues when installing linchpin. Seems one of its packages it depends on needs a version of babel not equal too 2.4.0. Switching to a minimum version of 2.3.4 allows the install to work.


herlo:
Also, thank you for your submission. It's exciting to see new contributors! :+1:

rywillia:
@herlo I have moved the requirement as requested. Thanks for pointing that out!

PR #218, opened 2017-04-25
Linchpin hooks :: resolves #96

### LINKED ISSUE #96: (POC) (8) Linchpin Hooks: Pinfile should provide a 'post_actions' yaml list
Currently, the Pinfile provides the ability to have a particular target, with both a topology and a layout option. Additionally, it may be useful to pull down and run ansible playbooks based upon the inventory returned from the provisioning. 

Simply put, this could make CI and developer tooling very useful. 

Consider the following Pinfile:

```
duffy3node:
  topology_name: duffy_3node_cluster.yml
  inventory_layout_file: duffy_3node_inventory_layout.yml
  post_actions:
    type: ansible
    base_dir: ./ansible
    targets:
      - setup.yml
      - origin_from_source.yml
      - openshift_ansible_from_source.yml
      - deploy_aosi.yml
      - run_e2e_tests.yml
    type: shell
    base_dir: ./scripts
    upstream: git://github.com/herlo/test_scripts_to_run.git
    targets:
      - gather_stats.sh
```

Essentially, this option would provision resources from the named topology / inventory combination. Proceeding from top down, the `post_actions` would first run ansible files already existing locally from the relative `./ansible` path. Starting with `setup.yml`, and continuing down the list. If at any point it failed, execution would stop. 

If successful, the shell `post_actions` would run. The scripts, if non-existent (or a cached timeout expired), would be cloned from the named repository, placed in the relative `./scripts` path. Following the download/checkout, the `gather_stats.sh` would be executed.

samvarankashyap:
Summary:
- linchpin now supports hooks in Pinfile in following states 
  preup, predown, postup and post down 
  ```
  ref: example below
  ```
- hooks are resolved by name ie., if we have a hook named do_something of ansible type the workspace should contain following layout with all the playbooks.
```
.
├── hooks
│   ├── ansible
│   │   ├── do_something
│   │   │   ├── test_playbook.yaml
│   │   │   ├── test_vars.yaml
```
example Pinfile for above ansible hook looks as follows:
```
---
openstack:
  topology: ex_os_server.yml
  layout: openshift-3node-cluster.yml
  hooks:
    preup:              
      - name: postdown_task
        type: ansible
        actions:
          - playbook: test_playbook.yaml
            vars: test_var.yaml
            extra_vars: { "test_var": "postdowntask"}
```
- hooks also supports custom action manager , where the action managers need to implement the following linchpin.hooks.action_manager.ActionManager abstract class .
for example if user would like to implement their own action_manager of type testtype :
hooks folder layout looks as follows : 
```
.
├── hooks
│   └── testtype
│       └── test_something
│           ├── testact.py
│           └── testact.pyc

```
example Pinfile for the above testtype hook looks as follows:
```
openstack:
  topology: ex_os_server.yml
  layout: openshift-3node-cluster.yml
  hooks:
    preup:              
      - name: test_something
        action_manager: testact.py
        type: testtype
        actions:
          - "This does nothing"
```

Example Pinfile:
```
---
openstack:
  topology: ex_os_server.yml
  layout: openshift-3node-cluster.yml
  hooks:
    postup:            
        - name: do_something
          type: python
          context: false
          actions:
            - test.py
        - name: testruby
          type: ruby
          context: false
          actions:
            - test.rb
        - name: testnode
          type: node
          context: true
          actions:
            - test.js
    preup:              
        - name: test_something
          action_manager: testact.py
          type: testtype
          actions:
            - "This does nothing"
        - name: post_up
          type: ansible
          context: false
          actions:
            - playbook: test_playbook.yaml
              vars: test_var.yaml
              extra_vars: { "test_var": "preup task"}
    postdown:
        - name: do_something
          type: shell
          actions:
            - echo " this is post down operation Hello hai how r u ?"
        - name: postdown_task
          type: ansible
          actions:
            - playbook: test_playbook.yaml
              vars: test_var.yaml
              extra_vars: { "test_var": "postdowntask"}
```
current workspace layout required for testing the above Pinfile
```
.
├── hooks
│   ├── ansible
│   │   ├── build_openshift_cluster
│   │   │   ├── test_playbook.yaml
│   │   │   └── test_var.yaml
│   │   ├── postdown_task
│   │   │   ├── test_playbook.yaml
│   │   │   └── test_var.yaml
│   │   └── post_up
│   │       ├── test_playbook.yaml
│   │       └── test_var.yaml
│   ├── node
│   │   └── testnode
│   │       └── test.js
│   ├── python
│   │   └── do_something
│   │       └── test.py
│   ├── ruby
│   │   └── testruby
│   │       └── test.rb
│   ├── shell
│   │   └── thisisshell.sh
│   └── testtype
│       └── test_something
│           ├── testact.py
│           └── testact.pyc
├── inventories
├── layouts
├── outputs
├── PinFile
├── resources
└── topologies
```

Test: 
- copy pinfile from linchpin/examples/OS_SERVER_PINFILE to workspace
- edit Pinfile <accordingly, if using ruby, node based hooks ruby, node should be installed on host>
- linchpin rise
- linchpin drop
example workspace for testing : https://github.com/samvarankashyap/linchtest

Issues to be resolved:
- Need to add documentation for states and hooks
- Need to add developer documentation for custom_action_manager
- Need to implement hooks with global scope.
Note : refer #96


linchpin/api/__init__.py
```diff
@@ -1,28 +1,77 @@
...
+        for callback in self._state_observers:
+            callback(self._state)
+
+    def bind_to_state(self, callback):
+        self._state_observers.append(callback)
+
+    def prepare_ctx_params(self):
+
+        """
+        prepares few context parameters based on the current target_data
```
herlo: Please use evars here, as context.py and linchpin.conf already support this functionality.

linchpin/api/__init__.py
```diff
@@ -70,16 +113,31 @@ def run_playbook(self, pinfile, targets='all', playbook='provision'):
...
-#                def invoke_linchpin(ctx, lp_path, self.ctx.evars, playbook='provision', console=True):
+                # set the current target data 
+                self.current_target_data = pf[target]
+                self.current_target_data["extra_vars"] = self.ctx.evars
+                # set the state to preup/predown based on playbook
+                # note : changing the state triggers the hooks
+                if playbook == "provision":
+                    self.state = "preup"
+                elif playbook == "destroy":
+                    self.state = "predown"
```
herlo: As I mentioned in my main comment, predown should not be used here, predestroy would be appropriate.

linchpin/api/__init__.py
```diff
@@ -1,28 +1,77 @@
...
+        """
+        prepares few context parameters based on the current target_data
+        that is being set. 
+        res_file and inventory file are not found
+        These parameters are based topology name.
+        """
 
+        topology = self.current_target_data["topology"]
+        name = topology.split("/")[-1].split(".")[-2]
+        inv_file = name+".inventory"
```
herlo: DO NOT HARDCODE THESE. If you need to have a full name for the inventory file, look at linchpin.conf as the defacto source and add a config option there. In fact, this function feels like it should be in the context.py.

linchpin/api/__init__.py
```diff
@@ -37,25 +86,19 @@ def run_playbook(self, pinfile, targets='all', playbook='provision'):
         """
 
         pf = yaml2json(pinfile)
-
```
herlo: Nitpick here. Removing the spacing makes it harder to read this code. I specifically separated these lines for two reasons. 1) readability and 2) logical sectioning. Can we put the spacing back where appropriate?

linchpin/api/__init__.py
```diff
@@ -70,16 +113,31 @@ def run_playbook(self, pinfile, targets='all', playbook='provision'):
...
                                     pf[target]["layout"]))
 
-
-#                def invoke_linchpin(ctx, lp_path, self.ctx.evars, playbook='provision', console=True):
+                # set the current target data 
+                self.current_target_data = pf[target]
+                self.current_target_data["extra_vars"] = self.ctx.evars
+                # set the state to preup/predown based on playbook
+                # note : changing the state triggers the hooks
+                if playbook == "provision":
```
herlo: Hardcoding this here means that you can't predict what the playbook is going to be. In fact, in upcoming changes the playbook is no longer called 'provision', but rather 'up'. Hardcoding this in means it has to be changed in every place it's updated. Instead, consider getting the list of states 'up', 'destroy', 'down', etc., from configurations and assigning them to a variable.

linchpin/api/__init__.py
```diff
@@ -92,21 +150,33 @@ def run_playbook(self, pinfile, targets='all', playbook='provision'):
                         '{0}/{1}/{2}'.format(self.ctx.workspace,
                                     self.ctx.cfgs['evars']['layouts_folder'],
                                     pf[target]["layout"]))
-
-
-
+                # set the current target data
+                self.current_target_data = pf[target]
+                self.current_target_data["extra_vars"] = self.ctx.evars
```
herlo: Why not add values to the ctx.evars variable as opposed to creating a new 'extra_vars' component here?

linchpin/api/__init__.py
```diff
@@ -1,28 +1,77 @@
 #!/usr/bin/env python
 
 import os
-
```
herlo: The spacing here is intentional. It keeps stdlibs and linchpin libs visibly separate and makes it easier to read and recognize what libraries are being imported quickly.

linchpin/api/__init__.py
```diff
@@ -1,28 +1,77 @@
...
+
+    @property
+    def state(self):
+        """ getter function for state property of the API object. """
+        return self._state
+
+    @state.setter
+    def state(self, state):
+        # call run_hooks after state is being set
+        print("State change initiated")
```
herlo: Remove print statements. If wanted, use ctx.log, ctx.log_state (which outputs to stderr), ctx.log_info, or ctx.log_debug. If that's not enough, more can be created in the context.py.

linchpin/examples/pinfiles/OS_SERVER_PINFILE
```diff
@@ -0,0 +1,45 @@
...
+          actions:
+            - "This does nothing"
+        - name: post_up
+          type: ansible
+          context: false
+          actions:
+            - playbook: test_playbook.yaml
+              vars: test_var.yaml
+              extra_vars: { "test_var": "preup task"}
+    postdown:
```
herlo: Should be postdestroy.

linchpin/hooks/__init__.py
```diff
@@ -0,0 +1,150 @@
...
+"""
+
+import pprint
+import ast
+import sys
+from core_action_managers import ACTION_MANAGERS
+
+
+class ActionBlockRouter(object):
+    def __init__(self, name, *args, **kwargs):
```
herlo: Please document the class and functions with a short description and any parameters.

linchpin/hooks/__init__.py
```diff
@@ -0,0 +1,150 @@
+"""
```
herlo: Are hooks really not part of the API? Seems they might fit better there. Thoughts?

linchpin/hooks/__init__.py
```diff
@@ -0,0 +1,150 @@
...
+    def __init__(self, api):
+        self.api = api
+        self.api.bind_to_state(self.run_hooks)
+
+    def run_hooks(self, state, is_global=False):
+        print("State change triggered in linchpin API")
+        print("Observed State in LinchpinHooks :: "+str(state))
+        hooks_data = self.api.current_target_data.get("hooks", None)
+        if hooks_data == None:
+            print("No hooks found for current target")
```
herlo: Use ctx.log_state here. No more print statements.

linchpin/hooks/__init__.py
```diff
@@ -0,0 +1,150 @@
...
+            return
+        # fetches all the state_data , ie., all the action blocks inside
+        # state of the target
+        state_data = hooks_data.get(str(state), None)
+        # Print out error message if the hooks are not found
+        if state_data == None:
+            print(str(state)+" state hook not found in PinFile")
```
herlo: Use ctx.log_state here. No more print statements.

linchpin/hooks/ansible_action_manager.py
```diff
@@ -0,0 +1,154 @@
...
+        }
+        v = Validator(schema)
+        status = v.validate(self.action_data)
+        if not status:
+            raise Exception("Invalid syntax: LinchpinHook:"+str((v.errors)))
+        else:
+            return status
+
+    def load(self):
+        print("Load module of Ansible Action Manager ")
```
herlo: Use ctx.log here. No more print statements.

linchpin/hooks/core_action_managers.py
```diff
@@ -0,0 +1,13 @@
+from subprocess_action_manager import SubprocessActionManager
```
herlo: This should be put in the __init__.py instead as it's just setting up data for the package and modules inside.

linchpin/hooks/python_action_manager.py
```diff
@@ -0,0 +1,58 @@
...
+        'actions': { 
+                     'type': 'list',
+                     'schema': {'type':'string'},
+                     'required': True
+                   }
+        }
+        v = Validator(schema)
+        status = v.validate(self.action_data)
+        if not status:
+            raise Exception("Invalid syntax: LinchpinHook:"+str((v.errors)))
```
herlo: Should be a LinchpinError here.

linchpin/hooks/python_action_manager.py
```diff
@@ -0,0 +1,58 @@
...
+                                    file_path)
+        params = ""
+        for key in self.target_data:
+            params += " %s=%s " %(key, self.target_data[key])
+        return "{0} {1} {2}".format(sys.executable,
+                                    file_path,
+                                    params)
+
+    def execute(self):
+        print("Execute module of PythonActionManager")
```
herlo: Print statements are bad. Use ctx.log_state to output to stderr

linchpin/hooks/ruby_action_manager.py
```diff
@@ -0,0 +1,54 @@
...
+        'actions': {
+                     'type': 'list',
+                     'schema': {'type':'string'},
+                     'required': True
+                   }
+        }
+        v = Validator(schema)
+        status = v.validate(self.action_data)
+        if not status:
+            raise Exception("Invalid syntax: LinchpinHook:"+str((v.errors)))
```
herlo: Use LinchpinError

linchpin/hooks/state.py
```diff
@@ -0,0 +1,24 @@
+class State(object):
+    VALID_STATES = ["preup", "predown", "postup", "postdown"]
+    #VALID_SUB_STATES = ["preup", "predown", "postup", "postdown"]
+    def __init__(self, state, sub_state ):
+        if self._validate_state(state):
+            self.state = state
+        else:
+            raise Exception("Invalid State mentioned")
```
herlo: Use LinchpinError

linchpin/hooks/subprocess_action_manager.py
```diff
@@ -0,0 +1,78 @@
...
+                               'schema': {
+                                          'type': 'string'
+                                         },
+                               'required': True
+                             }
+        }
+        v = Validator(schema)
+        status = v.validate(self.action_data)
+        if not status:
+            raise Exception("Invalid syntax: LinchpinHook:"+str((v.errors)))
```
herlo: Use LinchpinError

linchpin/hooks/node_action_manager.py
```diff
@@ -0,0 +1,54 @@
...
+        if not context:
+            return file_path
+        params = file_path
+        for key in self.target_data:
+            params += " %s=%s " %(key, self.target_data[key])
+        return "{0} {1}".format(file_path,
+                                    params)
+
+    def execute(self):
+        print("Execute module of RubyActionManager")
```
herlo: Use ctx.log_state or something in ctx.log* family.

linchpin/hooks/node_action_manager.py
```diff
@@ -0,0 +1,54 @@
...
+        'actions': {
+                     'type': 'list',
+                     'schema': {'type':'string'},
+                     'required': True
+                   }
+        }
+        v = Validator(schema)
+        status = v.validate(self.action_data)
+        if not status:
+            raise Exception("Invalid syntax: LinchpinHook:"+str((v.errors)))
```
herlo: Use LinchpinError

linchpin/hooks/ansible_action_manager.py
```diff
@@ -0,0 +1,154 @@
...
+                             'extra_vars': {'type': 'dict', 'required': False}
+                         }
+                     },
+                     'required': True
+                   }
+        }
+        v = Validator(schema)
+        status = v.validate(self.action_data)
+        if not status:
+            raise Exception("Invalid syntax: LinchpinHook:"+str((v.errors)))
```
herlo: Use LinchpinError

linchpin/hooks/__init__.py
```diff
@@ -0,0 +1,150 @@
...
+    def __init__(self, name, *args, **kwargs):
+        self.__implementation = self.__get_implementation(name)(name, *args, **kwargs)
+
+    def __getattr__(self, name):
+        return getattr(self.__implementation, name)
+
+    def __get_implementation(self, class_name):
+        action_class = ACTION_MANAGERS.get(class_name, None)
+        if action_class == None:
+            raise Exception("Action Class %s not found " % (class_name))
```
herlo: Use LinchpinError

linchpin/api/__init__.py
```diff
@@ -1,28 +1,77 @@
...
 from linchpin.api.invoke_playbooks import invoke_linchpin
 from linchpin.api.utils import yaml2json
+from linchpin.hooks import LinchpinHooks
+from linchpin.hooks.state import State
 
 
 class LinchpinError(Exception):
 
     def __init__(self,*args,**kwargs):
         Exception.__init__(self,*args,**kwargs)
```
herlo: Use LinchpinError

linchpin/api/__init__.py
herlo: After some conversation with @samvarankashyap, it's clear that this current_target_data dictionary needs to copy the ctx.evars value.

linchpin/hooks/ruby_action_manager.py
```diff
@@ -0,0 +1,54 @@
...
+        if not context:
+            return file_path
+        params = file_path
+        for key in self.target_data:
+            params += " %s=%s " %(key, self.target_data[key])
+        return "{0} {1}".format(file_path,
+                                    params)
+
+    def execute(self):
+        print("Execute module of RubyActionManager")
```
herlo: use ctx log values here. NO PRINTING! :)

linchpin/hooks/state.py
```diff
@@ -0,0 +1,24 @@
+class State(object):
+    VALID_STATES = ["preup", "predown", "postup", "postdown"]
```
herlo: This should come from a configuration file, not here.

linchpin/hooks/state.py
```diff
@@ -0,0 +1,24 @@
...
+        if self._validate_state(state):
+            self.state = state
+        else:
+            raise Exception("Invalid State mentioned")
+        if sub_state == None:
+            self.sub_state = sub_state
+        elif self._validate_sub_state(sub_state):
+            self.sub_state = sub_state
+        else:
+            raise Exception("Invalid SubState mentioned")
```
herlo: LinpinError

linchpin/hooks/state.py
```diff
@@ -0,0 +1,24 @@
...
+    def _validate_state(self, state):
+        return state in State.VALID_STATES
+    def _validate_sub_state(self, sub_state):
```
herlo: Spacing here would be good. More readable.

linchpin/hooks/subprocess_action_manager.py
```diff
@@ -0,0 +1,78 @@
...
+            os.environ["PATH"] += ":"+self.action_data["path"]
+
+    def add_context_params(self, action):
+        command = action
+        for key in self.target_data:
+            command += " %s=%s " %(key, self.target_data[key])
+        return command
+
+    def execute(self):
+        print("Running Execute module of SubprocessAction Manager")
```
herlo: ctx logging plz

samvarankashyap:
status_update: 
1. still need to fix hardcoded values,
2. move prepare_ctx_params to hooks/__init__.py
and done with this PR.

samvarankashyap:
status_update:
successfully moved prepare_ctx_params to hooks/init.py
remove hardcoded values. 
Things to be finished :  
1. need to add documentation wherever necessary in functions and proper spacing 
2. test the provisioning and teardown

linchpin/hooks/__init__.py
```diff
@@ -0,0 +1,184 @@
...
+          type: shell
+          actions:
+            - echo " this is post up operation Hello hai how r u ?"
+        - name: build_openshift_cluster
+          type: ansible
+          actions:
+            - playbook: test_playbook.yaml
+              vars: test_var.yaml
+              extra_vars: { "testvar": "world"}
+    postdown:
```
herlo: postdestroy? You still have a few of these.

samvarankashyap:
status_update:
1. finished documenting 

needs testing

samvarankashyap:
```
linchpin rise # with hooks output 
TASK [inventory_gen : Generate Generic Inventory] ******************************
changed: [localhost]

localhost                  : ok=28   changed=3    unreachable=0    failed=0   
```
```
linchpin drop # with hooks works
localhost                  : ok=22   changed=1    unreachable=0    failed=0   
```
example workspace : 
testworkspace:
https://github.com/samvarankashyap/linchtest.git
steps to test: 
```
git clone https://github.com/samvarankashyap/linchtest.git;
cd linchtest;
linchpin rise;
linchpin drop;
```

linchpin/hooks/__init__.py
```diff
@@ -0,0 +1,200 @@
...
+        
+        """
+        prepares few context parameters based on the current target_data
+        that is being set. these parameters are based topology name.
+        """
+        
+        topology = self.api.current_target_data["topology"]
+        name = topology.split("/")[-1].split(".")[-2]
+        inv_file = name+".inventory"
+        res_file = name+".output"
```
herlo: These two extensions seem like hardcoded values. I know that generally speaking, the names of the files will be called something.inventory and something.output, but might be something different (something.resources maybe?) in the future. My point here, though, is if we change the extension, we have to find this entry in the file while debugging. Can these values be stored in linchpin.conf or some such?

linchpin/hooks/__init__.py
```diff
@@ -0,0 +1,200 @@
...
+        """
+        Get the cfgs object
+        :param section: section from ini-style config file
+        :param key: key to get from config file, within section
+        """
+
+        if not str(state) in self.api.ctx.cfgs["playbook_pre_states"]:
+            self.prepare_ctx_params()
+        self.api.ctx.log("State change triggered in linchpin API")
+        self.api.ctx.log("Observed State in LinchpinHooks :: "+str(state))
```
herlo: Don't use log directly here. This is probably better a log_debug() call.

linchpin/hooks/__init__.py
```diff
@@ -0,0 +1,200 @@
...
+        hooks_data = self.api.current_target_data.get("hooks", None)
+        if hooks_data == None:
+            self.api.ctx.log("No hooks found for current target")
```
herlo: use log_debug?

linchpin/hooks/__init__.py
```diff
@@ -0,0 +1,200 @@
...
+            return
+        # fetches all the state_data , ie., all the action blocks inside
+        # state of the target
+        state_data = hooks_data.get(str(state), None)
+
+        # Print out error message if the hooks are not found
+        if state_data == None:
+            self.api.ctx.log(str(state)+" State hook not found in PinFile")
```
herlo: use log_debug?

linchpin/hooks/__init__.py
```diff
@@ -0,0 +1,200 @@
...
+        is_global: scope of the hook.
+        example: action_block: 
+        - name: do_something
+          type: shell
+          actions:
+            - echo " this is post up operation Hello hai how r u ?"
+        """
+
+        if is_global:
+            raise NotImplementedError("Run Hooks is not implemented \
```
herlo: :+1:

linchpin/hooks/__init__.py
```diff
@@ -0,0 +1,200 @@
...
+                                       for global scoped hooks")
+        else:
+            # a_b -> abbr for action_block
+            for a_b in action_blocks:
+                action_type = a_b["type"]
+                ctx = a_b['context'] if a_b.has_key('context') else True
+                if not a_b.has_key("path"):
+                    # if the path is not defined it defaults to
+                    # workspace/hooks/typeofhook/name
+                    a_b["path"] = "{0}/hooks/{1}/{2}/".format(
```
herlo: hooks is hardcoded here. Yes, hooks will likely not change, but let's not use hardcoded values. Make it a configuration variable.

linchpin/hooks/action_managers/action_manager.py
```diff
@@ -0,0 +1,13 @@
+from abc import ABCMeta, abstractmethod
```
herlo: why is this not the __init__.py?

linchpin/hooks/action_managers/ansible_action_manager.py
```diff
@@ -0,0 +1,182 @@
...
+from action_manager import ActionManager
+from cerberus import Validator
+
+from linchpin.exceptions import HookError
+
+
+class AnsibleActionManager(ActionManager):
+
+    def __init__(self, name, action_data, target_data, **kwargs):
+
```
herlo: Probably ought to document these params and how they are used below.

linchpin/hooks/action_managers/node_action_manager.py
```diff
@@ -0,0 +1,75 @@
+from action_manager import ActionManager
```
herlo: Small nitpick, should it not be called nodejs_action_manager.py?

linchpin/hooks/action_managers/node_action_manager.py
```diff
@@ -0,0 +1,75 @@
...
+from Naked.toolshed.shell import run_js
+from cerberus import Validator
+
+from linchpin.exceptions import HookError
+
+
+class NodeActionManager(ActionManager):
+    
+    def __init__(self, name, action_data, target_data, **kwargs):
+
```
herlo: Probably ought to document these params and how they are used below.

linchpin/hooks/action_managers/python_action_manager.py
```diff
@@ -0,0 +1,78 @@
+from action_manager import ActionManager
+import sys
+import subprocess
+from cerberus import Validator
+
+from linchpin.exceptions import HookError
+
+
+class PythonActionManager(ActionManager):
+
```
herlo: Probably ought to document these params and how they are used below.

linchpin/hooks/action_managers/ruby_action_manager.py
```diff
@@ -0,0 +1,74 @@
...
+from Naked.toolshed.shell import run_rb
+from cerberus import Validator
+
+from linchpin.exceptions import HookError
+
+
+class RubyActionManager(ActionManager):
+
+    def __init__(self, name, action_data, target_data, **kwargs):
+
```
herlo: Probably ought to document these params and how they are used below.

linchpin/hooks/action_managers/subprocess_action_manager.py
```diff
@@ -0,0 +1,101 @@
...
+from action_manager import ActionManager
+from cerberus import Validator
+
+from linchpin.exceptions import HookError
+
+
+class SubprocessActionManager(ActionManager):
+
+    def __init__(self, name, action_data, target_data, **kwargs):
+
```
herlo: Probably ought to document these params and how they are used below.

linchpin/hooks/state.py
```diff
@@ -0,0 +1,50 @@
+from linchpin.exceptions import StateError 
+
+class State(object):
+
+    # currently unimplemented
+    VALID_SUB_STATES = []
+
+    def __init__(self, state, sub_state , api=None):
+        self.api = api
```
herlo: Probably ought to document this and explain what it does.

linchpin/linchpin.conf
```diff
@@ -88,3 +88,18 @@ level = logging.INFO
...
 console = True
+
+[playbook_pre_states]
+up = preup
+destroy = predestroy
+
+[playbook_post_states]
+up = postup
+destroy = postdestroy
+
```
herlo: Since there aren't any current `substates`, couldn't the above sections be in the `[states]` section below? Something like:

```
[states]
preup = preup
postup = postup
predestroy = predestroy
postdestroy = postdestroy
```
And then just add a `[substates]` section later when needed?

linchpin/provision/roles/inventory_gen/tasks/main.yml
```diff
@@ -27,4 +27,4 @@
 - name: "Generate Generic Inventory"
   template:
     src: "templates/generic_inventory_formatter.j2"
-    dest: "{{ inventory_file }}.inventory"
+    dest: "{{ inventory_file }}"
```
herlo: :+1:

samvarankashyap:
status_update:
1. Need to resolve the two instances of hardcoded values , rest of the changes are made as suggested.

samvarankashyap:
@herlo status_update:
Done with all the changes requested.

linchpin/hooks/action_managers/action_manager.py
samvarankashyap: @herlo tried moving it into __init__.py , linchpin was unable to resolve the ActionManager class , so I had to keep it in a separate file.

samvarankashyap:
output:
linchpin rise: 
```
TASK [inventory_gen : Generate Generic Inventory] ******************************
changed: [localhost]

localhost                  : ok=28   changed=3    unreachable=0    failed=0 
```

linchpin drop:
```
TASK [inventory_gen : Generate Generic Inventory] ******************************
skipping: [localhost]

localhost                  : ok=22   changed=1    unreachable=0    failed=0  
```
workspace:
https://github.com/samvarankashyap/linchtest

samvarankashyap:
@herlo  Rebased and resolved conflicts in requirement.txt

linchpin/linchpin.conf
```diff
@@ -88,3 +89,23 @@ level = logging.INFO
...
+[states]
+# in future each state will carry a comma seperated substates
+preup = 'preup'  
+predestroy = 'predestroy'
+postup = 'postup'
+postdestroy = 'postdestroy'
+
+[extns]
+resource = '.output'
+inventory = '.inventory'
```
seandst: These single-quotes are included in the value, so the inventory file is created with them included, e.g. "filename'.inventory'"

```python
>>> from configparser import ConfigParser
>>> cp = ConfigParser()
>>> cp.read_string("""
... [section]
... opt = 'value'
... """)
>>> cp.get('section', 'opt')
"'value'"
```
^ note the included single quotes

seandst:
I did some testing with the [cinch playbooks](https://github.com/RedHatQE/cinch/tree/master/cinch), and these are the results. I apologize in advance for the megacomment.

# playbook with absolute path

I started with this PinFile:
```yaml
cinch:
  topology: cinch.yml
  layout: cinch.yml
  hooks:
    postup:
      - name: cinchup
        type: shell
        context: false
        actions:
          - echo Running postup
      - name: cinchup_ansible
        type: ansible
        actions:
          - playbook: /path/to/cinch/site.yml
    predestroy:
      - name: cinchdestroy
        type: shell
        context: false
        actions:
          - echo Running predestroy
      - name: cinchdestroy_ansible
        type: ansible
        actions:
          - playbook: /path/to/cinch/teardown.yml
```

The output indicates the site playbook is never run when the postup hook fires:
```
localhost                  : ok=27   changed=0    unreachable=0    failed=0   

Running postup

$
```

linchpin.log says:
```
INFO 2017-05-10 15:02:36,739 the file_name '/path/to/linchpin/workspace/hooks/ansible/cinchup_ansible/path/to/cinch/site.yml' does not exist, or is not readable
```

I think this is wrong on a few points:
* The path to the playbook is absolute, but treated as a relative path. It is not intuitive to always treat it as a relative path, particularly when given a relative path. Prior to handing the playbook to ansible, it might be good to run the path through [`os.path.relpath`](https://docs.python.org/3/library/os.path.html#os.path.relpath) with the workspace as a base, or something similar.
* The call to the playbook failed silently. There should be an error printed in the output of `linchpin up`.
* linchpin did log the error, but at the INFO level. Since this is an error, is should be logged at the ERROR level. I'm not sure where this gets handled, but it's raised an [an exception](https://github.com/ansible/ansible/blob/7b197d823ee775b0138495bcd3954dff28851dd5/lib/ansible/parsing/dataloader.py#L385-L387) by ansible.

# playbook with relative path and explicit root

I remembered the "path" option from yesterday's demo, so I tried that next.

```yaml
cinch:
  topology: cinch.yml
  layout: cinch.yml
  hooks:
    postup:
      - name: cinchup
        type: shell
        context: false
        actions:
          - echo Running postup
      - name: cinchup_ansible
        type: ansible
        path: /path/to/cinch/
        actions:
          - playbook: site.yml
    predestroy:
      - name: cinchdestroy
        type: shell
        context: false
        actions:
          - echo Running predestroy
      - name: cinchdestroy_ansible
        type: ansible
        path: /path/to/cinch/
        actions:
          - playbook: teardown.yml
```

This does find and run the cinch playbook, but errors are raised:
```
TASK [check_ssh : construct command line for SSH check] ************************
fatal: [redacted]: FAILED! => {"failed": true, "msg": "the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'inventory_file' is undefined\n\nThe error appears to have been in '/path/to/cinch/roles/check_ssh/tasks/main.yml': line 7, column 5, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n  - name: construct command line for SSH check\n    ^ here\n"}
```

This is surprising, since I'd expect the hook definition above to be little more than a convenient way to get linchpin to run `ansible-playbook -i inventories/cinch.inventory /path/to/cinch/site.yml`. This led to my next experiment.

# shelling out to ansible-playbook

```yaml
cinch:
  topology: cinch.yml
  layout: cinch.yml
  hooks:
    postup:
      - name: cinchup
        type: shell
        context: false
        actions:
          - echo Running postup
          - ansible-playbook -i inventories/cinch.inventory /path/to/cinch/site.yml
    predestroy:
      - name: cinchdestroy
        type: shell
        context: false
        actions:
          - echo Running predestroy
          - ansible-playbook -i inventories/cinch.inventory /path/to/cinch/teardown.yml
```

This works perfectly, though the ansible output is  swallowed up by linchpin until the ansible play completes, which doesn't look great. I'm curious about what the ansible action type might be doing differently than ansible-playbook itself.

# Revisting relative paths

Since I now know that if `path` is not set, the playbook location is relative to the workspace, I plugged in relative paths:

```yaml
cinch:
  topology: cinch.yml
  layout: cinch.yml
  hooks:
    postup:
      - name: cinchup
        type: shell
        context: false
        actions:
          - echo Running postup
      - name: cinchup_ansible
        type: ansible
        actions:
          - playbook: ../../../../../../../cinch/site.yml
    predestroy:
      - name: cinchdestroy
        type: shell
        context: false
        actions:
          - echo Running predestroy
      - name: cinchdestroy_ansible
        type: ansible
        actions:
          - playbook: ../../../../../../../cinch/teardown.yml
```

This fails silently, similar to when an absolute path was used, but in this case it's because ansible does not load the libraries contained in the cinch dir, again pointing to differences in how linchpin runs the playbook compared to how ansible itself runs the playbook.

linchpin/linchpin.conf
herlo: I think @samvarankashyap might be used to the fact that yaml required quotes around things while the ini style configuration does not. Additionally, the ini configuration functionality supports interpolation with ConfigParser, which is great to be able to reuse the variables in another variable.

samvarankashyap:
@seandst  Thank you for the detailed review. Will be working on the fixes and get you back.

samvarankashyap:
@seandst Except for logging error issue , rest i fixed it through recent commits. 
Could please verify the whether the recent commits fixed the problems with hooks you were facing .  
Meanwhile , I will be testing the same on my workspace and playbooks.

samvarankashyap:
@seandst : 
inventory_file is not provided as a extra_var , for ansible playbooks . Due to that you might be facing following error 
```
TASK [check_ssh : construct command line for SSH check] ************************
fatal: [redacted]: FAILED! => {"failed": true, "msg": "the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'inventory_file' is undefined\n\nThe error appears to have been in '/path/to/cinch/roles/check_ssh/tasks/main.yml': line 7, column 5, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n  - name: construct command line for SSH check\n    ^ here\n"}
```
try changing the playbooks to refer to linchpin_context["inventory_file"] it should work in following case 
## **playbook with relative path and explicit root**

seandst:
> @seandst :
> inventory_file is not provided as a extra_var , for ansible playbooks

`inventory_file` is a magic var[0] set by ansible based on the inventory file passed to `ansible-playbook` when it is run. Having to change a playbook that works just fine with `ansible-playbook` to do special stuff to work in linchpin seems unnecessary, and seems like an ansible antipattern. I think normal "ansible stuff" should work when using the ansible hook type.

As a hooks user, I'm more likely to just keep calling out to ansible-playbook directly in a shell task than I am to modify my playbooks to support linchpin-specific implementations of behavior normally provided by ansible.

[0]: http://docs.ansible.com/ansible/playbooks_variables.html#magic-variables-and-how-to-access-information-about-other-hosts

seandst:
Tested this again yesterday with the latest tip (f21dc927212fb2ee804e4a687c7c4e5a0288a906), no changes. The ansible hook fails with the `inventory_file` error, but shelling out to ansible-playbook works correctly.

linchpin/hooks/action_managers/ansible_action_manager.py
```diff
@@ -195,6 +195,8 @@ def execute(self):
             e_vars = action.get("extra_vars", {})
             extra_vars.update(e_vars)
             if self.context:
-                extra_vars["linchpin_context"] = self.get_ctx_params()
```
samvarankashyap: @herlo @seandst 
This should fix the inventory_file problem mentioned by @seandst . 
Now the inventory_file is passed as a extra_var parameter such that , it will be available for the playbooks .

seandst:
Testing on the latest tip, the ansible hook is throwing a warning:

`[WARNING]: Host file not found: /home/semyers/ve/cinch/cinch-docs-internal/examples/linch-pin-topologies/openstack-slave/inventories/cinch.inventory`

```
$ ls inventories/
cinch-test
```

So it looks like the inventory file created by linchpin doesn't match the inventory filename being passed to ansible by linchpin.

samvarankashyap:
@seandst could you please paste the topology you are using ?

linchpin/provision/roles/inventory_gen/tasks/main.yml
```diff
@@ -25,9 +25,14 @@
   set_fact:
     inventory_layout: "{{ layout_file | default( default_layouts_path+'/'+'openshift-3node-cluster.yml') | ordered_yaml }}"
 
+- name: "Generate topology_name"
+  set_fact:
+    topology_name: "{{ topology_file | strip_extension }}"
```
herlo: I don't think this is what we want to do. Rather, we want to base it upon the topology_name variable, just as it is for resources.

samvarankashyap:
@seandst : As per the discussion with @herlo , we came to a conclusion that a global shared context is necessary to handle different states and magic variables of ansible. However , currently the issue with inventory_file name mismatch is being fixed by additional function set_magic_vars . Which will be eventually be replaced by shared context among different components of linchpin. 
@seandst could you please verify your cinch playbook run and let us know if you face any issues.

linchpin/hooks/__init__.py
```diff
@@ -0,0 +1,232 @@
...
+        '''
+
+        # FIXME
+        # This statement makes no logical sense.
+        # Why would we prepare context parameters if the state isn't
+        # listed in the method? It doesn't appear as though that
+        # sets any related values in any way, shape, or form.
+        #self.prepare_ctx_params()
+
+        hooks_data = self.api.target_data.get('hooks', None)
```
herlo: This breaks when there are no hooks for a target. A simple fix would look like this::

  if hooks_data and hooks_data.has_key(str(state)):

linchpin/hooks/__init__.py
```diff
@@ -0,0 +1,232 @@
...
+
+        if hooks_data.has_key(str(state)):
```
herlo: This breaks when there are no hooks for a target. A simple fix would look like this::

```if hooks_data and hooks_data.has_key(str(state)):```

seandst:
The `inventory_file` problem seems resolved (thanks!), but I also had to set `inventory_dir` to get things working: https://github.com/samvarankashyap/linch-pin/pull/4

samvarankashyap:
@seandst : merged you pr

herlo:
@seandst How does this update look to you? I think we're nearly there, but wanted to get feedback on your latest runs against this code.

seandst:
I mentioned this in IRC but forgot to update the issue, my apologies:

I looks like there's still at least one dangling reference to default_inventories_path, maybe. I'm seeing this error: `the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'default_inventories_path' is undefined` coming out of linchpin/linchpin/provision/roles/inventory_gen/tasks/main.yml, line 28, column 3.

samvarankashyap:
@seandst : fixed the reference with amend commit . Could you please check it again.

samvarankashyap:
@seandst ping !
Waiting on your testing , 
Hope we could merge this PR today .

Issue #216, opened 2017-04-24
Rewrite os_heat to remove python-openstackclient as dependency


samvarankashyap:
Python-openstack client  is responsible for frequent contextual version conflict in linchpin . 
python-openstack client is necessary for provisioning openstack heat templates.
In order to remove it, os_heat resource needs to rewritten in linchpin using ansible heat module.
Further, due to python openstack client, many unnecessary packages are installed , like
```
pbr!=2.1.0,>=2.0.0 # Apache-2.0
six>=1.9.0 # MIT
Babel!=2.4.0,>=2.3.4 # BSD
cliff>=2.3.0 # Apache-2.0
keystoneauth1>=2.18.0 # Apache-2.0
openstacksdk>=0.9.14 # Apache-2.0
osc-lib>=1.2.0 # Apache-2.0
oslo.i18n>=2.1.0 # Apache-2.0
oslo.utils>=3.20.0 # Apache-2.0
python-glanceclient>=2.5.0 # Apache-2.0
python-keystoneclient>=3.8.0 # Apache-2.0
python-novaclient>=7.1.0 # Apache-2.0
python-cinderclient>=2.0.1 # Apache-2.0
```
Resolving this issue would definitely reduce the disk footprint and version conflicts of linchpin .

herlo:
@samvarankashyap You previously mentioned the ansible heat module was not sufficient for our needs. I assume this has now changed?

samvarankashyap:
@herlo I believe now we can rewrite linchpin heat resource with ansible module without effecting the existing topology schema.

herlo:
Ansible is moving the os_server module toward python-openstackclient. In further discussions, we've decided to start separating out each of the providers into their own packages. This will help alleviate dependencies, and allow users freedom of choice on which providers to use. Closing this ticket.

Issue #204, opened 2017-04-11
(1) Relax dependency pinning as much as possible


greg-hellings:
Some packages are pinned with absolute versions (e.g. beaker-client==23.3, ansible==2.2.1.0, etc). These should be relaxed as much as possible, hopefully down to >= dependencies.

Packaging linchpin for distros will feature interruptions to usability if exact versions are pinned in places and then those update from the underlying package manager.

samvarankashyap:
Ansible has recently released , 2.3 on pypi . I need check whether linchpin works fine w.r.to 2.3 release.

herlo:
![](https://github.trello.services/images/mini-trello-icon.png) [Reduce Dependencies / Multiple Packages](https://trello.com/c/e9Yr6e1R/16-reduce-dependencies-multiple-packages)

PR #219, opened 2017-04-26
Rewrite of os_heat resource provisioning in linchpin

samvarankashyap:
Summary:
- Updated provision_os_heat playbook to use ansible os_stack module
- Updated requirements.txt

Fixes: #216 #204


samvarankashyap:
https://github.com/ansible/ansible/issues/21770

samvarankashyap:
@herlo : os_stack ansible module does not work . 
However , we can drop the support for os_heat until the issue is fixed by ansible. and merge this PR for resolving #204 
any thoughts ?

herlo:
Instead of merging this to resolve #204, let's drop this PR altogether. You can reraise the fix to #204 in its own PR, and keep this tree without that fix for when ansible/ansible#21770 gets addressed. 

Sound fair?

PR #220, opened 2017-04-26
Enhancements for Python compatibility

rywillia:
This commit improves linch-pins compatibility with Python 2 and Python 3. Changes consist of fixing import statements, checking if dictionary objects have attributes and using built in functions. More details can be found in each commit.


herlo:
I want to wait a bit to merge this, but I like what I see here!

Issue #217, opened 2017-04-25
Issues to use linchpin API. lp_up and lp_destroy do not return results. No way to control console setting. Returned results are only for OK tasks.


smatula:
Trying to use linchpin through API, not able to set console to false and no results are returned.
To overcome this temporarily override the following methods in LinchpinAPI: run_playbook, lp_up and lp_destroy. Also, any other front facing calls that call run_playbook to have the ability to set console and should return results. 

Also, discovered through testing, the results returned by callback are only OK results. Failures should also be returned. Code snippet for class PlaybookCallback also below.
 
Code snippets of override code:
```
    def lp_up(self, pinfile, targets='all', console=True):
        return(self.run_playbook(pinfile, targets, playbook="provision", console=console))

    def lp_destroy(self, pf, targets, console=True):
        return(self.run_playbook(pf, targets, playbook="destroy", console=console))

    def run_playbook(self, pinfile, targets='all', playbook='provision', console=True):
        results = []
        pf = yaml2json(pinfile)

        # playbooks check whether from_cli is defined
        # if not, vars get loaded from linchpin.conf
        self.ctx.evars['from_cli'] = True
        self.ctx.evars['lp_path'] = self.lp_path

        self.ctx.evars['default_resources_path'] = '{0}/{1}'.format(
            self.ctx.workspace,
            self.ctx.cfgs['evars']['resources_folder'])
        self.ctx.evars['default_inventories_path'] = '{0}/{1}'.format(
            self.ctx.workspace,
            self.ctx.cfgs['evars']['inventories_folder'])

        self.ctx.evars['state'] = "present"

        if playbook == 'destroy':
            self.ctx.evars['state'] = "absent"

        # checks wether the targets are valid or not
        if set(targets) == set(pf.keys()).intersection(targets) and len(targets) > 0:
            for target in targets:
                self.ctx.log_state('{0} target: {1}'.format(playbook, target))
                topology = pf[target]['topology']
                topology_registry = pf.get("topology_registry", None)
                self.ctx.evars['topology'] = self.find_topology(pf[target]["topology"],
                                                                topology_registry)
                if pf[target].has_key("layout"):
                    self.ctx.evars['layout_file'] = (
                        '{0}/{1}/{2}'.format(self.ctx.workspace,
                                             self.ctx.cfgs['evars']['layouts_folder'],
                                             pf[target]["layout"]))


[...]

PR #221, opened 2017-04-26
More context and API adjustments (plus tests!)

herlo:
Per PR #217, linchpin API now returns values, and allows control of ansible console output, either via the context.verbose flag, or via the linchpin.conf ansible.console = True option.

Additionally, context.py was moved to the cli package, making room for a context provided by the API, or by the UI developer. 

Added a console logger. This allows logging out to stderr without intervention. By default, the log_state and the Context.verbose flags will log messages to stderr. 

Reworked the test layout structure some, to allow for pass/fail type modules. Essentially, the pass module will contain tests that are expected to pass, while the fail module, well expects tests to fail (giving a success). Following @samvarankashyap's lead with mockdata in the tests, I created mockdata/context.py (will be moved eventually to mockdata/cli/context.py) which provides basic data to test the functionality in linchpin/cli/context.py module. Currently, only the pass tests have been written.

Updated to version 1.0.0a4


linchpin/api/__init__.py
```diff
@@ -11,111 +17,165 @@ class LinchpinError(Exception):
...
 
 
-        # checks wether the targets are valid or not
+        # checks whether the targets are valid or not
         if set(targets) == set(pf.keys()).intersection(targets) and len(targets) > 0:
             for target in targets:
-                self.ctx.log_state('{0} target: {1}'.format(playbook, target))
+                self.ctx.log_state('target: {0}, action: {1}'.format(target, playbook))
                 topology = pf[target]['topology']
                 topology_registry = pf.get("topology_registry", None)
```
samvarankashyap: @herlo  : topology_registry is not used anywhere, am i missing something here?

samvarankashyap: @herlo : topology_registry _ variable is not used anywhere ! am i missing something ?

samvarankashyap:
@herlo https://github.com/CentOS-PaaS-SIG/linch-pin/pull/221/files#r113714735

PR #222, opened 2017-04-27
Develop :: Added support for security_groups parameter in os_server resource 

samvarankashyap:
Surprisingly , though linchpin can create security groups on openstack , it cannot assign security groups to the provisioned instances. 
Current pr resolves this issue by adding security_groups parameter to os_server resource which accepts comma separated values of security groups in topology.

PR #223, opened 2017-04-27
Master : Added support for security_groups parameter in os_server resource ref PR #222

samvarankashyap:
Changes in PR #222 adapted to master branch.
Ref #222


herlo:
When develop branch gets put into master, this will land. Not before.

PR #227, opened 2017-05-02
Allow multiple targets again

herlo:
Fix bug that was returning before finishing a loop on multiple targets

### DIFF
--- a/linchpin/api/__init__.py
+++ b/linchpin/api/__init__.py
@@ -3,13 +3,13 @@
 import os
 from collections import namedtuple
 
-from linchpin.api.callbacks import PlaybookCallback
 from ansible.inventory import Inventory
 from ansible.vars import VariableManager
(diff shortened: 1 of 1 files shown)


herlo:
@smatula If you have a moment, feel free to review this PR. I've added you to our collaborators group, so you should be able to give feedback.

linchpin/api/__init__.py
```diff
@@ -145,6 +142,7 @@ def run_playbook(self, pinfile, targets='all', playbook='up'):
                                                 console=self.console)
```
smatula: Should it not have a dictionary of results in the above block also?  Looks like this could also have multiple targets also. I would think length of 0 below is for all. This block is for 1 or more.

herlo: @smatula I've updated the PR as you mentioned. I agree, this is better. LMK what you think.

PR #228, opened 2017-05-02
Add failed results to callbacks.py

herlo:
Addresses API calls which do not use the console, The callback now
stores failed results so they can be returned.


herlo:
@smatula If you have a moment, feel free to review this PR. I've added you to our collaborators group, so you should be able to give feedback.

PR #234, opened 2017-05-10
whiteboard + hostrequires beaker support, with docs and examples updates

### LINKED ISSUE #182: (2) Enhancement Request - Set Beaker Jobs Whiteboard value.
Have field in topology or auto generate name to be set as beaker jobs whiteboard value. Name should be a descriptive name that one can easily tell the linchpin provision job that started the beaker job. 

The whiteboard is a free-form string to describe the job or recipe. It looks like I the whiteboard for the recipe can be set using the following property:
         whiteboard: "<value>"
   under recipesets:

   In beaker whiteboard is one of the fields displayed when listing jobs. When linchpin provisions a beaker machine the whiteboard field is blank. 

I think the job whiteboard can be passed as an option when creating the beaker job.

### LINKED ISSUE #196: (3) Add or verify support for the 'hostrequire' feature of beaker.
This feature will allow a Beaker client to specify a list of hostnames which are required for the provisioning task.  This is useful for teams which have reserved a host in Beaker for exclusive use.  Here are the docs from the Beaker project:

https://beaker-project.org/docs/user-guide/job-xml.html

If this feature is currently supported, we should test it and add it to our example topology file.

seandst:
This PR adds support for the beaker job whiteboard to beaker topologies, and improves support for the beaker API's `hostRequires` filtering. Docs related to the hostrequires configuration have been added, and I've also updated the beaker examples.

closes #182
closes #196

### DIFF
--- a/linchpin/defaults/schemas/schema_v4.json
+++ b/linchpin/defaults/schemas/schema_v4.json
@@ -90,6 +90,9 @@
                  "job_group": {
                      "type": "string"
                  },
+                 "whiteboard": {
+                     "type": "string"
+                 },
                  "recipesets": {
                      "type": "array",
                      "items": {
--- a/linchpin/defaults/schemas/schema_v4.json
+++ b/linchpin/defaults/schemas/schema_v4.json
@@ -102,9 +102,6 @@
                                  "distro": {
                                      "type": "string"
                                  },
-                                 "arches": {
-                                     "type": "array"
-                                 },
                                  "arch": {
                                      "type": "string"
                                  },
--- a/linchpin/defaults/schemas/schema_v4.json
+++ b/linchpin/defaults/schemas/schema_v4.json
@@ -114,8 +114,30 @@
                                  "keyvalue": {
                                      "type": "array"
                                  },
-                                 "hostrequire": {
-                                     "type": "array"
+                                 "hostrequires": {
+                                     "type": "array",
+                                     "items": {
+                                         "oneOf": [
+                                             {
+                                                 "type": "object",
+                                                 "properties": {
+                                                     "force": {
+                                                         "type": "string"
+                                                     }
+                                                 },
+                                                 "required": ["force"]
+                                             },
+                                             {
+                                                 "type": "object",
+                                                 "properties": {
+                                                     "tag": {
+                                                         "type": "string"
+                                                     }
+                                                 },
+                                                 "required": ["tag"]
+                                             }
+                                          ]
+                                     }
                                  },
                                  "bkr_data": {
                                     "type": "object"
--- a/docs/source/topologies_beaker.rst
+++ b/docs/source/topologies_beaker.rst
@@ -17,6 +17,7 @@ Beaker Server
       - resource_group_name: test1
         res_group_type: beaker
         job_group: ci-ops-central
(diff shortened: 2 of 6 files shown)


linchpin/defaults/schemas/schema_v4.json
```diff
@@ -108,8 +108,35 @@
...
-                                 "hostrequire": {
-                                     "type": "array"
+                                 "hostrequires": {
+                                     "type": "array",
+                                     "items": {
+                                         "not": {
+                                             "allOf": [
+                                                 {"required": ["force", "tag"]}
+                                             ]
+                                         },
```
seandst: This schema bit is attempting to reflect the mutually exclusive relationship between the use of `force:` under `hostrequires`, and the use of `tag:`, since using `force:` means that any `tag:` filters are ignored by beaker. I have no idea if it works, and would love some pointers on simple ways to actually test this scheme validation against the topologies I've written during the development and testing of this PR.

herlo: We have written a schemacheck.yml, included at the same level as site.yml, which you could use to check your schema. You use it something like this: 

```bash
ansible-playbook -vvvv linchpin/provision/schemacheck.yml -e \ 
"topology='/home/herlo/Projects/linchpin/linchpin/examples/topologies/dummy_cluster.yml'" \ 
-e 'schema=../defaults/schemas/schema_v4.json'
```

herlo: @samvarankashyap any opinions on this?

seandst: schemacheck looks perfect, I'll run some examples through the schema and see how I did. :)

seandst: Alright, "bad" values are rejected, but the mutual exclusion stuff doesn't work. I'll fix it up.

samvarankashyap:
@seandst @herlo : Looks good to me 
@seandst Is there any specific reason you used allof line 115 of schemav4 instead of oneof ?

linchpin/defaults/schemas/schema_v4.json
seandst: > Is there any specific reason you used allof line 115 of schemav4 instead of oneof ?

It's a under a "not", and I'm trying to express than you can have anyOf objects with a "tag" key or objects with a "force" key, and also that you can not have allOf objects with a "tag" key and objects with a "force" key. It only makes sense to use "force" once, but it's acceptable to have multiple tags.

oneOf might do what I want here, but I think I'm going to break this out into its own issue, and pull the mutual exclusion stuff out of this PR. The included docs for beaker topologies, at least, explain that any included "tag" objects will be ignored if "force" is used.

seandst:
Alright, I removed the mutual exclusion validation piece. "bad" objects under hostrequires are indeed rejected as invalid, which I think it good enough for me. If anyone thinks there's value in adding the mutual exclusion to linchpin validation I'll open an issue to deal with it. In hindsight, I don't think it adds value, because both the beaker client and the beaker API allow it, and just ignore any other host filters if "force" is set. Since, in my mind, the validation's main goal is to ultimately ensure we're sending good data to providers, the schema I've added is good enough.

PR #206, opened 2017-04-18
Rework CLI with a big emphasis on configuration

herlo:
Clean up the linchpin command line interface
------------------------------------------------------------
The linchpin command now provides a simpler interface. The topology and layout functionality has been removed temporarily to ensure a better approach. Below are the current options.

    linchpin
    Usage: linchpin [OPTIONS] COMMAND [ARGS]...

      linchpin: hybrid cloud orchestration

    Options:
      -w, --workspace PATH  Uses the familiar Jenkins $WORKSPACE environment
                            variable
      -v, --verbose         Enable verbose output
      --version             Prints the version and exits
      -h, --help            Show this message and exit.

    Commands:
      init     Initializes a linchpin project.
      up       Provisions nodes from the given target(s) in...
      destroy  Destroys nodes from the given target(s) in...

#########

    linchpin init --help
    Usage: linchpin init [OPTIONS]

      Initializes a linchpin project, which generates an example PinFile, and 
      creates the necessary directory structure for topologies and layouts.

    Options:
      -h, --help  Show this message and exit.

linchpin up (linchpin rise still works)
############################

    linchpin up  --help
    Usage: linchpin up [OPTIONS] TARGET

      Provisions nodes from the given target(s) in the given PinFile.

      TARGET ...
      Provision ONLY the listed target(s).

    Options:
      -p, --pinfile TEXT  Use a name for the PinFile different from the
                          configuration.
      -h, --help          Show this message and exit.

linchpin destroy (linchpin drop still works)
#################################

   linchpin destroy --help
    Usage: linchpin destroy [OPTIONS] TARGET

      Destroys nodes from the given target(s) in the given PinFile.

      TARGET ...
      Destroy ONLY the listed target(s).

    Options:
      -p, --pinfile TEXT  Use a name for the PinFile different from the
                          configuration.
      -h, --help          Show this message and exit.

Additional framework has been put in place for `linchpin down` (currently not implemented).

Standardized vars in roles and linchpin.conf
-------------------------------------------
vars (and extra-vars) are now named consistently throughout the
playbooks and configuration. This is an extension of the linchpin api
functionality which now mostly exists. Here are the new variables with
their previous counterpart(s), with explanations as necessary.

INPUTS
======

The following extra-vars can be passed on the command line with
`ansible-playbook`, using the -e or --extra-vars option.

* topology / topology_file (not changed) - YAML definition
for provisioning and teardown and which providers to use.
* schema / schema_file (schema_v3, schema_v4 are still available) - JSON
[...]

PR #249, opened 2017-05-30
Document ansible config & extra vars

### LINKED ISSUE #230: (3) Document ansible vars from PR #206
The description of https://github.com/CentOS-PaaS-SIG/linch-pin/pull/206#issue-222544291 (starting with the "INPUTS" heading) contains docs related to the ansible vars used by linchpin that might be useful to users, and should probably be included in the docs on the develop branch.

seandst:
These are the remaining docs from #206, ReST-ified.

closes #230

### DIFF
--- a/docs/source/config.rst
+++ b/docs/source/config.rst
@@ -10,3 +10,4 @@ to specify the resources required.
     config_general
     config_topologies
     config_layout
+    config_ansiblevars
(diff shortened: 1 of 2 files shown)


seandst:
One thing is already out of date, which (iirc) is that `default_inventories_path` is now `inventory_dir`. Are there other vars that we know of that need to be updated as a result of PRs merged since #206?

PR #291, opened 2017-06-22
Add userdata support to OpenStack

greg-hellings:
Add support for cloud-init's userdata argument to OpenStack
provisioning


greg-hellings:
@samvarankashyap The user can always do that with Ansible's lookup parameter. There's no need to make that the only behavior as, for most simple cases that would just complicate matters.

PR #312, opened 2017-07-19
remove redundancy when setting workspace

tvieira:
The workspace is set through the following order:
  1. by default, LinchpinContext class sets to current directory
  2. then if the environment variable WORKSPACE, it replaces the
     default
  3. last with most precedence, if the user sets the option -w or
     --workspace
There is no need to set the workspace to PWD in case the user has
not set the option or the environment variable because the class
already sets to os.path.curdir (step 1 above).


tvieira:
Sorry, ignore this PR. My code base wasn't updated.

PR #320, opened 2017-07-24
Create getter/setter for workspace

herlo:
* Simplify the access to the workspace, but have it use python get/set methods
* Remove cli/utils.py as it has no uncommented code
* Clean up some spacing issues


seandst:
`+76 −275`: I ❤️ deduplication

tvieira:
@samvarankashyap did you tested without setting the environment variable 'WORKSPACE' - this code doesn't see to initialise self.workspace if the WORKSPACE is not set.

tvieira:
Looking further into the code, I don't think it would protect the run of failure if user does not pass `--workspace` **and** if the environment variable `WORKSPACE` is not set.

The way it was before, having it by default set to the current directory where the client is being executed you protect from failure runtime failure. I vote for the error to come because the files that should be in the workplace directory are not found and the client handles a known error rather than have an exception for using an empty variable during runtime.

What do you think?

herlo:
@tvieira Can you look at #325? I've corrected the workspace issue there. I think you will find the fix to your satisfaction. Thanks!

PR #324, opened 2017-07-26
Bugfix: issue #323 set default workspace if not provided

### LINKED ISSUE #323: Regression in latest source code: linchpin init doesn't not work
Installed the source on centos machine:

```
linchpin --version 
linchpin version 1.0.2 # build from source
linchpin init
ctx.workspace: None
Error: mkpath: 'name' must be a string (got None)
```

looks like linchpin doesn't get its default workspace if not mentioned.

samvarankashyap:
Fixes #323

### DIFF
--- a/linchpin/__init__.py
+++ b/linchpin/__init__.py
@@ -122,6 +122,8 @@ def runcli(ctx, config, workspace, verbose, version, creds_path):
 
     if workspace is not None:
         ctx.workspace = os.path.realpath(os.path.expanduser(workspace))
+    else:
+        ctx.workspace = os.getenv('PWD')
 
     ctx.log_debug("ctx.workspace: {0}".format(ctx.workspace))
 


linchpin/__init__.py
```diff
@@ -122,6 +122,8 @@ def runcli(ctx, config, workspace, verbose, version, creds_path):
 
     if workspace is not None:
         ctx.workspace = os.path.realpath(os.path.expanduser(workspace))
+    else:
```
herlo: I've been sitting here debating this option, I don't think this is the best place for this option.  After thinking and reviewing more, I see what I missed and what needs to happen. Essentially, the `linchpin init` functionality needs to be moved from linchpin/__init__.py to linchpin/cli/__init__.py. This will allow the workspace to be set correctly, as it inherits from the api/__init__.py. Let me get a fix in place, but I don't think I want to accept this fix.

tvieira: This would revert back the PR I sent last week, #313. Before that the PR #320 has already broken because it removes the self.workspace from the LinchpinContext class, once the first default setting for this variable. :(

herlo: @tvieira #320 does not remove the self.workspace setting from LinchPinContext. Can you maybe clarify what you mean wrt this? I'm reworking a few things here to address the bug above, so I'd like your input. Thanks!

tvieira: I have checked the latest code and I can't find the setting of workplace in the initialisation of the LinchpinContext at https://github.com/CentOS-PaaS-SIG/linchpin/blob/develop/linchpin/api/context.py#L35.

Further investigation shows that this commit https://github.com/CentOS-PaaS-SIG/linchpin/pull/320/commits/d8e9ba136492dd44bd65881851038868a53b7a9a removes this set.

If the line:
self.workspace = os.path.realpath(os.path.curdir)

is added back to the __init__ this would fix the this current issue.

PR #325, opened 2017-07-26
set_workspace and pinfile correctly

herlo:
Issue #323 indicated that the set_workspace pr #320 broke when the workspace wasn't passed from the command line. This code fixes that issue.

During the work, it was discovered that setting the workspace affected the pinfile, which was done in the init function of runcli. To address this, the lp_init method was added to the LinchpinCli class. Additional code was moved around to accommodate this change.

This fix is more involved than #324, but better addresses the separation of the cli from its python classes. The workspace and pinfile are now set in the cli and api classes for reuse throughout a call.

### DIFF
--- a/linchpin/__init__.py
+++ b/linchpin/__init__.py
@@ -6,7 +6,6 @@ import click
 import shutil
 import logging
 
-from distutils import dir_util
 from jinja2 import Environment, PackageLoader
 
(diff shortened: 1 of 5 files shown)


herlo:
tagging @tvieira for review as well.

tvieira:
LGTM

Issue #328, opened 2017-07-27
Break out gce2 into proper ansible module


herlo:
It seems we bundle some libraries in linchpin that should really be bundled in another package. In this case, Ansible modules, should go into Ansible. This issue is to track getting gce2 moved over.

samvarankashyap:
@herlo gce2 is module is copied down from ansible 2.3
If ansible 2.3 is installed we can change the playbooks to use gce module to run instead of gce2

herlo:
@samvarankashyap and @herlo discussed this and agreed 2.3.1+ is a better direction. Core providers should in all cases support ansible 2.3.1+. Tests need to be written to accommodate each provider.

PR #336, opened 2017-08-01
Move ansible to >=2.3.1 in requirements.txt

herlo:
This seems to work in my initial testing.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ click
 tabulate
 camel
 yamlordereddictloader
-ansible==2.2.1
+ansible>=2.3.1
 python-openstackclient
 shade>=1.8.0,!=1.21.0
 naked


samvarankashyap:
@herlo Just tested openstack provisioning using ansible2.3 it works. 
But to be cautious wouldn't it be better if we use >=2.2.2 as its stable and minimum requirement for most of the projects .

herlo:
I am fine with having 2.2.2, but this then means in #328 your comment would not be supported in every case.  I'm keen to support the CORE providers. That's libvirt, openstack, Google Cloud (gce), and Amazon Web Service (aws_ec2). Others can break this standard, as we'll be breaking them out into their own linchpin modules in the future. 

Having Ansible be 2.3+ would provide better support for the core modules.

PR #343, opened 2017-08-15
Add example to provision single system using openstack

mrniranjan:
Add example clouds.yaml and topology file.

Signed-off-by: Niranjan M.R <mrniranjan@redhat.com>

### DIFF
--- a/docs/source/topologies_openstack.rst
+++ b/docs/source/topologies_openstack.rst
@@ -245,7 +245,73 @@ Openstack Full Stack
         test_var2: "test_var2 msg is grp2 hello"
         test_var3: "test_var3 msg is grp2 hello"
 
-.. note::
 
-  Source of the above mentioned examples can be found at `Example Topologies <https://github.com/CentOS-PaaS-SIG/linch-pin/tree/master/ex_topo>`_
+Steps to provision Single Host
+==============================
+
+.. contents:: Topics
+
+.. _Steps to provision Single Host:
+
+Credentials
+```````````
+* save openstack credentials in standard ``clouds.yml`` file using below
+  format and save the directory path containing clouds.yml in environment variable ``CREDS_PATH``.
+
+.. code-block:: yaml
+
+   
+   ---
+    clouds: 
+      devstack:
+        auth:
+          username: "admin"
+          password: "Secret123"
+          project_name: "my-tenant"
+          auth_url: "http://192.168.122.33:5000/v2.0"
+          
+Topology
+````````
+* create topology file under ``$WORKSPACE/topologies/openstack_topology.yml``
+  as show below:
+
+.. code-block:: yaml
+
+   ---
+    topology_name: "osp-test"
+    resource_groups:
+      -
+        resource_group_name: "lp-test"
+        resource_group_type: "openstack"
+        resource_definitions:
+          - name: "test1"
+            type: "os_server"
+            flavor: "m1.small"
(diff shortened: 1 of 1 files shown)


docs/source/topologies_openstack.rst
```diff
@@ -245,6 +245,66 @@ Openstack Full Stack
...
+      devstack:
+        auth:
+          username: "admin"
+          password: "Secret123"
+          project_name: "my-tenant"
+          auth_url: "http://192.168.122.33:5000/v2.0"
+          
+Topology
+````````
+* create topology file under ```$WORKSPACE/topologies/openstack_topology``` as show below:
```
herlo: the file should end in `.yml` here.

docs/source/topologies_openstack.rst
```diff
@@ -245,6 +245,66 @@ Openstack Full Stack
...
+Steps to provision Single Host
+==============================
+
+.. contents:: Topics
+
+.. _Steps to provision Single Host:
+
+Credentials
+```````````
+   * save openstack credentials in ```$WORKSPACE/credentials/clouds.yml``` file using below format
```
herlo: The path to credentials is variable, as you may have understood from below where you passed `--creds-path`. Consider instead of `$WORKSPACE/credentials/clouds.yml`, using something more generic. Like `/path/to/cloud_credentials.yml`.  Additionally, this setting is optional. Meaning the user could instead use the standard clouds.yaml location, or set the env variable `$CREDS_PATH`. I think it would be very helpful to point out that this is optional.

docs/source/topologies_openstack.rst
```diff
@@ -245,6 +245,66 @@ Openstack Full Stack
...
+
+Provision
+`````````
+
+* provision the above topology
+      
+.. code-block:: bash
+   
+   $ cd $WORKSPACE
+   $ linchpin --creds-path  ./credentials/ up
```
herlo: It would be helpful to note that --creds-path is completely optional. If the user were to use the standard locations, environment variables, or the `$CREDS_PATH` environment variable, the user would not need to pass this. As stated above, it may behoove you to adjust this to explain these options.

docs/source/topologies_openstack.rst
```diff
@@ -245,6 +245,66 @@ Openstack Full Stack
...
+* provision the above topology
+      
+.. code-block:: bash
+   
+   $ cd $WORKSPACE
+   $ linchpin --creds-path  ./credentials/ up 
+
 .. note::
 
   Source of the above mentioned examples can be found at `Example Topologies <https://github.com/CentOS-PaaS-SIG/linch-pin/tree/master/ex_topo>`_
```
herlo: Can you remove this line? It's completely incorrect now. I know that you are just updating, but it will mislead a user. Thanks!

docs/source/topologies_openstack.rst
```diff
@@ -245,7 +245,73 @@ Openstack Full Stack
...
+* provision the above topology
+      
+.. code-block:: bash
+   
+   $ cd $WORKSPACE
+   $ export CREDS_PATH="/path/to/credential_dir/"
+   $ linchpin -v up 
+
  
+* Alternatively one could pass credentials path as an argument to lichpin
```
herlo: Linchpin is misspelled here.

mrniranjan: sorry fixing it now,

Issue #306, opened 2017-07-12
(3) CLI does not exit with non zero when playbook execution fails


rywillia:
I am running Linchpin version 1.0.1 and the exit code when my Ansible playbook fails is 0. I would expect this code to be 1 or another integer value. Having this would be great for automation being able to determine if my command executed successfully or not.

```
--snippet--
TASK [openstack : Initiating  Provisioning/Deprovioning of resources Openstack resource group] ***
included: /usr/lib/python2.7/site-packages/linchpin/provision/roles/openstack/tasks/provision_resource_group.yml for localhost

TASK [openstack : Unset the authvar from previous run] *************************
ok: [localhost]

TASK [openstack : set default_cred_profile when res_grp[credentials] is undefined] ***
skipping: [localhost]

TASK [openstack : set default_cred_profile when res_grp[credentials] is undefined] ***
skipping: [localhost]

TASK [openstack : set cred_profile when res_grp[credentials] is defined] *******
ok: [localhost]

TASK [openstack : set default_cred_filename when res_grp[credentials] is defined] ***
ok: [localhost]

TASK [openstack : Get creds from auth driver] **********************************
changed: [localhost]

TASK [openstack : set auth_var] ************************************************
ok: [localhost]

TASK [openstack : provisioning resource definitions of current group] **********
included: /usr/lib/python2.7/site-packages/linchpin/provision/roles/openstack/tasks/provision_res_defs.yml for localhost

TASK [openstack : provision resources of type os_server] ***********************
included: /usr/lib/python2.7/site-packages/linchpin/provision/roles/openstack/tasks/provision_os_server.yml for localhost

TASK [openstack : provisioning os_server resources with provided auth] *********
fatal: [localhost]: FAILED! => {"changed": false, "extra_data": null, "failed": true, "msg": "Timeout waiting for the floating IP to be attached."}
	to retry, use: --limit @/usr/lib/python2.7/site-packages/linchpin/provision/site.retry

PLAY RECAP *********************************************************************
localhost                  : ok=18   changed=1    unreachable=0    failed=1   

hook postup initiated
[root@91b51ebff31f linchpin]# echo $?
0
```
I noticed there was a similar issue to this #159. This bug may still exist for certain use cases in the latest release?

herlo:
[...]

PR #344, opened 2017-08-15
CLI now exits with non-zero exit code on failure

agharibi:
This is a patch to issue #306.
I changed return value of run_playbook to a tuple which includes a 'results' dict and return_code to pass into _handle_results().  _handle_results() then calls the sys.exit() with the return_code argument.

### DIFF
--- a/linchpin/__init__.py
+++ b/linchpin/__init__.py
@@ -46,7 +46,7 @@ class LinchpinAliases(click.Group):
         return rv
 
 
-def _handle_results(ctx, results):
+def _handle_results(ctx, results, return_code):
     """
     Handle results from the Ansible API. Either as a return value (retval)
     when running with the ansible console enabled, or as a list of TaskResult
@@ -60,7 +60,6 @@ def _handle_results(ctx, results):
         The dictionary of results for each target.
     """
 
-    retval = 0
 
     for k, v in results.iteritems():
         if not isinstance(v, int):
@@ -73,10 +72,8 @@ def _handle_results(ctx, results):
                     msg = tr._check_key('msg')
                     ctx.log_state("Target '{0}': {1} failed with"
                                   " error '{2}'".format(k, tr._task, msg))
-                    sys.exit(retval)
-        else:
-            if v:
-                sys.exit(v)
+
+    sys.exit(return_code)
 
 
 @click.group(cls=LinchpinAliases,
@@ -170,9 +167,9 @@ def up(ctx, targets):
     pf_w_path = _get_pinfile_path()
 
     try:
-        results = lpcli.lp_up(pf_w_path, targets)
+        return_code, results = lpcli.lp_up(pf_w_path, targets)
 
-        _handle_results(ctx, results)
+        _handle_results(ctx, results, return_code)
 
     except LinchpinError as e:
         ctx.log_state(e)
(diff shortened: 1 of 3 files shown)


linchpin/api/__init__.py
```diff
@@ -338,7 +338,7 @@ def run_playbook(self, pinfile, targets='all', playbook='up'):
             if 'post' in self.pb_hooks:
                 self.hook_state = '{0}{1}'.format('post', playbook)
 
-        return results
+        return return_code, results
```
herlo: As a rule, when returning, use parentheses `eg. (return_code, results)` around the items being returned when they are grouped.

linchpin/__init__.py
```diff
@@ -73,10 +72,8 @@ def _handle_results(ctx, results):
                     msg = tr._check_key('msg')
                     ctx.log_state("Target '{0}': {1} failed with"
                                   " error '{2}'".format(k, tr._task, msg))
-                    sys.exit(retval)
-        else:
-            if v:
-                sys.exit(v)
+
+    sys.exit(return_code)
```
herlo: There are scenarios where performing a sys.exit here is not the desired case. Adjusting to this means linchpin exits here no matter what. If that's the case, then there is no reason to perform exits elsewhere in the code. For instance, if the `return_code` is 0, but there will be an error provisioning only one node of 3, this may not error out in the future Linchpin may print this out and continue on through the process.

Because of some code,  I recently had to update this as well. Here's how I implemented the changes. I'm not sure they are perfect yet, but they work well.

```
def _handle_results(ctx, results, return_code=1):
    ..snip..
    for k, v in results.iteritems():
        if not isinstance(v, int):
            trs = v

            if trs is not None:
                trs.reverse()
                tr = trs[0]
                if tr.is_failed():
                    msg = tr._check_key('msg')
                    ctx.log_state("Target '{0}': {1} failed with"
                                  " error '{2}'".format(k, tr._task, msg))
                    sys.exit(return_code)
        else:
            if v:
                sys.exit(v)
```
One thing to consider when updating an existing function is that adding a new argument may affect callers from other classes, functions, etc. In that case, it's a good idea to make it a kwarg and default it to some value. In that way, it doesn't affect other callers that haven't updated the function.

Additionally, I'm curious why you remove the last 3 lines. Maybe that's why it's not returning the proper error code? In those cases though, the function should only return errors, I believe. Can you verify?

agharibi:
@herlo I like your implementation but there is one part missing that does not handle the -v cases. When -v is passed, I noticed in my testing that there would be one item in the dict that was the following: `{target:None}`. Therefore, when the following line is executed, `if trs is not None`, the condition is not met since `trs` is `None`.  So to fix that, I just added the condition to check if the return code is not zero. I added the following condition:
```
elif return_code:
    sys.exit(return_code)
```
I did not think that this would cause any particular harm in the future, but maybe you might think that this will lead to a bug down the road. If you could give me input on this, it would be most appreciated.

herlo:
More review of this and I think this extra bit of code causes unclear results. Specifically because if no `return_code` is passed in, it will be set to 1 (which is an error on the cli) by default. In cases where there is not any TaskResultSets (trs), the system will just exit as normal, usually with success. This is expected behavior as a `trs` value is always None when the `-v` option is passed to linchpin. 

I was able to prove this out by not passing an error code to the `_handle_results()` function from the `up()` command line function. (eg `_handle_results(ctx, results)`). Leaving in the `elif` showed a console error code of 1 (because it's defaulted). But, if we leave the elif off, it returns an error code of 0.

Alternatively, we could require the `return_code`, and since this is the implementation of the cli, no promises were made on functions here. I'm one to let the `return_code` value fall through. Though if you like the explicitness of having the `elif` in place, that's a simple adjustment.

agharibi:
@herlo I'm thinking to make return_code a mandatory field. I think it's necessary since every time Ansible runs, it will return an error code anyway. Is this a correct way of thinking or is there a reason that this should not be the case? I agree that I don't think that return_code should default to one and I also don't think that there is a case where return_code shouldn't be passed. Any feedback?

herlo:
This is probably ok. Submit the code and I'll do another review round. Thanks

PR #381, opened 2017-10-24
Provide for booting openstack from volume

greg-hellings:
Allows user to specify a specific volume to boot from, which will remain
intact after instance shut down. This volume must be pre-created and
already be flagged as a bootable volume

Allows user to request a new volume be automatically created on boot,
which will then be auto-terminated when the instance is shut down

Lacks the ability to check required/excluded options from schema. That
checking is handled by the Ansible module

### DIFF
--- a/linchpin/defaults/schemas/schema_v4.json
+++ b/linchpin/defaults/schemas/schema_v4.json
@@ -61,9 +61,18 @@
                 },
                 "userdata": {
                     "type": "string"
+                },
+                "boot_from_volume": {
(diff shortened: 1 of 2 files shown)


herlo:
I know the flake8 error is annoying. I've already fixed it in my branch. Please ignore. It passes otherwise.

samvarankashyap:
@greg-hellings @herlo LGTM

greg-hellings:
Shall I merge it, or should I let one of you do that?

PR #382, opened 2017-11-03
RunDB integration

herlo:
RunDB provides tooling for the following:

- A simple database (json based, called TinyDB) to store records of each transaction
- Each target is assigned its own table, and records are updated via the rundb ansible module
- Inputs and outputs are stored in a list of dictionaries. 
  - Inputs include topology_file and layout_file
  - Outputs store the results of the transaction, including return values from the ansible module
- An action key to help the destroy find the last record
- A unique-ish hash (called 'uhash') for naming resources as appropriate
- A start and end time for each run for reporting transaction length
- The return code from each provisioning action
- Each task is unique unless `--run-id <id>` is passed to the command line.
  - Passing the `--run-id` value makes an up action idempotent.

Currently, viewing the rundb data requires reading json. In the future, there will be a tool for getting values from any specific `run-id`, or the last few records. 

Additionally, at the end of any linchpin action (up or destroy), there is not currently a way to know the run_id or uhash. This can be seen by passing the `-v` option. However, it is planned to provide a 'report` at the end of each run.

Lastly, this does cause one challenge for destroy. Currently, the destroy action can only tear down one target per invocation when the `--run-id` option is passed. If it is not passed, the last record with an `up` action will be destroyed.


samvarankashyap:
@herlo Approved

PR #383, opened 2017-11-06
Feature addition  linchpin-libvirt cloud init 

samvarankashyap:
Feature addition summary:
1. Adds cloud init config parameters to be passed while booting an instance.
2. Adds remote_user parameter when booting machine on remote libvirt machines
3. Adds network_bridge parameters that helps users to specify the network bridge when its not defaulted to virbr0
4. Added copy_ssh_keys to inject default ssh keys to machines booted on libvirt
5. Added example topology to docs

Exhaustive example for all feature set as follows
```yaml
---
topology_name: "libvirt"
resource_groups:
  -
    resource_group_name: "ex"
    res_group_type: "libvirt"
    res_defs:
      - res_name: "fedoramachine"
        res_type: "libvirt_node"
        uri: "qemu:///system"
        remote_user: "root"
        count: 1
        driver: qemu
        #image_src: 'https://pubmirror2.math.uh.edu/fedora-buffet/alt/atomic/stable/Fedora-Atomic-25-20170705.0/CloudImages/x86_64/images/Fedora-Atomic-25-20170705.0.x86_64.qcow2'
        image_src: "file:///home/output/images/untested-atomic.qcow2"
        memory: 6144 
        vcpus: 3
        arch: x86_64
        copy_ssh_keys: true
        network_bridge: "virbr3"
        additional_storage: 15G
        cloud_config:
          users:
            - name: admin
              gecos: Admin User
              groups: wheel
              sudo: ALL=(ALL) NOPASSWD:ALL
              ssh-import-id: None
              lock_passwd: true
        networks:
          - name: default
```

Note: 
1. The above topology is tested on centos images and fedora_atomic images
2. Needs to be reviewed and merged after #382 is merged into develop branch
3. Needs more testing POST merging of #382 as plugging in runDB affects this current PR #383


herlo:
The remote_user is not needed in the topology as it can be passed in the uri. Please refer to https://libvirt.org/guide/html/Application_Development_Guide-Architecture-Remote_URIs.html. I suggest removing that from the schema, it may cause confusion.

samvarankashyap:
@herlo : I used remote_user for following case: 
Assume that you have a remote machine running libvirt and we would like to copy the image downloaded from local machine(machine running linchpin) to remote machine(machine running libvirtd) we might want to connect to remote machine via certain user. remote_user enables this functionality. 
 
However , I completely agree with your suggestion as I can extract remote_user information from uri itself with simple command or filter . I will add a commit doing the same.

herlo:
@samvarankashyap sure, that can be done, but wouldn't the remote_user be whatever is in ansible using become_user? I'm just curious how the use case for this works if you are copying an image from the local to remote (or vice versa). Are you not using ansible to do that work?

samvarankashyap:
@herlo : I have changed the installation scripts for libvirt_dependencies: 
One has to now install 
```
# install libvirt dependencies
sh scripts/install_libvirt_deps.sh
# install selinux in virtualenv
sh scripts/install_selinux_venv.sh
```
Example: topology used for testing: 
```yaml
---
topology_name: "libvirt"
resource_groups:
  -
    resource_group_name: "ex"
    res_group_type: "libvirt"
    res_defs:

      - res_name: "centos71"
        res_type: "libvirt_node"
        driver: "qemu"
        uri: 'qemu:///system'
        count: 1
        image_src: 'http://cloud.centos.org/centos/7/images/CentOS-7-x86_64-GenericCloud-1608.qcow2.xz'
        memory: 1024
        vcpus: 1
        arch: x86_64
        networks:
          - name: default
        copy_ssh_keys: true
        additional_storage: 10G
        cloud_config:
          users:
            - name: admin
              gecos: Admin User
              groups: wheel
              sudo: ALL=(ALL) NOPASSWD:ALL
              ssh-import-id: None
              lock_passwd: true
```
once linchpin provisioning is done after running 
```
linchpin -v up
```
You should be able to login to the machineip using
```
# ssh -i ~/.ssh/<resource_group_name> admin@<machineip>
# above example 
ssh -i ~/.ssh/ex admin@<machineip>
```

herlo:
Another question I have here, what does 'copy_ssh_keys' do, exactly? I don't think I want all of my keys copied to every instance spun up with this topology.

samvarankashyap:
@herlo : 
when copy_ssh_keys = true linchpin creates and injects new key.  per resource group . 
Few images have designated username and password by default they donot need ssh keys to be injected into image . In those cases copy_ssh_keys should be false.

samvarankashyap:
@herlo : I just debugged into copy_ssh_keys value. Injecting the keys should be default behaviour in our case. Its no more necessary in topology i updated the checks.

herlo:
@samvarankashyap in each case, one with and one without copy_ssh_keys, you do not pass a single key. Is copy_ssh_keys supposed to relate to ssh-import-id in the cloud_config subsection? I'm just not seeing where the keys are and which ones are being copied to the new instance. It seems all public keys from ~/.ssh/ would get copied or something in that scenario, but it isn't clear.

samvarankashyap:
@herlo : 
provision/roles/libvirt/tasks/provision_libvirt_node.yml
```
- name: "Generate ssh keys"
  shell: "ssh-keygen -t rsa -f ~/.ssh/{{ res_grp_name }} -N '';cat ~/.ssh/{{ res_grp_name }} >> ~/.ssh/authorized_keys"
  args:
    creates: "~/.ssh/{{ res_grp_name }}"
  when: uri_hostname =='localhost'

- name: "Generate ssh keys"
  shell: "ssh-keygen -t rsa -f ~/.ssh/{{ res_grp_name }} -N '';cat ~/.ssh/{{ res_grp_name }} >> ~/.ssh/authorized_keys"
  args:
    creates: "~/.ssh/{{ res_grp_name }}"
  remote_user: "{{ res_def['remote_user'] | default('root') }}"
  delegate_to: "{{ uri_hostname }}"
  when: uri_hostname != 'localhost'

- name: "cat out contents of public key"
  shell: "cat ~/.ssh/{{ res_grp_name }}.pub"
  register: pubkey_local
  when: uri_hostname == 'localhost'

- name: "cat out contents of public key"
  shell: "cat ~/.ssh/{{ res_grp_name }}.pub"
  register: pubkey_remote
  remote_user: "{{ res_def['remote_user'] | default('root') }}"
  delegate_to: "{{ uri_hostname }}"
  when: uri_hostname != 'localhost'
```
^^^ as you can see , the above code generates a ssh keypair on either localhost or remote host and and registers its public key contents on pubkey_local or pubkey_remote , these varibles are rendered in [https://github.com/samvarankashyap/linchpin/blob/libvirt_cloudconfig2/linchpin/provision/roles/libvirt/templates/cloud-config/user-data-local](https://github.com/samvarankashyap/linchpin/blob/libvirt_cloudconfig2/linchpin/provision/roles/libvirt/templates/cloud-config/user-data-local)
In above steps only one key is being used and injected into image via cloud-init scripts

herlo:
I wonder if this option is the best way. What if a user has a standard key they use and want to use it here? How do they specify it? Can you turn off the ssh key generation? It seems to me that it might be better to think that part through a little. I think it would be nice to be able to give the path of my key and have that be used rather than generating one. What do you think? What do the users think here?

samvarankashyap:
@herlo : 
users have following options in above scenario:
1. Use their custom key by giving key path. (unimplemented but can be done)
2. Ask linchpin to generate a key for them and inject it ( implemented )
3. Instead of generating the key we could always default it to use ( ~/.ssh/id_rsa.pub ) making things easier . However , What if the users don't have a key generated in home dir. ( easier to implement as we failover to generate a new key) ( can be implemented )

- Ideally making things easier , Having a unique key per resource group makes end user life easier. As a end user I need not do anything to make things work. I just need to specify the user details as part of cloud-config. 

- We could always give an option to specify their custom key . In this case end user has to specify either of the following in topology file .
    1. Key file path in topology file ( which is not suggested as hardcoded paths effects topology portability)
    2. keyfile definition as part of topology. ( not a great idea but possible)

- In an ideal scenario , We can also provide linchpin keystore( just like we did with credentials) so that we can have a place where all the ssh keys stored by default and can be used in topologies by name. 
        advantage: we can store all keys (openstack , aws, gcloud , libvirt) at one place use them across resource types by name.

herlo:
@samvarankashyap I think we're close on this with the suggestions you made. First off #1 and #3 seem less than ideal. The feature you have already implemented (#2) is good and should not be removed. This is useful as you mentioned in your argument. 

The other component you mentioned was a custom key. This key would already exist on the system. The default location could be `~/.ssh/` as expected.  We could provide an option in the topology, linchpin.conf, or on the command line to the ssh_key_path (or a similar named variable). 

How difficult to would it be to add this?

samvarankashyap:
@herlo As per discussion with the latest commit , default_ssh_key_path  is added to linchpin.conf and provisioning playbooks are updated . Further , users can specify keynames with ssh_key attribute in topology.
updated topology is as follows:
```yaml
---
topology_name: "libvirt"
resource_groups:
  -
    resource_group_name: "ex"
    res_group_type: "libvirt"
    res_defs:
      - res_name: "centos71"
        res_type: "libvirt_node"
        driver: "qemu"
        uri: 'qemu:///system'
        count: 1
        image_src: 'http://cloud.centos.org/centos/7/images/CentOS-7-x86_64-GenericCloud-1608.qcow2.xz'
        memory: 2048
        vcpus: 2
        ssh_key: "testkey"
        arch: x86_64
        networks:
          - name: default
        additional_storage: 10G
        cloud_config:
          users:
            - name: admin
              gecos: Admin User
              groups: wheel
              sudo: ALL=(ALL) NOPASSWD:ALL
              ssh-import-id: None
              lock_passwd: true
```

scripts/install_libvirt_deps.sh
```diff
@@ -1,34 +1,7 @@
...
 
-if [ $# -lt 1 ]; then
-  echo "Usage $0 <lp-path>"
-  echo
-  echo lp-path: path to linchpin source
-  exit 1
-fi
+dnf install libvirt-devel -yq
+dnf install libguestfs-tools python-libguestfs
+pip install linchpin[libvirt]
```
herlo: This will install from the linchpin package on pypi. What if a developer wants to do this? The purpose of these scripts was to give that flexibility. Having the lp-path listed solved this issue. If a user did:

```
./scripts/install_libvirt_deps.sh linchpin
```
The path would be converted (on line 26 below) to `pip install linchpin[libvirt]`. This new way subverts that process and REQUIRES a user use only the packaged version on pypi.

scripts/install_libvirt_deps.sh
```diff
@@ -1,34 +1,7 @@
 #!/bin/bash
 
-if [ $# -lt 1 ]; then
-  echo "Usage $0 <lp-path>"
-  echo
-  echo lp-path: path to linchpin source
-  exit 1
-fi
+dnf install libvirt-devel -yq
```
herlo: Removing the test for the root user is problematic. It requires that the user know that they need to run this script as root. It doesn't do the necessary to determine the user for them, which might prompt for a password via the sudo settings. The old script didn't assume the user would be root. Especially since the pip install below this would need to be for the virtualenv, whcih would likely be an unprivileged user.

scripts/install_libvirt_deps.sh
```diff
@@ -1,34 +1,7 @@
...
-
-if [ -n "${VIRTUAL_ENV}" ]; then
-
-    if [ "${USER}" != "root" ]; then
-        sudo dnf install libvirt-devel libselinux-python -yq &> /dev/null
-    else
-        dnf install libvirt-devel libselinux-python -yq &> /dev/null
-    fi
-
-    ln -s ${LIBSELINUX_PATH}/selinux ${VIRTUAL_ENV}/${VENV_LIB_PATH} &> /dev/null
```
herlo: It makes sense to have the libselinux stuff separate for other items. But since it is also a prerequisite for libvirt to work, why not have this script now call the one you moved out of this one?

scripts/install_selinux_venv.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+VENV_LIB_PATH=lib/python2.7/site-packages
+LIBSELINUX_PATH=/usr/lib64/python2.7/site-packages
+
+if [ -n "${VIRTUAL_ENV}" ]; then
```
herlo: The dnf command here assumes root user. Why not ask if the user is root, and if not, prepend sudo onto the command? I explained why this would be useful above?

herlo:
I went ahead and fixed the scripts, since I needed them to test. I updated the libvirt install script to assume PWD if no input was provided. Here you go:

scripts/install_libvirt_deps.sh 
```
#!/bin/bash

if [ "${1}99" != "99" ]; then
    LP_PATH=${1}
else
    LP_PATH=${PWD}
fi

if [ "${USER}" != "root" ]; then
    sudo dnf install libvirt-devel -yq
    sudo dnf install libguestfs-tools python-libguestfs -yq
else
    dnf install libvirt-devel -yq
    dnf install libguestfs-tools python-libguestfs -yq
fi

pip install ${LP_PATH}\[libvirt\]

if [ -n "${VIRTUAL_ENV}" ]; then
   ${PWD}/scripts/install_selinux_venv.sh
fi
```
scripts/install_selinux_venv.sh 
```
#!/bin/bash

VENV_LIB_PATH=lib/python2.7/site-packages
LIBSELINUX_PATH=/usr/lib64/python2.7/site-packages

if [ -n "${VIRTUAL_ENV}" ]; then
    if [ "${USER}" != "root" ]; then
        sudo dnf install libvirt-devel libselinux-python -yq &> /dev/null
    else
        dnf install libvirt-devel libselinux-python -yq &> /dev/null
    fi

    ln -s ${LIBSELINUX_PATH}/selinux ${VIRTUAL_ENV}/${VENV_LIB_PATH} &> /dev/null
    ln -s ${LIBSELINUX_PATH}/_selinux.so ${VIRTUAL_ENV}/${VENV_LIB_PATH} &> /dev/null
else
    echo "A virtual environment is required"
    exit 1
fi
```

linchpin/provision/roles/libvirt/tasks/provision_libvirt_node.yml
```diff
@@ -23,69 +27,276 @@
...
+  delegate_to: "{{ uri_hostname }}"
+  when: uri_hostname != 'localhost'
+
+- name: "Add additional storage"
+  command: "qemu-img resize {{ definition[1] }}/{{ definition[0] }}_{{ definition[4] }}.{{ definition[2] }} +{{ definition[3] }}G"
+  with_nested:
+    - ["{{ libvirt_resource_name }}"]
+    - ["{{ libvirt_image_path }}"]
+    - ["{{ img_src_ext }}"]
+    - ["{{ res_def['additional_storage'] | default('1') }}"]
```
herlo: This feels like good option. But it doesn't appear to be in the schema? Maybe update that so it checks for it appropriately?

linchpin/defaults/schemas/schema_v4.json
```diff
@@ -709,6 +709,12 @@
                 "boot_dev": {
                     "type":"string"
                 },
+		"additional_storage": {
```
herlo: When you update the schema, it seems that you don't convert tabs to spaces. If you are using vim, you can use `expandtab`. Here's a document on how to fix this as it causes inconsistencies in the schema documents at a minimum.

http://vim.wikia.com/wiki/Converting_tabs_to_spaces

Here's a screenshot of the problem in my vim editor.

screenshot from 2017-11-10 15-18-55

herlo:
After updating to the latest code fixes, I've been able to get further. Now, however, I run into the following error:

```
TASK [libvirt : local: Generate ssh keys when they donot exist] *****************************************
fatal: [localhost]: FAILED! => {"failed": true, "msg": "The task includes an option with an undefined variable. The error was: 'dict object' has no attribute 'ssh_key'\n\nThe error appears to have been in '/home/herlo/Projects/centos/git/linch-pin/linchpin/provision/roles/libvirt/tasks/provision_libvirt_node.yml': line 71, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: \"local: Generate ssh keys when they donot exist\"\n  ^ here\nThis one looks easy to fix.  It seems that there is a value started\nwith a quote, and the YAML parser is expecting to see the line ended\nwith the same kind of quote.  For instance:\n\n    when: \"ok\" in result.stdout\n\nCould be written as:\n\n   when: '\"ok\" in result.stdout'\n\nOr equivalently:\n\n   when: \"'ok' in result.stdout\"\n\nexception type: <class 'ansible.errors.AnsibleUndefinedVariable'>\nexception: 'dict object' has no attribute 'ssh_key'"}
```
I'm pretty sure the reason this is happening is because the res_defs['ssh_key'] is not set in my topology file. Nor does it seem that it's optional or required in the schema. I wonder what this is attempting to do without that value.

samvarankashyap:
@herlo : Latest commit should fix the issue .

herlo:
```
TASK [libvirt : local: Generate ssh keys when they donot exist] ******************************
fatal: [localhost]: FAILED! => {"failed": true, "msg": "template error while templating string: unexpected '}', expected ')'. String: {{ default_ssh_key_path | default('~/.ssh/') }}/{{ res_def['ssh_key'] | default(res_grp['resource_group_name'] }}"}
```

This is the new error I get using the same topology as before.

samvarankashyap:
@herlo : copy_ssh_id is removed as its no longer necessary.

herlo:
@samvarankashyap I'm getting success now with the linchpin up / destroy commands. I can ssh in using the standard ping hook as well. I'll next perform a code review and give feedback there. This is looking good though! :)

samvarankashyap:
@herlo Awesome :+1:

linchpin/provision/roles/libvirt/tasks/provision_libvirt_node.yml
```diff
@@ -23,69 +27,279 @@
...
+- name: "Create directories"
+  file:
+    path: "/tmp/vm-{{ definition[0] }}_{{ definition[1] }}"
+    state: "directory"
+  with_nested:
+    - ["{{ libvirt_resource_name }}"]
+    - "{{ res_count.stdout }}"
+  loop_control:
+    loop_var: definition
+  remote_user: "{{ res_def['remote_user'] | default('root') }}"
```
herlo: This seems to be a theme. I think we should take a minute and discuss what this is doing and why root is necessary. It may matter, but let's discuss and make sure it's the best way to do things. There may be better ways to grant remote (or local) privilege escalation.

linchpin/provision/roles/libvirt/tasks/provision_libvirt_node.yml
```diff
@@ -96,14 +310,31 @@
...
+      sleep 1
+    done
+    arp -an | grep -F {{ extract_mac_address_result.results[mac].stdout }} | cut -f 2 -d "(" | cut -f 1 -d ")"
+  with_items:
+    - "{{ res_count.stdout }}"
+  loop_control:
+    loop_var: mac
+  register: extract_ip_address_result_remote
+  remote_user: "{{ res_def['remote_user'] | default('root') }}"
+  delegate_to: "{{ uri_hostname }}"
```
herlo: Why is delegate_to necessary here when the uri_hostname is localhost?

linchpin/provision/roles/libvirt/tasks/provision_libvirt_node.yml
```diff
@@ -23,69 +27,279 @@
...
+    - ["{{ libvirt_resource_name }}"]
+    - "{{ res_count.stdout }}"
+  loop_control:
+    loop_var: definition
+  remote_user: "{{ res_def['remote_user'] | default('root') }}"
+  delegate_to: "{{ uri_hostname }}"
+  when: node_exists['failed'] is defined and uri_hostname != 'localhost' and res_def['cloud_config'] is defined
+
+- name: "localhost: Generate ci data cd image for cloud-init when cloud config is defined"
+  command: mkisofs -o /tmp/vm-{{ definition[0] }}_{{ definition[1] }}.iso -V cidata -r -J --quiet /tmp/vm-{{ definition[0] }}_{{ definition[1] }}/user-data /tmp/vm-{{ definition[0] }}_{{ definition[1] }}/meta-data
```
herlo: Does this mean another requirement for linchpin?

linchpin/provision/roles/libvirt/tasks/provision_libvirt_node.yml
```diff
@@ -23,69 +27,279 @@
...
+    - ["{{ libvirt_resource_name }}"]
+    - "{{ res_count.stdout }}"
+  loop_control:
+    loop_var: definition
+  remote_user: "{{ res_def['remote_user'] | default('root') }}"
+  delegate_to: "{{ uri_hostname }}"
+  when: res_def['cloud_config'] is defined and node_exists['failed'] is defined and uri_hostname != 'localhost'
+
+- name: "Install VM"
+  command: "virt-install --connect {{ definition[6] }} --import --name {{ definition[0] }}_{{ definition[7] }} --autostart --network bridge={{ definition[3] }},model=virtio --ram {{ definition[4] }} --disk path={{ definition[1] }}/{{ definition[0] }}_{{ definition[7] }}.{{ definition[2] }},format={{ definition[5] }},bus=virtio,cache=none --disk path=/tmp/vm-{{ definition[0] }}_{{ definition[7] }}.iso,device=cdrom --wait 10 --os-type=linux --nographics"
```
herlo: Is this another requirement for linchpin now?

linchpin/provision/roles/libvirt/tasks/provision_libvirt_node.yml
```diff
@@ -23,69 +27,279 @@
...
+    - ["{{ res_def['format'] | default('qcow2')  }}"]
+    - ["{{ res_def['uri'] }}"]
+    - "{{ res_count.stdout }}"
+  loop_control:
+    loop_var: definition
+  become: yes
+  ignore_errors: yes
+  when: node_exists['failed'] is defined and res_def['cloud_config'] is defined
+
+- name: "Remove cloud-init cdrom "
```
herlo: Seems that we're just shelling out to shell out here. Can you not use the virt module to accomplish this task? Additionally, does using virsh create yet another dependency?

linchpin/provision/roles/libvirt/tasks/teardown_libvirt_node.yml
```diff
@@ -27,4 +40,15 @@
     loop_var: instance
   register: res_def_output
   when: not async
+  ignore_errors: yes
+
+- name: debug
```
herlo: Comment this debug out, yeah?

linchpin/provision/roles/libvirt/templates/hosts
```diff
@@ -0,0 +1,558 @@
+
```
herlo: What is this for? Seems like a bunch of unnecessary html.

linchpin/provision/roles/libvirt/templates/libvirt_node.xml.j2
```diff
@@ -19,17 +19,17 @@
...
-{% if definition[0]['driver'] is defined and definition[0]['driver'] == 'qemu' %}
+{% if definition[3]['driver'] is defined and definition[1]['driver'] == 'qemu' %}
     <emulator>/usr/bin/qemu-system-x86_64</emulator>
 {% else %}
     <emulator>{{ qemu_path }}qemu-kvm</emulator>
 {% endif %}
     <disk type='file' device='disk'>
       <driver name='qemu' type='qcow2'/>
-      <source file='{{ definition[2] }}{{ libvirt_resource_name }}_{{ definition[1] }}.{{ definition[3] }}'/>
+      <source file='{{ definition[4] }}/{{ libvirt_resource_name }}_{{ definition[3] }}.{{ definition[5] }}'/>
```
herlo: In the virt: define action above, the code moved `libvirt_resource_name` into the loop. It seems libvirt_resource_name is appropriate here, but it should be consistent.

linchpin/provision/roles/libvirt/tasks/provision_libvirt_node.yml
```diff
@@ -23,69 +27,279 @@
   set_fact:
     img_src_ext: ''
 
-- name: "copy image_src"
-  include: copy_image_src.yml
-  when: node_exists['failed'] is defined
+- name: "copy image_src: localhost"
+  include: copy_image_src_local.yml remote_user={{ res_def['remote_user'] | default('root') }}
```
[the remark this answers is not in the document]
samvarankashyap: when the remote_uri is something like uri://192.168.122.1/system . The following are the steps taken by the linchpin :  
1. gets the host name 
2. uses remote_user to copy the downloaded qcow2/xz image on host( machine running linchpin) to remote machine. ( something similar to scp but done through ansible) 

As you mention assuming remote user to be 'root' is bad thing , I would default it to the current user running linchpin. how does that sound.

linchpin/provision/roles/libvirt/tasks/provision_libvirt_node.yml
samvarankashyap: 'root' is not necessary , its defaulted . I will change it current;y active user running linchpin . in case of remote_machine. As its tmp folder , it doesnt need privilege acceleration in this case

linchpin/provision/roles/libvirt/tasks/provision_libvirt_node.yml
samvarankashyap: @herlo : mkisofs is by default available in fedora , ( not sure about centos but it should be ) . That is why i didnt add it as a dependency.

linchpin/provision/roles/libvirt/tasks/provision_libvirt_node.yml
samvarankashyap: @herlo : As mentioned , virt-install is a part of libguestfs . Which is a dependency needed to achieve cloud-init config

linchpin/provision/roles/libvirt/tasks/provision_libvirt_node.yml
samvarankashyap: @herlo : virt module doesnot yet support change-media function . 
We could raise a PR in ansible to make it do so .

linchpin/provision/roles/libvirt/tasks/provision_libvirt_node.yml
samvarankashyap: Noted 
delegate_to is required when uri_hostname is not equal to localhost because , 
the ip addresses are local to machine libvirt is running . We need to get inside the machine to extract machine ip from arp command .

linchpin/provision/roles/libvirt/tasks/provision_libvirt_node.yml
herlo: Probably a good idea. Or maybe suggest a cloud-init module? Not sure here. Let me know what you decide.

linchpin/provision/roles/libvirt/tasks/copy_image_src_local.yml
```diff
@@ -0,0 +1,105 @@
...
+- name: "set local_image_path"
+  set_fact:
+    local_image_path: "{{ libvirt_image_path }}"
+
+- name: "Inside copy image src "
+  debug:
+    msg: "{{ local_image_path  }}  {{ libvirt_image_path }}"
+
+
+- name: "delegate_to localhost :ensure local_image_path directory exists"
```
herlo: This should have a `become: yes` on it for non-root users.

samvarankashyap: @herlo fixed

PR #385, opened 2017-11-07
Added missing package requirements

beerparty:
When I was trying to install LinchPin in a CentOS container (which already had EPEL), the LinchPin install failed because required packages were not available. I have added those requirements to the requirements.txt file of LinchPin.


herlo:
Hi @beerparty, thanks for the update to requirements.txt. Unfortunately, the travis-ci tooling noted that you removed tinydb which is required now. It was just recently added as a dependency, so that may be the reason it's not in requirements.txt

Would you mind updating the file to have this dependency? 

Thanks!

beerparty:
Sorry for the omission. __TinyDB__ has been added to the requirements.txt file.

herlo:
What version of CentOS are you using? 

I'm currently testing on 

```
$ cat /etc/redhat-release 
CentOS Linux release 7.4.1708 (Core)
```
Using the following requirements.txt in a clean virtualenv:

```
setuptools>=35.0
pytz
pyopenssl
boto>=2.40.0
apache-libcloud>=0.20.1
click
camel
yamlordereddictloader
ansible>=2.3.1
shade>=1.22.2
naked
cerberus
tinydb
```

And it's working for me. setuptools needed to be updated. The additional dependencies added are:

```
Babel!=2.4.0,>=2.3.4
cryptography>=1.9
requests>=2.14.2
```
These get satisfied by other dependencies and get installed as needed. One thing that is not clear with pip is the fact that it's pip and setuptools that usually need an update prior to running the pip install against the local directory. The package installation documentation from python.org shows in the first paragraph that the standard action is to update pip and setuptools. Please see https://packaging.python.org/tutorials/installing-packages/ for reference. 

Essentially, this seems to be a documentation issue, especially in the installation section of the LinchPin docs. These docs are going to be getting a refresh in the next couple of weeks. But if you wish to update them before this point, please feel free to do so, it would be much appreciated.

beerparty:
I am installing on CentOS7. **Note** that I am creating an OpenShift container image (via a Dockerfile), not installing on an existing system. Installing on my local fedora system (with or without virtualenv) worked fine without these changes.

The problem may be with EPEL. I was told to install EPEL first, and the packages I have mentioned have earlier versions (I think) because of the EPEL installation (epel-release-latest-7). Without updating the packages mentioned, the installation fails due to incompatible package versions.

PR #390, opened 2017-11-14
Docker images for testing.

p3ck:
First attempt at docker images to verify linchpin works on
centos and fedora environments.  Currently it only tests libvirt
and dummy.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,5 @@ nosetests.xml
 .idea
 .pydevproject
 
+Dockerfiles/pip_install.sh
+Dockerfiles/lp_test_workspace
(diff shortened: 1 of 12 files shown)


herlo:
@p3ck I'm curious how you ensure that libvirt is running in the docker instance with these distros. I saw the script for cento6, but what about the rest?

p3ck:
@herlo it's done via systemd services.  notice the command is ['/usr/sbin/init']

herlo:
@p3ck I saw that in place.  The containers have different dependencies which I was trying to sort out. I was trying to get this to work with nested virt in VMs to verify some of the issues, but things were not easy to configure and those scripts fail on the VMs I tried (centos 6 and fedora 25, specifically). I'll do some more testing tomorrow and update this PR at that time.

PR #392, opened 2017-11-14
Update installation doc and add OS sub-sections

jpryor-rh:
Update generic installation notes
Added sub-section for installing on Fedora 26
Added sub-section for installing on RHEL 7.4

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -89,7 +89,7 @@ Install RPM pre-reqs
 
 .. code-block:: bash
 
-    $ dnf -y install python-virtualenv libffi-devel openssl-devel libyaml-devel gmp-devel libselinux-python make gcc redhat-rpm-config libxml2-python
+    $ sudo dnf -y install python-virtualenv libffi-devel openssl-devel libyaml-devel gmp-devel libselinux-python make gcc redhat-rpm-config libxml2-python
 
 
 Create a working-directory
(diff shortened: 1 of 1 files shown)


jpryor-rh:
This is an attempt at providing potential users of linchpin well formed examples of how to install it as a CLI on Fedora and RHEL7.
And @BillRainford wanted to be at-ed.

docs/source/installation.rst
```diff
@@ -76,6 +81,120 @@ If testing or docs is desired, additional steps are required.
...
+    (linchpin) $ pip install linchpin
+
+Make a workspace, and initialize it to prove that linchpin itself works
+
+.. code-block:: bash
+
+    (linchpin) $ mkdir workspace
+    (linchpin) $ export WORKSPACE=./workspace
+    (linchpin) $ linchpin init
+    PinFile and file structure created at /root/work/workspace
```
herlo: Could this not assume root? Maybe make it say something like `/home/user/workspace`?

docs/source/installation.rst
```diff
@@ -76,6 +81,120 @@ If testing or docs is desired, additional steps are required.
...
```
herlo: Same as comment above. Additionally, the WORKSPACE variable isn't specifically required if the workspace is $PWD. It might be good to make a not of this somewhere.

docs/source/installation.rst
```diff
@@ -76,6 +81,120 @@ If testing or docs is desired, additional steps are required.
...
 
+
+Installing LinchPin on Fedora 26
+---------------------------------
+
+Install RPM pre-reqs
+
+.. code-block:: bash
+
+    $ dnf -y install python-virtualenv libffi-devel openssl-devel libyaml-devel gmp-devel libselinux-python make gcc redhat-rpm-config libxml2-python
```
herlo: As non-root, encourage use of sudo here.

Issue #389, opened 2017-11-14
don't assume $HOME/.config/ exists


p3ck:
a fresh install and new user will not have a .config directory if they never ran any other software that would create it.  linchpin should not assume it exists and create the directory if needed.

robled:
For CI workflows, this file (and all others needed to execute linchpin successfully) should also be optionally stored/read from the linchpin workspace, as writing to `~/.config` is not compatible with CI jobs that should be self-contained.

herlo:
@p3ck I've fixed this in my latest work. I have yet to merge it. I'll try to update this ticket when I push a new PR. Thanks

herlo:
@robled in terms of why this exists, the runDB component cannot live in the workspace. If it does, provisioning multiple systems of the same type will not be centrally tracked. The runDB feature provides a central location for storing the database. The $HOME/.config/linchpin path is where this is stored by default, thus it must be created to accommodate this location. 

If you desire to switch it, however, it can be updated in the linchpin.conf, which can be stored in the workspace.

herlo:
@robled you'll find this setting on line 10 of the current linchpin.conf file in the develop branch. The value is: `default_config_path = ~/.config/linchpin`. If you want to specifically change the rundb connection string, it's on line 21, which reads `rundb_conn = %(default_config_path)s/rundb-::mac::.json`. These values can be changed to whatever you'd like. Keep in mind, the `::mac::` component is used to make the database unique on a per machine basis. Feel free to tweak it and see what works for you.

herlo:
Fixed by #398

jaypoulz:
@herlo Running the the release1.5 candidate branch, I get the following:
```
[x86_64] + linchpin --workspace workspace --pinfile workspace/PinFile --template-data '{ arch: x86_64 }' --verbose up jenkins-slave
[x86_64] Using pinfile named 'workspace/PinFile'
[x86_64] ctx.workspace: /tmp/workspace/tiarch-ci-test-template_dev-JU5BEVQJQDO2U7Y2VYVV6NLD2RLNE6NBOV6Y7QE2OCYCR46V26PQ/workspace
[x86_64] Traceback (most recent call last):
[x86_64]   File "/usr/bin/linchpin", line 11, in <module>
[x86_64]     load_entry_point('linchpin==1.5.0rc1', 'console_scripts', 'linchpin')()
[...]

Issue #388, opened 2017-11-14
failure to initialize on centos6


p3ck:
using pinfile and topology from https://github.com/herlo/lp_test_workspace

```
linchpin up libvirt
/usr/lib64/python2.6/site-packages/cryptography/__init__.py:26: DeprecationWarning: Python 2.6 is no longer supported by the Python core team, please upgrade your Python. The next version of cryptography will drop support for Python 2.6
  DeprecationWarning
/usr/lib/python2.6/site-packages/linchpin/api/context.py:89: DeprecationWarning: The SafeConfigParser class has been renamed to ConfigParser in Python 3.2. This alias will be removed in future versions. Use ConfigParser directly instead.
  config = ConfigParser.SafeConfigParser()
/usr/lib/python2.6/site-packages/linchpin/api/context.py:91: DeprecationWarning: This method will be removed in future versions.  Use 'parser.read_file()' instead.
  config.readfp(f)
Traceback (most recent call last):
  File "/usr/bin/linchpin", line 11, in <module>
    load_entry_point('linchpin==1.0.4', 'console_scripts', 'linchpin')()
  File "/usr/lib/python2.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python2.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python2.6/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python2.6/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python2.6/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python2.6/site-packages/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args[1:], **kwargs)
  File "/usr/lib/python2.6/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python2.6/site-packages/linchpin/__init__.py", line 176, in up
    return_code, results = lpcli.lp_up(pf_w_path, targets, run_id=run_id)
  File "/usr/lib/python2.6/site-packages/linchpin/api/__init__.py", line 245, in lp_up
    return self.run_playbook(pinfile, targets, action="up", run_id=run_id)
[...]

PR #398, opened 2017-11-16
Misc fixes

### LINKED ISSUE #397: Duffy allocation fails
using latest version from git I get the following failure when trying to bring up a duffy node.

 linchpin up duffy
[DEPRECATION WARNING]: The use of 'include' for tasks has been deprecated. Use 
'import_tasks' for static inclusions or 'include_tasks' for dynamic inclusions. This 
feature will be removed in a future release. Deprecation warnings can be disabled by 
setting deprecation_warnings=False in ansible.cfg.
[DEPRECATION WARNING]: include is kept for backwards compatibility but usage is 
discouraged. The module documentation details page may explain more about this 
rationale.. This feature will be removed in a future release. Deprecation warnings 
can be disabled by setting deprecation_warnings=False in ansible.cfg.
[DEPRECATION WARNING]: Included file '/usr/lib/python2.7/site-
packages/linchpin/provision/update_aws_cfn_outputs.yml' not found, however since this
 include is not explicitly marked as 'static: yes', we will try and include it 
dynamically later. In the future, this will be an error unless 'static: no' is used 
on the include task. If you do not want missing includes to be considered dynamic, 
use 'static: yes' on the include or set the global ansible.cfg options to make all 
includes static for tasks and/or handlers. This feature will be removed in version 
2.7. Deprecation warnings can be disabled by setting deprecation_warnings=False in 
ansible.cfg.
	to retry, use: --limit @/usr/lib/python2.7/site-packages/linchpin/provision/site.retry
Target 'duffy': TASK: duffy : Set cred path failed with error 'The conditional check 'creds_path == default_credentials_path' failed. The error was: error while evaluating conditional (creds_path == default_credentials_path): 'creds_path' is undefined

The error appears to have been in '/usr/lib/python2.7/site-packages/linchpin/provision/roles/duffy/tasks/main.yml': line 13, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

# duffy puts its key by default in the user's home directory.
- name: "Set cred path"
  ^ here
'

using the following topology from lp_test_workspace:
---
topology_name: duffy1
resource_groups:
  - resource_group_name: duffy1
    resource_group_type: duffy
    resource_definitions:
      - type: duffy_node
        name: duffy
        version: 7
        arch: x86_64
        count: 1
    credentials: duffy.key

herlo:
Fixes #397 #389 #388 

Changes the default count for linchpin journal to 3, as that makes logical sense.

### DIFF
--- a/linchpin/__init__.py
+++ b/linchpin/__init__.py
@@ -78,7 +78,6 @@ def _handle_results(ctx, results, return_code):
             if trs is not None:
                 trs.reverse()
                 tr = trs[0]
-                ctx.log_state('result: {}'.format(tr._result.get('rc')))
                 if tr.is_failed():
                     msg = tr._check_key('msg')
                     ctx.log_state("Target '{0}': {1} failed with"
--- a/linchpin/__init__.py
+++ b/linchpin/__init__.py
@@ -72,7 +72,20 @@ def _handle_results(ctx, results, return_code):
         rundb_data = data['rundb_data']
         task_results = data['task_results']
 
-        return_code = 99
+        if not isinstance(task_results, int):
+            trs = task_results
+
+            if trs is not None:
+                trs.reverse()
+                tr = trs[0]
+                ctx.log_state('result: {}'.format(tr._result.get('rc')))
+                if tr.is_failed():
+                    msg = tr._check_key('msg')
+                    ctx.log_state("Target '{0}': {1} failed with"
+                                  " error '{2}'".format(target, tr._task, msg))
+        else:
+            if task_results:
+                return_code = task_results
 
         # PRINT OUTPUT RESULTS HERE
         for rundb_id, data in rundb_data.iteritems():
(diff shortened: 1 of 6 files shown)


herlo:
@p3ck I'm not experiencing the issue you are reporting. 

Running CentOS 7 and a python virtual environment, this succeeds for me with the topology from https://github.com/herlo/lp_test_workspace/blob/master/topologies/duffy.yml

samvarankashyap:
@herlo : 
I could not find duffy.py ansible module in linchpin/provision/library. 
Is it accidentally deleted in one of the PRs ? or am i missing something ?
It might be the reason for @p3ck  error! .

herlo:
Ah, good point. @p3ck might be interested that a symlink from the library
folder in https://github.com/CentOS-PaaS-SIG/duffy-ansible-module to the
linchpin/provision/library path in will do the trick. I also need to make
sure this is in the docs.

Thanks!

p3ck:
@herlo is there not an install procedure for installing this module?

herlo:
@p3ck No, there is not at this time. It was always thought that linchpin would eventually be broken into core and contributed modules. Time and other factors have prevented this from happening. 

The procedure is pretty straightforward though, and in general modules land in a library directory. According to ansible, modules live in a library path, which can be confiigured -- http://docs.ansible.com/ansible/latest/intro_configuration.html#library

If it's any consolation, Makefiles were going to be the direction for managing modules and the like. If updating that is of interest, please let me know

herlo:
I should also mention, that the duffy module is something that most users will never encounter. Duffy lives in that small space of CentOS CI that most will likely never touch. Not that it couldn't change and become the best tool ever, but I don't foresee that as a thing. This, is the exact reason the duffy module was not included in linchpin.

PR #409, opened 2017-11-21
Added bridge parameter to libvirt_network provisioning

samvarankashyap:
$Subject

Added bridge parameter to network and defaulted it to virbr0

### DIFF
--- a/linchpin/provision/roles/libvirt/templates/libvirt_network.xml.j2
+++ b/linchpin/provision/roles/libvirt/templates/libvirt_network.xml.j2
@@ -1,5 +1,8 @@
 <network ipv6='yes'>
   <name>{{ res_def['res_name'] | default(res_def['name'])  }}</name>
+  {% if res_def['bridge'] is defined %}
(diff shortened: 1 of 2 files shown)


linchpin/provision/roles/libvirt/templates/libvirt_network.xml.j2
```diff
@@ -1,5 +1,6 @@
 <network ipv6='yes'>
   <name>{{ res_def['res_name'] | default(res_def['name'])  }}</name>
+  <bridge name={{ res_def['bridge'] | default('virbr0') }} stp='on' delay='0'/>
```
herlo: Instead of making it default to something, wouldn't it be better to do a `{% if res_def['bridge'] %}` section? Essentially, the bridge gets created unless you specify the name. So the bridge could be `virbr0` or it could be virbr9 (I have one of those from my testing back when I wrote this code). Thoughts?

samvarankashyap:
@herlo Updated!

Issue #339, opened 2017-08-08
Feature request: linchpin hooks : failhooks , nohooks, runhooks  flags while linchpin up


samvarankashyap:
Currently linchpin runs hooks by default inside the Pinfile
For example if the user would like to do following things there isnt any option:

- Fail all hooks when one hook fails ( --failhooks)
- Donot run hooks over provisioned instances ( --nohooks)
- Runhooks over provisioned instances . Do not reprovision instances ( --runhooks)
- Run a specific hook by their names ( --runhooks name1, name2 )

The above can be achieved by passing flags to linchpin up and linchpin destroy commands 

@herlo Anythoughts /additions to features are appreciated

herlo:
Reference #241 for another conversation around this...

PR #418, opened 2017-12-01
Dynamic Input Prep - LinchPin API rework

### LINKED ISSUE #316: Can the  output in resources be simplified
Can the output for the provisioners created in the resources directory be simplified, maybe only have one level and some consistent keys across different the different provisioners to simplify retrieving output from automation tools.  They seem to have complex output with multiple levels and specific output for each provisioner.  This makes it difficult for automation to switch between different provisioner options w/o having to write a specific parser for each provisioner.

### LINKED ISSUE #352: Create cerberus schema validation for openstack module
Per #235, each core provider needs to validate against its own schema. One for the openstack provider, which includes all sub-components must be created. This can be done similarly to the example provided by the dummy module, which was completed as a proof of concept. See the following files for examples:

* https://github.com/CentOS-PaaS-SIG/linchpin/blob/release1.5/linchpin/provision/dummy.yml
* https://github.com/CentOS-PaaS-SIG/linchpin/tree/release1.5/linchpin/provision/roles/dummy/

Specifically pointing out the files/schema.json in the role as a requirement.

The goal is to simplify each provider into its own ansible playbook. Then pass the arguments from the library and run each as necessary. This will improve performance, and simplify contributions. The user interface should also improve from this work.

### LINKED ISSUE #353: Create cerberus schema validation for gce (gcloud) module
Per #235, each core provider needs to validate against its own schema. One for the google compute engine (gce, gcloud) provider, which includes all sub-components must be created. This can be done similarly to the example provided by the dummy module, which was completed as a proof of concept. See the following files for examples:

* https://github.com/CentOS-PaaS-SIG/linchpin/blob/release1.5/linchpin/provision/dummy.yml
* https://github.com/CentOS-PaaS-SIG/linchpin/tree/release1.5/linchpin/provision/roles/dummy/

Specifically pointing out the files/schema.json in the role as a requirement.

The goal is to simplify each provider into its own ansible playbook. Then pass the arguments from the library and run each as necessary. This will improve performance, and simplify contributions. The user interface should also improve from this work.

### LINKED ISSUE #354: Create cerberus schema validation for libvirt module
Per #235, each core provider needs to validate against its own schema. One for the libvirt provider, which includes all sub-components must be created. This can be done similarly to the example provided by the dummy module, which was completed as a proof of concept. See the following files for examples:

https://github.com/CentOS-PaaS-SIG/linchpin/blob/release1.5/linchpin/provision/dummy.yml
https://github.com/CentOS-PaaS-SIG/linchpin/tree/release1.5/linchpin/provision/roles/dummy/

Specifically pointing out the files/schema.json in the role as a requirement.

The goal is to simplify each provider into its own ansible playbook. Then pass the arguments from the library and run each as necessary. This will improve performance, and simplify contributions. The user interface should also improve from this work.

### LINKED ISSUE #355: Create cerberus schema validation for amazon web services (aws/ec2) module
Per #235, each core provider needs to validate against its own schema. One for the amazon web services (aws/ec2) provider, which includes all sub-components must be created. This can be done similarly to the example provided by the dummy module, which was completed as a proof of concept. See the following files for examples:

https://github.com/CentOS-PaaS-SIG/linchpin/blob/release1.5/linchpin/provision/dummy.yml
https://github.com/CentOS-PaaS-SIG/linchpin/tree/release1.5/linchpin/provision/roles/dummy/
Specifically pointing out the files/schema.json in the role as a requirement.

The goal is to simplify each provider into its own ansible playbook. Then pass the arguments from the library and run each as necessary. This will improve performance, and simplify contributions. The user interface should also improve from this work.

### LINKED ISSUE #415: Extending linchpin.conf doesn't seem to work properly.
While working on some new code, I tried to override the [lp] section in the linchpin.conf and was unable to override the external_providers_path. Need to dig into this and fix the bug.

herlo:
This PR provides a much needed rework of the LinchPin API. Removing any file-based procedures from the LinchpinAPI class, and pushing them down the API tree to `linchpin.cli` and the `linchpin.shell` modules where they belong. The new structure looks like so::

* `linchpin`: Previously `linchpin.api` package. Contains `LinchpinAPI (__init__.py)`, the base of most everything.
  * Additional classes include `context.py`, `version.py`, `callbacks.py` and `ansible_runner.py`. These
    classes all correspond to core API functionality.
* `linchpin.cli`: Previously `linchpin.cli` package. Mostly unchanged here, but will undergo many changes to accommodate the dynamic inputs feature.
* `linchpin.shell`: Previously `linchpin` package. This is where the command shell now lives. A few changes have already been made here, but more are likely to come to lend flexibility to the command line interface.
    
The directory structure has changed only a little. Some folders were removed for clarity.

```    
linchpin
├── cli
├── exceptions
├── fetch
├── hooks
├── linchpin.conf
├── provision
├── rundb
├── shell
└── tests
```

A few other items that came along with this update:

* The default schemas are no longer necessary. They will be removed in a future PR.
  * Each provider has its own playbook and schema instead
* Each topology must comply to a strict format until it reaches the `resource_definitions` list of dictionaries
  * More detail on this will come in documentation
  * This is required due to the constraint of having dynamic inputs, which will need to have a standard to follow
* The `external_providers_path` variable was added to linchpin.conf
  * This can be updated like the $PATH variable to point to locations with additional provider playbooks and schemas

External playbooks can now be included. This feature also provides flexibility for developers as they can now write a schema and a simple playbook to have linchpin execute. A few strict requirements and an update the linchpin.conf with the path to the playbook will allow more providers to be added simply. Documentation on this feature will come in the next few updates.

A couple notes::

* Hooks currently do not work in this PR. This will be addressed in a future update.
* Testing is currently being updated for this PR, see the comments below for more information.

Fixes #316 
Fixes #352 
Fixes #353 
Fixes #354 
Fixes #355 
Fixes #415


herlo:
The tests need to be updated before this will pass Travis.

p3ck:
Couple of things I think should happen for this PR

- I don't think beaker, openshift, duffy and rackspace should be removed from the main repo.
- The examples under docs don't seem to match the changes you made here.  I still see res_defs in some topologies instead of resource_definitions and I also see type: used in libvirt_storage.  Won't this be a problem as well?
- The example topologies from lp_test_workspace should just live in this repo as well.

Thanks

herlo:
@p3ck In all cases you mention this will be done. 

* I'm going to move all of the providers from linchpin-providers-external into this PR shortly. 
* The documentation is the next phase and will match up before the 1.5 release hits. It will not be in this PR.
* I'm planning also to remove the example topologies here and update them (in documentation) and ONLY have those as examples anymore. Currently, they are all over the place and someone could find an example that doesn't legitimately work.

Does this seem acceptable?

p3ck:
When you say "ONLY have those", what is being referred to?  the examples from lp_test_workspace?
If yes, then I think that will work. Having some simple examples is very important IMHO.

herlo:
@p3ck Right now, there are several different places where examples of topologies, layouts, etc. exist. In the update, I'm planning to remove the current examples everywhere I find them, and include them ONLY in the documentation. The examples will be the ones from the lp_test_workspace repo.

linchpin/provision/roles/gcloud/tasks/provision_res_defs.yml
```diff
@@ -1,7 +1,7 @@
 ---
 - name: "Set gce_res_type value"
   set_fact:
-    gce_res_type: "{{ res_def['type'] | default(res_def['res_type']) }}"
```
p3ck: now I see the different providers are completely inconsistent.  aws looks for res_type and will fall back to type and this one looks for type and falls back to res_type.

I can see why you want to move to role.   

How did this mess happen to begin with?

herlo: I explained this a bit above. I hope it's enough.

linchpin/__init__.py
```diff
@@ -1,386 +1,497 @@
...
+            self.ctx.log_debug('calling: {0}{1}'.format('pre', action))
 
-    no_records = []
-    output = 'run_id\t'
+#        if 'pre' in self.pb_hooks:
+#            self.hook_state = '{0}{1}'.format('pre', action)
 
-    for f in fields:
-        output += '{0:>10}\t'.format(f)
+        # FIXME need to add rundb data for hooks results
```
samvarankashyap: I guess , with rundb library in place its easy to include the results into database . A later point of time ,  I will try to work on it.

linchpin/__init__.py
```diff
@@ -1,386 +1,497 @@
...
+                                       console=ansible_console)
+            )
 
+            if not return_code:
+                self.ctx.log_state("Action '{0}' on Target '{1}' is "
+                                   "complete".format(action, target))
 
-    try:
-        journal = lpcli.lp_journal(targets=targets, fields=fields, count=count)
+        # FIXME Check the result[target] value here, and fail if applicable.
```
samvarankashyap: I think this is related to RFE #339

linchpin/__init__.py
```diff
@@ -1,386 +1,497 @@
...
+                    uhash = data.get('uhash')
+                    self.ctx.log_debug("using data from"
+                                       " run_id: {0}".format(run_id))
+                else:
+                    raise LinchpinError("Attempting to perform '{0}' action on"
+                                        " target: '{1}' failed. No records"
+                                        " available.".format(action, target))
+            else:
+                raise LinchpinError("run_id '{0}' does not match any existing"
+                                    " records".format(run_id))
```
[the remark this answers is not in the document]
herlo: run_id is passed into the method. See https://github.com/CentOS-PaaS-SIG/linchpin/pull/418/files/3fbfdbf8d0a5985467719e061ef9edf100daee5a#diff-eda7cdccb38e3e91756c08ffdbc7ac4eR272

linchpin/__init__.py
herlo: This will get addressed in a later release.

PR #438, opened 2017-12-22
Minor Beaker fixes

herlo:
Since the Beaker ansible job allows the job_group to be blank, adjusting the schema to match.

Fixed an issue reported by @scoheb around beaker inventory generation. The inventory_gen role had an outdated when statement. This is now updated and should address the issue.


scoheb:
This PR fixes the issue with beaker.
See https://paste.fedoraproject.org/paste/R3OMtjm-B~elE662E7wxEw for an example of the generated inventory.

PR #453, opened 2018-01-16
fix example links for documentation

herlo:
Change docs links to point to develop instead of herlo repository. Other minor changes.


herlo:
If you look at the tests for this PR, they pass. However, due to what I believe is [this issue](https://twitter.com/travisci/status/953329362283180032), the travis job has not finished. I expect it will finish when they have the issue resolved.

PR #455, opened 2018-01-18
Changes to support testing PR's in Openshift

p3ck:


p3ck:
After review and any additional changes I can compress this into one commit before merging.

PR #460, opened 2018-01-25
Linchpin Constants

### LINKED ISSUE #427: linchpin_constants
Make a simple method called linchpin_constants in the API and load the constant values here. Emulate the the Ansible constants.py concept.

herlo:
Fixes #427 - Linchpin's configuration file (linchpin.conf), has been converted into linchpin.constants. The linchpin.conf continues to exist, and now performs the duty of applying configuration changes as needed in the standard locations. An example linchpin.conf has been moved to the config directory as well.

Code fixes are in place to handle loading the constants. Specifically the LinchpinContext loads the constants, the LinchpinCliContext takes care of loading any linchpin.conf files as before.

Documentation and tests have been updated to correspond to the latest changes.


docs/source/glossary.rst
```diff
@@ -54,8 +54,8 @@ The following is a list of terms used throughout the LinchPin documentation.
 
     evars
     extra_vars
-        Variables that can be passed into Ansible playbooks from external sources.
-        Used in linchpin via the linchpin.conf `[evars]` section.
+        Variables that can be passed into Ansible playbooks from external
+        sources. Used in linchpin via the linchpin.conf `[evars]` section.
 
     hook
         Certan scripts can be called when a particular :term:`hook` has been
```
samvarankashyap: I guess its,  typo "Certain"

linchpin/__init__.py
```diff
@@ -109,7 +108,6 @@ def setup_rundb(self):
                     else:
                         raise
```
samvarankashyap: raise ?? It would be great if  we might have to raise a custom exception instead empty raise.

linchpin/tests/mockdata/contextdata.py
```diff
@@ -161,9 +137,12 @@ def parse_config(self, config_data=None):
...
-                self.parser.set(k, kv, vv)
+        try:
+            for k, v in config_data.iteritems():
+                self.parser.add_section(k)
+                for kv, vv in v.iteritems():
+                    if type(vv) is not str:
+                        vv = str(vv)
+                    self.parser.set(k, kv, vv)
+        except ValueError:
+            pass
```
samvarankashyap: It would be better if we could log this error into a log file @herlo

linchpin/__init__.py
herlo: This line hasn't been changed in the latest commit. Is there a new reason to point it out now? Last time it was changed was with commit 356e6af0 (2017-12-14).

If you are curious, however, in this case, if the error is unknown and we'd want to know the actual exception raised since it would be fatal. I could raise a LinchpinError, except then it would hide the actual exception.

PR #462, opened 2018-01-30
Make generating resources optional (on by default for 1.x compatibility)

### LINKED ISSUE #459: "resources/file.output" no longer available in linchpin workspace
While testing the recent work done for #430, I was able to successfully provision an instance in OpenStack using linchpin in Jenkins as of [this commit](https://github.com/CentOS-PaaS-SIG/linchpin/commit/b11d18b412b1a94c43b915765e7df1f6d2de0106) with our existing linchpin workspaces.  However, the Jenkins teardown step fails, because the resources directory and corresponding output file that we use for the destroy operation is no longer created:

```
[jenkins@jmaster1 openstack-slave]$ find .
.
./PinFile
./credentials
./credentials/cinch.yml
./hooks
./hooks/.gitkeep
./inventories
./inventories/.gitkeep
./inventories/group_vars
./inventories/group_vars/all
./inventories/cinch-test.inventory
./keystore
./keystore/ci-ops-central
./layouts
./layouts/cinch.yml
./topologies
./topologies/cinch.yml
./.rundb
./.rundb/rundb-274973442355001.json
```

Error from Jenkins job workflow:

`15:32:13 ERROR: Failed to copy artifacts from cinch-example-provision with filter: cinch-example/examples/linchpin-topologies/openstack-slave/resources/cinch-test.output`

JJB template that has been using this workflow with linchpin 1.0.4 and prior:

https://github.com/RedHatQE/cinch/blob/master/jjb/ci-jslave-project-sample.yaml#L28

herlo:
Fixes #459 

Providing a flag that generates the resources files by default for compatibility to 1.0.x releases.


config/linchpin.conf
```diff
@@ -97,8 +97,8 @@ enable_uhash = True
 # in older versions of linchpin (<v1.0.4), a resources folder exists, which
 # dumped the data that is now stored in the RunDB. To disable the resources
 # output, set the value to False.
-#generate_resource = True
-generate_resource = False
+#generate_resources = True
+generate_resources = False
```
robled: If the backwards-compatible resources output is off by default, then a lot of existing linchpin users would need to update a large number of CI jobs to turn this back on.  Can we set this to **True** and generate the resources files by default so existing CI jobs that invoke linchpin can continue to work without modification?

linchpin/provision/roles/gather_resources/tasks/main.yml
```diff
@@ -41,3 +41,21 @@
...
+  when: generate_resources
+  ignore_errors: yes
+
+- name: "set topology_name"
+  set_fact:
+    topology_name: "{{ topo_data['topology_name'] }}"
+
+- name: "Generate resources file (v1.0.x compatibility)"
+  template:
+    src: "../templates/output_formatter.j2"
```
p3ck: This path seems to be incorrect.  That is what is causing travis to fail.

config/linchpin.conf
herlo: Hi @robled. What you may have missed was a slight adjustment made here. PR #460 explains the adjustment.  Essentially, there is now a set of defaults, managed by linchpin.constants. In that [file](https://github.com/CentOS-PaaS-SIG/linchpin/blob/develop/linchpin/linchpin.constants#L55), you will see that generate_resources is set to **False** as desired. The config/linchpin.conf no longer controls the defaults in linchpin. The linchpin.conf you referenced here is provided as an example and doesn't change the settings. 

The reason for this change is so that backward compatibility can be set as the default without having to do anything. Overriding settings, as you have been doing with the linchpin.conf in the workspace should continue to work as normal. Please test this and verify that indeed the resources are being created as you expect. If you find any errors, please let me know.

linchpin/provision/roles/gather_resources/tasks/main.yml
herlo: I missed adding it, just updated it. Thanks @p3ck! :)

PR #468, opened 2018-02-08
Allow passing provisioning options through to bkr_info

### LINKED ISSUE #394: Error / Timeout in beaker provision should cancel the job
### Problem Description
The way beaker provisioning currently works presents a problem when the provisioning fails because of timeout when waiting for the resource. The problem is that the submitted beaker job is not canceled, and there is not sufficient information passed to the caller to cancel it manually. 

### Why this is a Problem
- In a test environment where linchpin is called to provision a resource to be used for the test, this results in wasted resources, since restarting the job queues a second job requesting the same resources, making it more likely to fail on a timeout since the first job is already queued.
- Requests that fail in a timeout fashion don't ever supply the job id, since that information is supplied in the *.output file this is not generated. Solving #180 may alleviate this problem in the future; however, even if #180 was fixed, we'd desire a way to ensure that the *beaker job* could be canceled from the command line, hopefully via a linchpin command. 
- Currently, the only command capable of canceling the beaker job is *linchpin destroy*, which expects an inventory file. This file cannot be generated until a machine is reserved by beaker, which itself cannot be guaranteed to occur before the timeout is hit. This means you don't have sufficient information to manually cancel the job on a timeout failure.

### Proposed Solution
Linchpin should cancel the job if it times out. This prevents wasted resources in beaker, and prevents the user from manually having to cancel jobs in beaker via alternative means.

The alternatives I can think of are:
1.  Eliminate the ansible watchdog timeout, so that linchpin only returns when the job is aborted, cancelled, or running.
2. Provide the user the job id of the beaker request even when the provision times out, and a way to monitor the job or cancel it manually via the linchpin cli.

### LINKED ISSUE #399: RFE: Print URL of Beaker job so user gets more info about the job
As per what @robled said in #180:
_Another nice thing to do would be to print the URL to the Beaker job so that it's easy for a user to get more info about the job without having to manually look up the job ID in the Beaker web interface._

Indeed, it would be an nice feature to have since right now I have to manually toggle over to the beaker webui to figure out what job & machine it landed on.

### LINKED ISSUE #424: Beaker timeout should be configurable via the topology
# Overview
Currently, the bkr_info.py file specifies that it will check for provisioning updates once every minute for 60 minutes, and if the server isn't provisioned in that amount of time, than the provision has failed.

The amount of time you wait for the timeout should be a parameter you can pass along in the topology file.

seandst:
Previously, the bkr_server would receive all provisioning options and
then pass only job IDs through to bkr_server for monitoring, which made
it impossible to pass those options through. It's useful to be able to
pass those through, as that allows for changing things like how long
bkr_info will wait for a provisioning request to succeed or fail.

This also includes a variety of quality of life fixes:

---
closes #424
CentOS-PaaS-SIG#424

Being able to specify `max_attempts` and `attempt_wait_time` in the
topology allows you to effectively specify the timeout, where `timeout =
max_attempts * attempt_wait_time`.

---
closes #399
CentOS-PaaS-SIG#399

In addition to the job IDs, job URLs are now also printed out in the
course of provisioning. This debug print happens between the bkr_server
provisioning step and the bkr_info monitoring step, so users can monitor
or alter the job in beaker should they choose to do so.

---
closes #394
CentOS-PaaS-SIG#394

When bkr_info exceeds its `max_attempts` and times out, it issues a
job-cancel for the job being monitored in beaker, freeing up resources
that may have been provisioned during the request.

### DIFF
--- a/docs/source/examples/workspace/topologies/bkr-new.yml
+++ b/docs/source/examples/workspace/topologies/bkr-new.yml
@@ -5,7 +5,6 @@ resource_groups:
     resource_group_type: beaker
     resource_definitions:
       - role: bkr_server
-        whiteboard: Provisioned with linchpin
         job_group: ci-ops-central
         recipesets:
           - distro: RHEL-6.5
@@ -17,10 +16,21 @@ resource_groups:
               - tag: device
                 op: "="
                 type: "network"
+              - tag: hostname
+                op: "like"
+                value: "%.subdomain.example.tld"
             count: 1
       - role: bkr_server
-        whiteboard: Provisioned with linchpin
         job_group: ci-ops-central
+        # option to set job whiteboard message
+        whiteboard: Provisioned with linchpin
+        # options to configure amount of time spent polling beaker:
+        # 60 attempts with 30 seconds wait time between them,
+        # provisioning timeout is roughly 1800 seconds
+        max_attempts: 60
+        attempt_wait_time: 30
+        # option to set cancellation message if linchpin cancels provisioning
+        cancel_message: Job cancelled on account of rain
         recipesets:
           - distro: RHEL-6.5
             arch: x86_64
--- a/docs/source/examples/workspace/topologies/bkr-new.yml
+++ b/docs/source/examples/workspace/topologies/bkr-new.yml
@@ -6,6 +6,15 @@ resource_groups:
     resource_definitions:
       - role: bkr_server
         job_group: ci-ops-central
+        # option to set job whiteboard message
+        whiteboard: Provisioned with linchpin
+        # options to configure amount of time spent polling beaker:
+        # 60 attempts with 60 seconds wait time between them,
+        # provisioning timeout is roughly 3600 seconds
+        max_attempts: 60
+        attempt_wait_time: 60
+        # option to set cancellation message if linchpin cancels provisioning
+        cancel_message: Job cancelled on account of rain
         recipesets:
           - distro: RHEL-6.5
             arch: x86_64
@@ -16,21 +25,9 @@ resource_groups:
               - tag: device
                 op: "="
                 type: "network"
-              - tag: hostname
-                op: "like"
-                value: "%.subdomain.example.tld"
             count: 1
       - role: bkr_server
         job_group: ci-ops-central
-        # option to set job whiteboard message
-        whiteboard: Provisioned with linchpin
-        # options to configure amount of time spent polling beaker:
-        # 60 attempts with 30 seconds wait time between them,
-        # provisioning timeout is roughly 1800 seconds
-        max_attempts: 60
-        attempt_wait_time: 30
-        # option to set cancellation message if linchpin cancels provisioning
-        cancel_message: Job cancelled on account of rain
         recipesets:
           - distro: RHEL-6.5
             arch: x86_64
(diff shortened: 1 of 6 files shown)


linchpin/provision/library/bkr_info.py
```diff
@@ -107,21 +101,21 @@ def _check_jobs(self, ids):
...
     })
     beaker = BeakerTargets(mod.params)
     try:
         results = beaker.get_system_statuses()
-        mod.exit_json(hosts=results, changed=True, success=True)
     except Exception as ex:
-        mod.fail_json(msg=str(ex))
+        msg, results = ex.args
+        mod.warn(msg)
+    mod.exit_json(hosts=results, changed=True)
```
seandst: On second thought, calling exit_json here instead of fail_json means linchpin still exits 0 even if a provision fails, which seems bad. It's probably worth reverting this bit back to the old behavior and implementing a cleanup step that ensures all outstanding beaker provision requests are cancelled if any single one fails.

p3ck:
[test] enabled web hooks

p3ck:
[test] one more time to hopefully trigger jenkins

p3ck:
[test] one more time (hopefully)

docs/source/examples/workspace/topologies/bkr-new.yml
```diff
@@ -17,10 +16,21 @@ resource_groups:
               - tag: device
                 op: "="
                 type: "network"
+              - tag: hostname
+                op: "like"
+                value: "%.subdomain.example.tld"
```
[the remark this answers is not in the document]
seandst: I was not aware that the examples in the docs are used in integration tests, so I'll just remove this.

linchpin/provision/roles/beaker/tasks/provision_resource_group.yml
```diff
@@ -1,48 +1,45 @@
...
-    topology_outputs_beaker_server: "{{ topology_outputs_beaker_server + bkr_item.hosts }}"
-  with_items: "{{ _topology_outputs_beaker_server.results }}"
-  loop_control:
-    loop_var: bkr_item
+  - name: "append beaker outputs"
+    set_fact:
+      topology_outputs_beaker_server: "{{ topology_outputs_beaker_server + bkr_item.hosts }}"
+    with_items: "{{ _topology_outputs_beaker_server.results }}"
+    loop_control:
+      loop_var: bkr_item
```
seandst: @robled caught a bug here related to my last-minute change to take bkr_info out of a loop...there's no need to append beaker outputs when the beaker outputs are already combined in the single call to bkr_info. I'm testing a fix for this now, in addition to the other requested changes

herlo:
[test]

linchpin/provision/library/bkr_info.py
```diff
@@ -12,91 +11,82 @@
...
-        raise Exception("{0} system(s) never completed in {1}"
-                        " polling attempts. {2}".format(all_count - pass_count,
-                                                        attempts,
-                                                        dumps(job_results)))
+                return job_results
+            sleep(self.wait_time)
 
+        # max attempts exceeded, cancel jobs and fail
+        for job_id in _jprefix(self.ids):
+            self.hub.taskactions.stop(job_id, 'cancel',
```
jaypoulz: The one thing I would request as part of this changeset is that the cancellation of the beaker job also be attempted in the case of abort or cancel.

I think it would be best if the net result of a beaker provisioning transaction is either the job is running, the job was never submitted, or the job has been canceled.

I discussed this / the motivation for my reasoning in:
https://github.com/CentOS-PaaS-SIG/linchpin/issues/394#issuecomment-363466239

p3ck:
[merge]

Issue #380, opened 2017-10-23
Need to be able to pass reservetime to the reservation task during a beaker provision


danalanerh:
By default, the reserve time for a linchpin beaker task is 24 hours. There is a task parameter reservetime: that you can attach to the reserve task that will set the watchdog timer to be the time (in seconds) that you list instead of the default 24 hours.

seandst:
I've been looking into how to do this, and it's not entirely clear to me how to tell the python client to do this. I'm going to keep digging, but if you have an example of doing this with beaker's python API, I'm definitely interested.

seandst:
With the schema fix committed in https://github.com/CentOS-PaaS-SIG/linchpin/pull/483#issue-171488879, you can now add a "RESERVETIME" taskparam.

PR #483, opened 2018-02-26
Add taskparam support to beaker schema

seandst:
Turns out that bkr_server already had support for task params, we just
need to allow it in recipes; all other machinery is in place.

To add a custom reservesys time, add this to a recipe inside recipesets,
in the same place as things like hostrequires, distro, etc.
```yaml
        taskparam:
          - RESERVETIME=7200
```

Value is in seconds, so this is a 2-hour reservation.

closes: https://github.com/CentOS-PaaS-SIG/linchpin/issues/380


herlo:
[merge]

Issue #471, opened 2018-02-12
LinchPin should store beaker_job_id in the RunDB at runtime


jaypoulz:
# Overview
This issue is a follow-up case for when a beaker topology linchpin up could result in wasted beaker resources. After a discussion with @herlo, we've determined that we can fix this by stored the beaker_job_id in RunDB as soon as it becomes available.

## Problem Workflow
1. The Jenkins job is running, and linchpin is provisioning from beaker and has reached the step where it submits the beaker job.
2. An interrupt occurs. Perhaps this is caused by using Control-C on the machine, or perhaps, like my workflow, it is because a user cancels the Jenkins job.
3. The linchpin run ends without canceling the beaker job.
4. My code notices this and attempts to do a linchpin teardown on the new resource so that the job is canceled.
5. The teardown does not cancel the beaker job.

The job ends successfully, but with wasted beaker resources still provisioning.

## Solution
If we store the beaker_job_id in RunDB asap, the teardown call should successfully cancel the beaker job. This will save beaker resources, and result in a clean end state.

herlo:
Thanks @jaypoulz. Per our discussion, it should be pretty simple to add this into the RunDB from the beaker provider immediately after the beaker_job_id value is obtained.

herlo:
To be clear, this actually already happens during the gather_facts role, which occurs after the beaker provisioning role occurs. The problem here is that because an interrupt occurs, LinchPin (and ansible) do not trap this interrupt. However, simply adding the beaker_job_id to the RunDB record should allow for teardown to be successful whether or not the job was interrupted.

seandst:
This makes perfect sense to me, so I'll try to get this on my next sprint.

Issue #486, opened 2018-03-01
SPIKE: Create LinchPin External Context


herlo:
# Purpose

As part of the request in https://github.com/CentOS-PaaS-SIG/linchpin/issues/448, I've done some preliminary research, interviews, and brainstorming sessions.

It's become obvious that linchpin needs to provide additional components to make environmental variables available. Additionally, it seems there's a common need for a context to provide data to external resources, be it environment variables, LinchPin hooks, or some other method of exporting the values from within LinchPin after a provisioning/teardown action. As such, this proposal is being presented for comment.

## The LinchPin Context

Currently, there is a concept of a context within LinchPin. It's used to manage all of the variables generated from the linchpin.conf, settings from the CLI shell, incoming environment variables, and various other generated inputs, such as the uhash, inventory_path, etc. This context serves the purpose of also exporting these variables into the extra_vars which are passed when executing the internal playbooks. 

The LinchPin Context is useful internally, but could be useful to external scripts as well. Some of the data provided is sensitive, other data may not be relevant to external scripts. To that end, it may be useful to massage the data to remove this information.

Although it's not likely to be a large part of this discussion, an additional consideration for this context is developers who may create new providers in the future. Knowing what values are available as inputs, and what is required to output to the context will be useful.

## Namespacing

As with many of the variables in Ansible, LinchPin context variables will have a namespace. Currently, this is not the case, but this work will attempt to ensure any context variables will be prepended with 'linchpin_' so as to not conflict with ansible specific variables in the playbooks, or with any other applications which may use similar values in their external application.

## Initial Work

The initial phase is to provide a mechanism for adding and exporting these values from the context. The RunDB is a perfect resource for storing this data as it's already collecting most of it already. Items that are currently not in the RunDB include relevant extra_vars, and inventory data. There may be other data included as more investigation occurs.

Once the data is stored in the RunDB on each run, each variable will be added to the context object. If a variable doesn't have the 'linchpin_' namespace, it will be updated where applicable. The object will then be exported to the requested environment contexts. Currently, those are environment and hooks contexts, but this may be expanded in the future. 

### The Linchpin ID

Currently, LinchPin provides a 'run_id' for each execution of LinchPin. Usually, this is on a target by target basis, and is unique in terms of each target and run_id combination. This is convenient to reuse the data from the previous provision (up) action to teardown (destroy) the same resources. However, it is possible to provision multiple targets at one time. To that end, a new mechanism is warranted to track each invocation.

[...]

PR #492, opened 2018-03-08
Stash beaker job info in rundb immediately

seandst:
This tweaks some already duplicated code in bkr_server to provide
outputs that match the same format as bkr_info, which lets us store
usable topology information in the rundb immediately after creating
beaker jobs. This makes "linchpin destroy" still work even if later
beaker provisioning steps fail, when using rundb.

closes https://github.com/CentOS-PaaS-SIG/linchpin/issues/471

### DIFF
--- a/linchpin/provision/library/bkr_server.py
+++ b/linchpin/provision/library/bkr_server.py
@@ -185,23 +185,20 @@ class BkrFactory(BkrConn):
         return kwargs
 
     def check_jobs(self, jobs):
-        """
-            Get state of a job in Beaker
-        """
-        results = {}
+        # slightly modified copy of the same method in bkr_info, used here to
+        # let bkr_server return beaker job info in the same structure as
+        # bkr_info so we can stash provisioning details in rundb before the
+        # bkr_info task blocks on waiting for the beaker job(s) to finish.
+        # Importing code from bkr_info is not an option due to ansible's
+        # sandboxing; bkr_info is not guaranteed to be importable at runtime.
         resources = []
         bkrcmd = BeakerCommand('BeakerCommand')
         bkrcmd.check_taskspec_args(jobs)
         for task in jobs:
             myxml = self.hub.taskactions.to_xml(task)
             myxml = myxml.encode('utf8')
-            LOG.debug(xml.dom.minidom.parseString(myxml)
-                         .toprettyxml(encoding='utf8'))
             root = eT.fromstring(myxml)
-            # Using getiterator() since its backward compatible with Python 2.6
-            for job in root.getiterator('job'):
-                results.update({'job_id': job.get('id'),
-                                'results': job.get('result')})
+            # Using getiterator() since its backward compatible with py26
             for recipe in root.getiterator('recipe'):
                 resources.append({'family': recipe.get('family'),
                                   'distro': recipe.get('distro'),
@@ -209,9 +206,9 @@ class BkrFactory(BkrConn):
                                   'variant': recipe.get('variant'),
                                   'system': recipe.get('system'),
                                   'status': recipe.get('status'),
-                                  'result': recipe.get('result')})
-                results.update({'resources': resources})
-        return results
+                                  'result': recipe.get('result'),
+                                  'id': recipe.get('job_id')})
+        return resources
 
     def cancel_jobs(self, jobs, msg):
         """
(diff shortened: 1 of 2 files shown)


linchpin/provision/library/bkr_server.py
```diff
@@ -185,33 +185,30 @@ def create_recipesets(self, recipeset, **kwargs):
...
                                   'arch': recipe.get('arch'),
                                   'variant': recipe.get('variant'),
                                   'system': recipe.get('system'),
                                   'status': recipe.get('status'),
-                                  'result': recipe.get('result')})
-                results.update({'resources': resources})
-        return results
+                                  'result': recipe.get('result'),
+                                  'id': recipe.get('job_id')})
+        return resources
```
seandst: These changes to check_jobs are a dirty copy/paste of the same method in bkr_info, as explained in the comment. If anyone has less unsightly ways to share common code in ansible that don't end up requiring complete refactors of both modules, I'm interested.

herlo:
@seandst can you provide what the RunDB json looks like with this after one beaker job, then two complete? I know this would write to the RunDB multiple times, it would be good to see what a simple run might populate. Thanks!

seandst:
I'm assuming that the diff of the two suffices here, but can post the full json files if needed.
```diff
--- rundb-1-pretty.json 2018-03-13 12:25:45.487390849 -0400
+++ rundb-2-pretty.json 2018-03-13 12:25:37.054033147 -0400
@@ -89,6 +89,95 @@                         
     }                      
    ],                 
    "rc": 0             
+  },                    
+  "2": {               
+   "uhash": "640b",        
+   "inputs": [
+    { 
+     "topology_data": {
+      "topology_name": "smyers",
+      "resource_groups": [
+       {
+        "resource_group_name": "node",
+        "resource_definitions": [         
+         {                 
+          "job_group": "ci-ops-central",                     
+          "whiteboard": "linchpin teardown testing",
+          "role": "bkr_server",
+          "recipesets": [
+           {               
+            "count": 1,
+            "arch": "x86_64",
+            "hostrequires": [
+             {
+              "tag": "hostname",
+              "value": "host%",
+              "op": "like"
+             }
+            ],            
+            "variant": "Server",                                                                                      
+            "distro": "RHEL-7.4"
+           }
+          ]
+         }
+        ],
+        "resource_group_type": "beaker"
+       }
+      ]
+     }
+    },
+    {
+     "layout_data": {
+      "inventory_layout": {
+       "hosts": {
+        "cinch-group": {
+         "count": 1,
+         "host_groups": [
+          "rhel7",
+          "certificate_authority",
+          "repositories",
+          "jenkins_master"
+         ]
+        }
+       }
+      }
+     }
+    }
+   ],
+   "end": "03/13/2018 10:55:21 AM",
+   "start": "03/13/2018 10:30:57 AM",
+   "action": "up",
+   "outputs": [
+    {
+     "resources": [
+      {
+       "status": "New",
+       "family": "RedHatEnterpriseLinux7",
+       "variant": "Server",
+       "system": null,
+       "result": "New",
+       "arch": "x86_64",
+       "id": "2370390",
+       "distro": "RHEL-7.4"
+      }
+     ]
+    },
+    {
+     "resources": [
+      {
+       "status": "Running",
+       "family": "RedHatEnterpriseLinux7",
+       "variant": "Server",
+       "system": "host2",
+       "result": "Pass",
+       "arch": "x86_64",
+       "id": "2370390",
+       "distro": "RHEL-7.4"
+      }
+     ]
+    }
+   ],
+   "rc": 0
   }
  }
 }
```

Suffice it to say that the first run looks just like this one, except for the part where its ID is 1, not 2, it has a different uhash (ab90 in this case), and it provisioned a different host. Also timestamps, of course.

There's some minor weirdness where "outputs" ends up with two entries instead of one, but beaker is happy to cancel the same job multiple times (we already use/abuse this in beaker teardown), so everything still appears to work fine. My rundb-fu is weak, so ideas on how to fix this without having to do crazy things would be useful. That second "outputs" entry comes from existing machinery, and would ideally overwrite the first.

`linchpin teardown -r <n>` does what you'd expect, so that's good. :)

herlo:
@seandst, ideally, you wouldn't have two run_id values for a single run, but I'm not sure it causes any problems in this use case. What I was actually looking for is the difference between hitting the RunDB entry in the beaker module and at the end. However, I think this gives enough info to go on, so thank you.

In the very near future, I'll be standardizing how linchpin will put data into the RunDB. In the meantime, while this creates extra data in the RunDB, it doesn't seem like it will affect the performance of anything and the code works. Thanks!

seandst:
Yeah, I thought it was likely that this was all going to potentially change with https://github.com/CentOS-PaaS-SIG/linchpin/issues/486, so I went with the "minimally invasive" approach here, knowing that beaker would be just fine with canceling all the job IDs we can throw at it in the destroy phase.

Issue #487, opened 2018-03-05
Document what is required for each provider topology


herlo:
Each provider topology has a list of the variables and whether they are required or not in the files/schema.json file within the role definition. However, it's not documented what is required and what is not in each case. This issue is filed to update the providers documentation with the fields available and their required or optional status.

PR #494, opened 2018-03-13
Added support for attaching volumes to provisioned instances.

### LINKED ISSUE #479: Openstack provider should be able to create and attach volumes to a VM
We have a use case where we need to run a test with a VM that has a large volume.
We would like to be able to specify this need in the openstack topology.

parameters to define:

1. Volume size
2. Volume name

samvarankashyap:
Updated the schema.json and playbooks for OpenStack to support list of pre-provisioned volumes
names as a list of attachments to provisioned instances. No new changes were added to os_server2
module since attaching volumes is already supported.
fixes #479

### DIFF
--- a/docs/source/openstack.rst
+++ b/docs/source/openstack.rst
@@ -15,6 +15,47 @@ Openstack instances can be provisioned using this resource.
    the variables used are identical to the Ansible os_server module, except for
    adding a ``count`` option.
 
+Topology Schema
+~~~~~~~~~~~~~~
+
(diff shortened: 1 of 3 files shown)


herlo:
[test]

herlo:
As part of #487, it would be good to document this provider in the Providers section. I've provided an example of what should be included. It can likely just be cut and pasted, then added to this PR. 

https://github.com/herlo/linchpin/blob/topology_schema_detail/docs/source/aws.rst#topology-schema

Thanks!

herlo:
[merge]

PR #496, opened 2018-03-13
rework how credentials are used.

p3ck:
This makes it much more generic to support drivers that need
authentication passed in.

Support is there for testing aws but the credential code needs to be
fixed first.

This may break duffy testing if the same CREDS_PATH issue is present in the duffy driver.


herlo:
[merge]

Issue #499, opened 2018-03-15
LinchPin Context Distiller


herlo:
In line with #486 and per conversations with @flowersncsu @robnester-rh, an outputter is needed to capture and provide specified data to external sources. This data will consist of generic components as described in #486 as well as provider specific, and requested information from specific LinchPin transactions.

While RunDB can hold all outputs, the distiller will allow for capturing specified data on a provider by provider basis. The distiller for now will be simple, but provide a way for a user to have common (and requested) data extracted from the RunDB, and put into output. Output formats could be environment variables, hooks context, etc. 

In this specific case, the request is for a file, placed into the workspace, containing data from an AWS run.

The file will include, but not limited to the following bits of data for the AWS provider:

* private_dns_name
* public_dns_name
* public_ip
* private_ip
* instance_ids

This data can be extracted from the RunDB at this time, and placed in a file of the user's choosing (how that is determined is still being investigated).

robnester-rh:
To be clear, the ask wasn't limited to AWS, but would be required for any / all providers. The ask for the moment is, at a minimum, aws and beaker.

herlo:
@robnester-rh indeed, which is why the 'not limited to' state is in that specific line about AWS. I intend to start with AWS as proof of concept for you, but I expect all providers to be able to be distilled and output in some form or fashion.

herlo:
Fixed by #525. The roles which resources are distilled currently include: 

* bkr_server
  * id, url, system
* aws_ec2
  * instances.id, instances.public_ip, instances.private_ip, instances.public_dns_name, instances.private_dns_name, instances.tags:name
* os_server
  * servers.id, servers.interface_ip, servers.name, servers.private_v4, servers.public_v4
* dummy_node (testing only)
  * hosts

PR #507, opened 2018-03-19
Transactions in linchpin journal

### LINKED ISSUE #501: Add views to linchpin journal
In the process of working on #486, providing outputs based upon a complete linchpin transaction is necessary for reporting and querying data.

Per that SPIKE issue, the ide of a LINCHPIN_ID came about, making it so that a user could gather data from the Linchpin RunDB in a useful manner. The LinchPin Context Distiller and Outputter #499 tooling will take advantage of this LINCHPIN_ID and provide output based upon defaults and user requests (design still to come on the user specific bits). 

The views will be expanded in the future, but for now, there will be a 'target' (default) and exec views. The target view will be the same as it is now, showing the output from the latest provisioning actions for each target listed (or all if no target is passed). The exec view will show a specific LinchPin transaction, and provide the Linchpin 'ID' value.

### Target view
```
$ linchpin journal dummy-new

Target: dummy-new
run_id	    action	     uhash	        rc	
--------------------------------------------------
211    	       up	     89ea	        0	
210    	       up	     bb24	        0	
209    	       up	     c37c	        0	
```

### Transactional view

```
$ linchpin journal --view=tx
ID: 44
Action: up

dummy-new           	   211	 89ea	        0
libvirt-new         	   219	 ba82	        0
```

The purpose of this functionality currently may not be clear to the casual reader, or even the standard user of LinchPin. However, the idea is to use this functionality to find the proper transaction. The data would then provide a way to gather outputs from said transaction in useful ways. One example is to be able to generate more than just ansible static inventories. Another useful way is to have 'linchpin journal' provide more options like generating json outputs with additional views.

herlo:
Fixes #501 and begins prep for #499

### DIFF
--- a/docs/source/examples/workspace/topologies/dummy.yml
+++ b/docs/source/examples/workspace/topologies/dummy.yml
@@ -1,12 +1,12 @@
 ---
-topology_name: "dummy_cluster" # topology name
+topology_name: dummy_cluster # topology name
 resource_groups:
-  - resource_group_name: "dummy"
-    resource_group_type: "dummy"
+  - resource_group_name: dummy
+    resource_group_type: dummy
     resource_definitions:
-      - name: "web"
-        type: "dummy_node"
+      - name: web
+        type: dummy_node
         count: 3
-      - name: "test"
-        type: "dummy_node"
+      - name: test
+        type: dummy_node
         count: 1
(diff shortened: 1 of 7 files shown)


herlo:
Provides a new view in linchpin journal called 'tx' (for transactions). The benefit of this view is that relationships between targets is much more clear. Eventually, the linchpin journal will be able to provide a view based upon the ID given (130 in this example) which would display resource data, inventories, etc. upon request.

```
$ linchpin journal --view tx --count 1

ID: 130			Action: up

dummy-new           	   279	 920c	        0
libvirt             	   121	 ef96	        0

=================================================
```

herlo:
[merge]

PR #519, opened 2018-03-28
Final pieces to make beaker testing possible

p3ck:
Tested locally with a bkr-new.tgz file in
keys/bkr-new which contains the following files:

bkr-new.keytab
client.conf
install.sh
RedHatInternalCA.pem


herlo:
[merge]

PR #525, opened 2018-04-06
Context Distiller

herlo:
The Context Distiller takes results data from provisioning resources, and distills it down into json data based upon fields that are set in the linchpin.constants. To enable this feature, the lp.distill_data flag needs to be set in linchpin.conf.


linchpin/__init__.py
```diff
@@ -707,6 +705,69 @@ def do_action(self, provision_data, action='up', run_id=None, tx_id=None):
         return (return_code, lp_data)
 
 
+    def get_run_data(self, tx_id, fields, targets=()):
+        """
+        Returns the RunDB for data from a specified field given either a run_id
+        or tx_id. The fields consist of the major sections in the RunDB (target
```
p3ck: Is this correct?  it says either a run_id or tx_id but I only see tx_id passed in to the method.

linchpin/cli/__init__.py
```diff
@@ -173,6 +174,99 @@ def _get_data_path(self):
...
+            try:
+                for res in resources:
+                    for k, v in fields.iteritems():
+                        res_data = {}
+                        if not v:
+                            dd = dist_data.get(target, {})
+                            if len(dd):
+                                for items in dd:
+                                    if k in items.keys():
+                                        items[k].extend(res.get(k))
```
p3ck: Can you explain what this loop is doing?  Seems a little odd to me to be modifying the dictionary you are iterating over.

herlo:
@p3ck this passed and could use a review. Thanks.

linchpin/cli/__init__.py
herlo: Sure. There are actually two dictionaries here. The first pulls out the fields which need to be grabbed from the particular provider. The second matches those items and  puts them into the dictionary.

linchpin/__init__.py
herlo: tx_id holds the run_ids for a particular run (or transaction). To be able to get the data, the tx_id is used to get the run_ids. In turn, the resource data is pulled by the run_id into a returnable format.

herlo: I see what you mean. I missed the point. The docs just need to remove the reference to the run_id. I'll get that fixed in the next update coming shortly.

herlo:
Ignore the above comment. I was using the mobile version and it formats things a little differently.

linchpin/cli/__init__.py
p3ck: ok - I get it now.

herlo:
[merge]

Issue #527, opened 2018-04-09
AWS fails on centos7


p3ck:
TASK [aws : Provisioning AWS_EC2 Resource when not async] ****************************
fatal: [localhost]: FAILED! => {"msg": "template error while templating string: no fil
ter named 'selectattr'. String: {{ res_grp_vars | selectattr('resource_group_name', 'e
qualto', res_grp_name ) | first }}"}

It looks like jinja2 on centos7 is not new enough to include this function.

p3ck:
this was fixed with #537

PR #537, opened 2018-04-11
use match instead of equalto.

p3ck:
rhel/centos 7 has an older version of jinja2 which doesn't include equalto.


p3ck:
fixes #527

p3ck:
[merge]

PR #541, opened 2018-04-13
Add support for passing instance_tags for aws driver.

### LINKED ISSUE #503: Need instance_tags on linchpin AWS instances
As per the ansible EC2 module documentation (http://docs.ansible.com/ansible/latest/ec2_module.html), instance_tags is an available parameter, but linchpin does not accept that parameter. We need that parameter available.

Example usage:
```
instance_tags:
  color: blue
  shape: oval
```

p3ck:
fixes #503


p3ck:
[merge]

PR #546, opened 2018-04-19
Use pytest on travis

herlo:
jsonschema is no longer required ( schemacheck ansible module still needs to be removed)
pytest finds and performs tests, updated this in the .travis.yml as well.
testing shade 1.18 for compatibility with F27 distro packaging (EPEL doesn't have shade at all, only pip)


herlo:
[merge]

PR #552, opened 2018-04-24
Add auto_ip option to os_server resources

### LINKED ISSUE #522: add option to toggle auto_ip for openstack deployments
Provisioning on Openstack will by default assign a floating IP even when fip isn't set in topology. To avoid that auto_ip: false needs to also be send to the os_server module. 

Please provide a way to set auto_ip so that it will be possible to provision instances without a floating IP assigned

herlo:
Fixes #522 

Updated os-server-new.yml topology to test 'auto_ip: no' setting
Made networks a little cleaner to deal with when they are left empty


herlo:
[merge]

PR #556, opened 2018-04-25
This changes the selective target testing logic slightly.

p3ck:
- don't break out of the target selection routine when you get a match.
  This allows multiple targets to match for the same file.
- Change the dummy regex to match all files except linchpin/version.py
  This means we will only test the dummy target for non-target specific
  changes.  But if the linchpin/version.py file is changed we will test
  every target.  This is a good compromise for speed for every PR
  and completeness when doing a release.


johnbieren:
[test]

p3ck:
[merge]

PR #559, opened 2018-04-26
Remove CRUFTY bits and use git for pypi packages

herlo:
After releasing v1.5.3a1, I noticed that items in docs/source/examples/workspace/inventories/ were being packaged up even though that data is from specific runs of linchpin during testing.

This new create_pypi_package.sh checks out from git.

The requirements are that the tags be pushed to upstream prior to running this command. If not, the creation of packages will fail.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,8 @@
 ansible.cfg
 *.swp
 .fuse*
+.cache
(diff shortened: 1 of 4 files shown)


herlo:
[merge]

PR #561, opened 2018-04-28
Add openstack tests

14rcole:
Add openstack tests for provisioning compute instances, cinder volumes, and security groups


herlo:
[merge]

Issue #563, opened 2018-04-30
Linchpin does not work with ansible 2.3.x 


samvarankashyap:
The following error occurs when we run linchpin up with dummy topology
```
Traceback (most recent call last):
  File "/home/srallaba/virtualenvs/test/bin/linchpin", line 11, in <module>
    sys.exit(runcli())
  File "/home/srallaba/virtualenvs/test/lib/python2.7/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/home/srallaba/virtualenvs/test/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/home/srallaba/virtualenvs/test/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/srallaba/virtualenvs/test/lib/python2.7/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/srallaba/virtualenvs/test/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/home/srallaba/virtualenvs/test/lib/python2.7/site-packages/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args[1:], **kwargs)
  File "/home/srallaba/virtualenvs/test/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/home/srallaba/virtualenvs/test/lib/python2.7/site-packages/linchpin/shell/__init__.py", line 239, in up
    tx_id=tx_id)
  File "/home/srallaba/virtualenvs/test/lib/python2.7/site-packages/linchpin/cli/__init__.py", line 306, in lp_up
    tx_id=tx_id)
  File "/home/srallaba/virtualenvs/test/lib/python2.7/site-packages/linchpin/cli/__init__.py", line 421, in _execute_action
    run_id=run_id)
  File "/home/srallaba/virtualenvs/test/lib/python2.7/site-packages/linchpin/cli/__init__.py", line 554, in _execute
    tx_id=tx_id)
  File "/home/srallaba/virtualenvs/test/lib/python2.7/site-packages/linchpin/__init__.py", line 651, in do_action
    console=ansible_console)
[...]

PR #564, opened 2018-04-30
Bugfix: Making linchpin work with Ansible 2.5.x

samvarankashyap:
PlaybookCallback defined does not have attribute _loader_name which is
required by the ansible executor.py is added in constructor before the
super call. Further, reverted ansible version in requirements.txt
back to >=2.3.1.
Fixes #564 #563


samvarankashyap:
fixes #564 #563

p3ck:
[merge]

PR #568, opened 2018-05-02
Return the rc's based on config varible distill_on_error

samvarankashyap:
Returns 0 even if one target provision is successfull when
distill_on_error = True
Returns rc > 0 even if one target provision fails when
distill_on_error = False

### DIFF
--- a/docs/source/examples/workspace/linchpin.conf
+++ b/docs/source/examples/workspace/linchpin.conf
@@ -91,6 +91,11 @@ distill_data = True
 # data upon failure).
 distill_on_error = False
 
+# User can make linchpin use the actual return codes for final return code
+# if enabled True, even if one target provision is successfull linchpin 
+# returns exit code zero else returns the sum of all the return codes
+# use_actual_rcs = False
+
 # LinchPin sets several extra_vars (evars) that are passed to the playbooks.
 # This section controls those items.
 [evars]
--- a/linchpin/cli/__init__.py
+++ b/linchpin/cli/__init__.py
@@ -209,8 +209,8 @@ class LinchpinCli(LinchpinAPI):
                     roles.append(rd.get('role'))
 
             fields = {}
-            outputs = data.get('outputs')
-            resources = outputs.get('resources')
+            outputs = data.get('outputs', {})
+            resources = outputs.get('resources', [])
 
             for dist_role, flds in dist_roles.iteritems():
                 if dist_role in roles:
@@ -257,7 +257,6 @@ class LinchpinCli(LinchpinAPI):
                         dist_data[target].append(res_data)
             except Exception:
                 pass
-
         with open(context_file, 'w+') as f:
             f.write(json.dumps(dist_data))
 
--- a/linchpin/linchpin.constants
+++ b/linchpin/linchpin.constants
@@ -26,6 +26,7 @@ external_providers_path = %(default_config_path)s/linchpin-x
 use_rundb_for_actions = False
 distill_data = False
 distill_on_error = False
+use_actual_rcs = False
 
 [evars]
 _check_mode = False
--- a/linchpin/shell/__init__.py
+++ b/linchpin/shell/__init__.py
@@ -4,6 +4,7 @@ from __future__ import print_function
 
 import os
 import sys
+import ast
 import json
 import click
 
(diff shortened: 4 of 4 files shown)


linchpin/cli/__init__.py
```diff
@@ -255,9 +255,8 @@ def _write_distilled_context(self, run_data):
 
                     if len(res_data) and res_data not in dist_data[target]:
                         dist_data[target].append(res_data)
-            except Exception:
-                pass
-
+            except Exception as e:
+                self.ctx.log_state('Error: {0}'.format(e))
```
p3ck: Thanks you! ^^

linchpin/shell/__init__.py
```diff
@@ -91,12 +89,17 @@ def _handle_results(ctx, results, return_code):
 
     # PRINT OUTPUT RESULTS HERE
     ctx.log_state(output)
+    distil_on_error = json.loads(lpcli.get_cfg('lp',
```
herlo: distill has two l's, probably ought to fix that so it is a proper spelling.

linchpin/shell/__init__.py
```diff
@@ -91,12 +89,17 @@ def _handle_results(ctx, results, return_code):
...
     # PRINT OUTPUT RESULTS HERE
     ctx.log_state(output)
+    distill_on_error = json.loads(lpcli.get_cfg('lp',
+                                                'distill_on_error').lower())
 
-#   FIXME: have use_actual_rcs be a flag
+    #   FIXME: have use_actual_rcs be a flag
     use_actual_rcs = True
-    if use_actual_rcs:
+    if use_actual_rcs and not(distill_on_error):
```
herlo: I see you do have 'use_actual_rcs', but that should come from the linchpin.constants (or overridden by the linchpin.conf) and set to a default. Not hard coded here. Additionally, distill_on_error is a separate function and should occur if and only if that flag is set, it shouldn't be dependent on whether 'use_actual_rcs' is set. This would mean that one is dependent on the other to be true for certain conditions.

linchpin/shell/__init__.py
```diff
@@ -91,10 +89,12 @@ def _handle_results(ctx, results, return_code):
 
     # PRINT OUTPUT RESULTS HERE
     ctx.log_state(output)
+    use_actual_rcs = json.loads(lpcli.get_cfg('lp',
+                                              'use_actual_rcs').lower())
```
herlo: If you are using True/False, the proper use of this is to use ast.literaleval(value.title()), https://github.com/CentOS-PaaS-SIG/linchpin/blob/develop/linchpin/cli/__init__.py#L323 is a good example.

linchpin/linchpin.constants
```diff
@@ -26,6 +26,7 @@ external_providers_path = %(default_config_path)s/linchpin-x
 use_rundb_for_actions = False
 distill_data = False
 distill_on_error = False
+use_actual_rcs = False
```
herlo: This also needs a corresponding example and explanation in the https://github.com/CentOS-PaaS-SIG/linchpin/blob/develop/docs/source/examples/workspace/linchpin.conf

samvarankashyap:
@herlo: changes made.

docs/source/examples/workspace/linchpin.conf
```diff
@@ -91,6 +91,11 @@ distill_data = True
 # data upon failure).
 distill_on_error = False
 
+# User can make linchpin use the actual return codes for final return code
+# if enabled True, even if one target provision is successfull linchpin 
+# returns exit code zero else returns the sum of all the return codes
+use_actual_rcs = False
+
```
herlo: Are you overriding what is in linchpin.constants here? It appears not, so it would be prudent to just comment it out.

linchpin/shell/__init__.py
```diff
@@ -91,10 +90,12 @@ def _handle_results(ctx, results, return_code):
 
     # PRINT OUTPUT RESULTS HERE
     ctx.log_state(output)
+    use_actual_rcs = ast.literal_eval(lpcli.get_cfg('lp',
+                                                    'use_actual_rcs'))
```
herlo: as a rule, you should add the .title() function to the item to get the 'True' or 'False' state before it's evaluated by the ast.literal_eval method. Something like this would be appropriate.

```
uar = lpcli.get_cfg('lp', 'use_actual_rcs'))
use_actual_rcs = ast.literal_eval(uar.title())
```
In this way, if the value is 'true', or 'false', or any combination thereof, the value will get converted to proper title case, and return the proper boolean.

samvarankashyap: Okay , I was not aware of the rule . I tried ast.literal_eval(lpcli.get_cfg('lp','use_actual_rcs')) and it worked . Will make the change.

samvarankashyap:
[merge]

PR #572, opened 2018-05-07
Openstack Security Group testing fix

herlo:
The code below should continue to run all tests as normal. But the os-sg-new test, which creates an openstack security group will have the distro prepended to the name of the security group. This should help with the race condition discovered in [#571#issuecomment-387111515](https://github.com/CentOS-PaaS-SIG/linchpin/pull/571#issuecomment-387111515) when running all distros in parallel.

There was also a small bug fixed in the linchpin cli code which was not passing the template-data to all components of the PinFile structure.


herlo:
[merge]

PR #574, opened 2018-05-08
Reinstate --output-pinfile

herlo:
During the rework of the context distiller, the code for the --output-pinfile option was inadvertently removed. This patch puts the functionality back in place.


herlo:
[merge]

PR #578, opened 2018-05-09
JenkinsfileRelease releases to test.pypi now

herlo:
python-twine performs releases using credentials provided by ID 'linchpin-releaser'. The paas-bot user is also needed to perform tag pushes to github.
    
Tests were performed against https://github.com/herlo/jenkins-test.git


config/Dockerfiles/JenkinsfileRelease
```diff
@@ -0,0 +1,182 @@
...
+
+[pypitest]
+repository:https://test.pypi.org/legacy/
+username = ${TWINE_USERNAME}
+password = ${TWINE_PASSWORD}
+EOF
+                                """
+                                )
+                            }
+                            sh "python setup.py sdist bdist_wheel"
```
p3ck: I would just combine all these sh commands into the above sh block.

config/Dockerfiles/JenkinsfileRelease
```diff
@@ -0,0 +1,182 @@
...
+                            }
+
+                        }
+                    }
+                    stage('release-version') {
+                        dir(repoName) {
+
+                            string version = sh (
+                                script: "cat linchpin/version.py | grep __version__ | cut -d' ' -f3 | cut -d\"'\" -f2",
+                                returnStdout: true
```
p3ck: while this works it feels fragile to me and would rather cat out a python file to import this version file and simply print it.

cat << EOF >return_version
#!/usr/bin/python
import linchpin.version
print linchpin.version.__version__
EOF

Then the command would be:
string version = sh (
   script: "./return_version", returnStdout: true
)

config/s2i/jenkins/slave/Dockerfile
```diff
@@ -1,4 +1,7 @@
 FROM openshift/jenkins-slave-base-centos7:v3.7
 RUN yum install -y epel-release
+RUN yum install -y gcc python-devel libyaml-devel
```
p3ck: combine these RUN commands so we don't create so many layers in our docker images. And we should also yum -y clean all to clean up the database

config/Dockerfiles/create_pypi_package.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
```
p3ck: What is the reason for moving the package from script to config/Dockerfiles?

p3ck: Also - are we not using this file anymore?  I see everything is mostly commented out.  is the twine command doing the upload now?  Just trying to make sure I understand it all. :-)

herlo: Yeah. I just need to remove this file. I'll update that shortly, in addition to a couple other cleanups mentioned above.

config/Dockerfiles/JenkinsfileRelease
herlo: Actually, there's a setup.py way to get this, I'll adjust that.

config/Dockerfiles/JenkinsfileRelease
```diff
@@ -0,0 +1,182 @@
...
+
+import com.cloudbees.plugins.credentials.Credentials
+
+// Execution ID for this run of the pipeline
+executionID = UUID.randomUUID().toString()
+// Use repoName everywhere for dirs and repo values
+repoName = 'jenkins-test'
+
+
+env.ghprbGhRepository = env.ghprbGhRepository ?: 'herlo/' + repoName
```
p3ck: I know you are testing...  but make sure this changes for the final version of the patch.

config/Dockerfiles/JenkinsfileRelease
```diff
@@ -0,0 +1,182 @@
...
+                              remote: "https://github.com/CentOS-Paas-SIG/ci-pipeline"])
+
+library identifier: "cico-pipeline-library@master",
+        retriever: modernSCM([$class: 'GitSCMSource',
+                              remote: "https://github.com/CentOS/cico-pipeline-library"])
+
+
+properties([
+  buildDiscarder(logRotator(artifactNumToKeepStr: '20', numToKeepStr: '20')),
+  [$class: 'GithubProjectProperty', displayName: '', projectUrlStr: 'https://github.com/herlo/' + repoName],
```
p3ck: this needs to change as well..  why not use the variable you defined above?

config/Dockerfiles/JenkinsfileRelease
```diff
@@ -0,0 +1,182 @@
...
+      string(defaultValue: '',
+             description: 'Pull Request Number',
+             name: 'ghprbPullId'),
+      string(defaultValue: '',
+             description: 'Pull Request Author username',
+             name: 'ghprbPullAuthorLogin'),
+      string(defaultValue: '',
+             description: 'Git Hub Repository',
+             name: 'ghprbGhRepository'),
+      string(defaultValue: 'stable',
```
p3ck: you can take this variable out since you don't use it below.  Or update the containerTemplate below to use it.

herlo: @p3ck I'm not sure how I might use this in the containerTemplate, assuming you are meaning the ghprbGhRepository variable. But removing it would mean that I can't use it below, correct?

p3ck: I'm talking about the stable tag.

herlo:
@p3ck not sure why this continuing to fail. Maybe I did something wrong?

herlo:
[merge]

herlo:
[merge]

PR #586, opened 2018-05-11
github_issue_547: Basic support for remote libvirt hypervisors

joanbe:
Minimal set of changes to enable support for remote libvirt hypervisors.

PLEASE NOTE:
The changes to the mode of the '{{ local_image_path }}' and the explicit path to the 'arp' binary are highly dependent on my environment and should be reviewed before merge. There are very likely better ways to do this.


joanbe:
There is (at least) one more issue that I have not been able to solve: the 'ansible_python_interpreter' seems to be pointing to the python path on the local machine (the machine running 'linchpin up' command) through the entire play. This implies that the remote hypervisor must provide a python binary in the same path as on the local machine. When using a python virtual environment on the local machine this becomes quite messy.

p3ck:
[test]

p3ck:
[test]

linchpin/provision/roles/libvirt/tasks/provision_libvirt_node.yml
```diff
@@ -307,7 +307,7 @@
 
 - name: "mac_and_ip | wait up to 5 mins for dhcp ip address"
   shell: |
-    arp -an | grep -F {{ extract_mac_address_result.results[mac].stdout }} | cut -f 2 -d "(" | cut -f 1 -d ")"
+    /usr/sbin/arp -an | grep -F {{ extract_mac_address_result.results[mac].stdout }} | cut -f 2 -d "(" | cut -f 1 -d ")"
```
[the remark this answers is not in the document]
joanbe: I replied via email but I'm not entirely sure that worked so lets duplicate that reply here :)

It is my understanding that the remote user is set via {{
res_def['remote_user'] }}. In my case this is not 'root' but a user in
the libvirt and qemu groups (thus allowed to provision virtual
resources). This user apparently does not have 'arp' in $PATH.

The relevant configuration entries from my topology file are:

  uri: qemu+ssh://admin@remote-kvm-host.example.com/system
  remote_user: admin

I would ideally like to run this playbook without (too many) privileges.
However, since it is already a requirement for the remote_user to be a
member of certain groups perhaps it's not a stretch to require that
remote_user to have /usr/sbin in the $PATH.

herlo:
[test]

herlo:
@joanbe would you mind rebasing this against the develop branch again? We just did an update to our tests!

linchpin/provision/roles/libvirt/tasks/copy_image_src_remote.yml
```diff
@@ -7,7 +7,7 @@
   file:
     path: "{{ local_image_path }}"
     state: directory
-    mode: 0755
+    mode: 0770
```
herlo: I understand why you took away access from the others, but why is the 770 necessary? If writing is necessary, the user rights would match first anyway. Thus, adding the second 7 there would only be for additional users of whatever group the remote_user belongs to and would provide full directory access to the 'local_image_path'.  Is that necessary?

linchpin/provision/roles/libvirt/tasks/copy_image_src_remote.yml
```diff
@@ -48,41 +48,41 @@
 - name: "set image_src for uncompressed files"
   set_fact:
     img_src: "{{ local_image_src }}"
-  when: lismt_remote.stat.mime_type == "application/octet-stream"
+  when: lismt_remote.stat.mimetype == "application/octet-stream"
```
herlo: I'm assuming there's a good reason for this change? The mime_type still works as an alias according to [the ansible stat module (v2.5)](http://docs.ansible.com/ansible/latest/modules/stat_module.html). Even older versions still have mime_type listed under get_mime. Though I did see a bug reported that the documentation was incorrect, but not yet fixed.

joanbe: As seen below the output of the stat module have no mime_type keyword:

    "msg": {
        "changed": false,                                                                                                                                                                                          
        "failed": false,                                                                  
        "stat": {                                                                                                                                                                                                  
            "atime": 1527073187.537445,                                                                                                                                                                            
            "attr_flags": "e",                                                                                                                                                                                     
            "attributes": [                                                                                                                                                                                        
                "extents"                                                                                                                                                                                          
            ],                                                                                                                                                                                                     
            "block_size": 4096,                                                                                                                                                                                    
            "blocks": 5857544,                                                                                                                                                                                     
            "charset": "binary",                                                                                                                                                                                   
            "checksum": "09bb9738b2903e9eefd64f69e3836d052fc292ec",                                                                                                                                                
            "ctime": 1527073198.1374464,                                        
            "dev": 64770,                                                                                                                                                                                          
            "device_type": 0,                                                                                                                                                                                      
            "executable": false,                                                               
            "exists": true,                                                                                                                                                                                        
            "gid": 1001,                                                                                                                                                                                           
            "gr_name": "admin",                                                                                                                                                                                    
            "inode": 11,                                                                                                                                                                                           
            "isblk": false,                                                                                                                                                                                                    "ischr": false,                                                                                                                                                                                        
            "isdir": false,                                                                                                                                                                                        
            "isfifo": false,                                                                                                                                                                                       
            "isgid": false,                                                                                                                                                                                        
            "islnk": false,                                                                                                                                                                                        
            "isreg": true,                                                                                                                                                                                         
            "issock": false,                                                                                                                                                                                       
            "isuid": false,                                                                                                                                                                                        
            "mimetype": "application/octet-stream",                                                                                                                                                                
            "mode": "0664",                                                       
            "mtime": 1527073187.535445,                                                                                                                                                                            
            "nlink": 1,                                                                                                                                                                                            
            "path": "/var/lib/libvirt/images/linchpin//consul-2018-05-17.qcow2",                                                                                                                                   
            "pw_name": "admin",                                                                                                                                                                                    
            "readable": true,                                                                                                                                                                                      
            "rgrp": true,                                                                                                                                                                                          
            "roth": true,                                                                                                                                                                                          
            "rusr": true,                                                                                                                                                                                          
            "size": 2999058432,                                                                                                                                                                                   
            "uid": 1001,                                                                                                                                                                                           
            "version": "18446744073194595580",                                                                                                                                                                     
            "wgrp": true,                                                                                                                                                    
            "woth": false,                                                              
            "writeable": true,                                                                                                                                                                                     
            "wusr": true,                                                                                                                                                                                          
            "xgrp": false,                                                                                                                                                                                         
            "xoth": false,                                                                                                                                                                                         
            "xusr": false                                                                                                                                                                                          
        }                                                                               
    }                              

Changing the conditional expression back to

when: lismt_remote.stat.mime_type == "application/octet-stream"'           

results in the following error:

    fatal: [localhost]: FAILED! => {"msg": "The conditional check 'lismt_remote.stat.mime_type == \"application/octet-stream\"' failed. The error was: error while evaluating conditional (lismt_remote.stat.mime_type 
== \"application/octet-stream\"): 'dict object' has no attribute 'mime_type'\n\nThe error appears to have been in '/home/joanbe/github/linchpin/linchpin/provision/roles/libvirt/tasks/copy_image_src_remote.yml': 
line 51, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: \"set image_src for uncompressed files\"\n  ^ here\n"}

linchpin/provision/roles/libvirt/tasks/copy_image_src_remote.yml
joanbe: Like I stated in pull request, this is highly dependent on our setup and may not be appropriate in all environments. 

In our environment the remote_user (not root!) is a member of the qemu group and the premissions on the local_image_path looks like this:
drwxrwx---. 2 root qemu 4096 Jun  5 20:55 /var/lib/libvirt/images/linchpin/

I really don't have a general solution to this. Perhaps the requirement should be dropped altogether?

herlo: @joanbe would just using the umask suffice here? We could just remove the mode line and see if that works? I think that would accomplish the same thing on both sides.

linchpin/provision/roles/libvirt/tasks/copy_image_src_remote.yml
herlo: What version of ansible are you using here? It might behoove us to ensure back to our minimum version (I believe ansible 2.3.1 at the moment) that this aligns. I'll add a task to write a basic test here to test on different ansible versions.

linchpin/provision/roles/libvirt/tasks/copy_image_src_remote.yml
joanbe: Oh it works without the mode line

linchpin/provision/roles/libvirt/tasks/copy_image_src_remote.yml
joanbe: Running ansible inside the virtualenv gives:
$ ansible --version
ansible 2.5.2
  config file = /home/joanbe/.virtualenvs/linchpin-src/ansible.cfg
  configured module search path = [u'/home/joanbe/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /home/joanbe/.virtualenvs/linchpin-src/lib/python2.7/site-packages/ansible
  executable location = /home/joanbe/.virtualenvs/linchpin-src/bin/ansible
  python version = 2.7.15 (default, May 16 2018, 17:50:09) [GCC 8.1.1 20180502 (Red Hat 8.1.1-1)]

herlo: Having mimetypes like this should work for the moment since it passes all of the tests.

linchpin/provision/roles/libvirt/tasks/copy_image_src_remote.yml
herlo: Can you make that adjustment and let's see how the tests go. Thanks!

joanbe: Will do. Commit inbound.

herlo:
[test]

herlo:
[test] TEST_TARGETS was still lying around the jenkins tests, removed it.

herlo:
[merge]

PR #588, opened 2018-05-17
bugfix: fix distiller render for aws instances

samvarankashyap:
Fixes AWS instance render to resources file.


samvarankashyap:
@robnester-rh : should fix the error reported on IRC

herlo:
[merge]

PR #589, opened 2018-05-18
Fix parsing regression for layouts and topologies

herlo:
It seems that in 1.5.3, the pf_data somehow changed back from self.pf_data. This adjusts that back so pf_data (template data) will be propagated and processed through to both the topology and inventory_layout.


herlo:
[merge]

PR #592, opened 2018-05-21
Add support for domain specification in dnsmasq

### LINKED ISSUE #543: Enhance libvirt topology to provide a domain name, for internal name resolution
Linbvirt is already running it's own dnsmasq, and with some basic changes to the network xml, it can be used for DNS inside the libvirt network. 

I'd like to be able to use that feature, but fot that I need to provide a domain name in the libvirt_network resource definition and have it generate an extra line as follows: 
```xml
<domain name='qxyz.de' localOnly='yes'/>
```
More detail at [1]

[1] https://liquidat.wordpress.com/2017/03/03/howto-automated-dns-resolution-for-kvmlibvirt-guests-with-a-local-domain/

p3ck:
There are some caveats to this.

- Currently additional networks only works if cloud-config is not used. (This is not new but it was not clear to me, so it might not be clear to others)
- You will need to use virt-customize on the image to set the hostname.  The default hostname of localhost will not update dns records.

A future RFE would be to drop the requirement on virt-install and try to integrate virt-customize.

fixes #543

### DIFF
--- a/docs/source/examples/workspace/topologies/libvirt-el7net.yml
+++ b/docs/source/examples/workspace/topologies/libvirt-el7net.yml
@@ -10,6 +10,7 @@ resource_groups:
         ip: 192.168.74.100
         dhcp_start: 192.168.74.101
         dhcp_end: 192.168.74.112
+        domain: linchpin.net
       - name: centos74
         role: libvirt_node
         uri: qemu:///system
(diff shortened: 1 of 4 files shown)


herlo:
[test]

p3ck:
@samvarankashyap Is there formal documentation for the libvirt_network schema?  All I found is a link to the example xml which I have updated in this PR.

p3ck:
[test]

p3ck:
[merge]

PR #601, opened 2018-06-04
Replaces Inventory generation role with python code

### LINKED ISSUE #448: Feature Request: variables to grab latest generated runId and inventory path
# Overview
This is a feature request to make it easy to grab the latest generated inventory and look up its info in the runDB. It's great to have all of the inventories generated be unique (to prevent thrashing), but currently, you'd have to parse either the run output or the linchpin journal history to be able to quickly grab the last generated inventory file. By giving us environment variables for these, we can now use the environment variables to perform commands on the generated inventory (e.g. running playbooks) and quickly perform a teardown using the runId environment variable.

### LINKED ISSUE #599: Drop output_writer ansible role
Now that gather_outputs and RunDB are working well, it is time to remove the output_writer role. This role is now obsolete.

samvarankashyap:
Summary: 

- Adds linchpin.latest to resources folder which stores the details of linchpin's latest run
- Deletion of all uneccessary folders, files, and filters
- Support of multiple inventory format generation.
- Adds support for inventory generations API function calls
- CLI implementation of _write_to_inventory_file 
- Linchpin shell implementation of generate command 
```
linchpin generate --inventory -f json -o /tmp/test.txt 
```
- Additional parameter to linchpin up to specify the inventory format 
```
linchpin up -if/--inventory-format json
```
fixes #599
fixes #448  
*** Will be updated with more details ***


linchpin/cli/__init__.py
```diff
@@ -119,6 +119,43 @@ def lp_init(self, providers=['libvirt']):
...
+                            the_file.write(inventory)
+                        inv_file_count += 1
+                    else:
+                        with open(inventory_path, 'w') as the_file:
+                            the_file.write(inventory)
+        except Exception as e:
+            self.ctx.log_state('Error: {0}'.format(e))
+            sys.exit(1)
+        return True
+
```
herlo: Will this function eventually write to STDOUT to take advantage of the dynamic inventory option that ansible can use?

linchpin/cli/__init__.py
```diff
@@ -262,6 +299,17 @@ def _write_distilled_context(self, run_data):
...
+        latest_run_data = self._get_run_data_by_txid()
+        resources_path = self.get_evar('resources_folder')
+        context_path = '{0}/{1}'.format(self.workspace, resources_path)
+        if not os.path.exists(context_path):
+            os.makedirs(context_path)
+        context_file = '{0}/{1}'.format(context_path, 'linchpin.latest')
+        with open(context_file, 'w+') as f:
+            f.write(json.dumps(latest_run_data))
+
```
herlo: This is great! I'm excited to see this out there. Additionally, it might be advantageous to turn this into the outputs (to STDERR) at the end of a linchpin run. The [_handle_results function](https://github.com/CentOS-PaaS-SIG/linchpin/blob/develop/linchpin/shell/__init__.py#L24) could be replaced.

linchpin/linchpin.constants
```diff
@@ -11,10 +11,12 @@ rundb_conn_type = file
 rundb_schema = {"action": "",
                 "inputs": [],
                 "outputs": [],
+                "topology_outputs": [],
```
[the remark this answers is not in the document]
samvarankashyap: @herlo 
Though the outputs and topology outputs seems to be same , There is a subtle difference between both:
```
(Pdb) import pprint
(Pdb) pprint.pprint(targets[name]["topology_outputs"])
{u'dummy_res': [{u'changed': False,
                 u'dummy_file': u'/tmp/dummy.hosts',
                 u'failed': False,
                 u'hosts': [u'web-fde56f-0.example.net']},
                {u'changed': False,
                 u'dummy_file': u'/tmp/dummy.hosts',
                 u'failed': False,
                 u'hosts': [u'web2-fde56f-0.example.net',
                            u'web2-fde56f-1.example.net']}]}
(Pdb) pprint.pprint(targets[name]["outputs"])
{u'resources': [{u'changed': False,
                 u'dummy_file': u'/tmp/dummy.hosts',
                 u'failed': False,
                 u'hosts': [u'web-fde56f-0.example.net']},
                {u'changed': False,
                 u'dummy_file': u'/tmp/dummy.hosts',
                 u'failed': False,
                 u'hosts': [u'web2-fde56f-0.example.net',
                            u'web2-fde56f-1.example.net']}]}
```
As you can see , topology_outputs segregates the resources based on the type of the resource that is being generated Which made the existing inventory generation code reusable. So in the above case if there is aws_ec2 resource provisioned inside the topology  along with dummy topology their respective topology_outputs and outputs looks as follows:
```
(Pdb) import pprint
(Pdb) pprint.pprint(targets[name]["topology_outputs"])
{u'dummy_res': [{u'changed': False,
                 u'dummy_file': u'/tmp/dummy.hosts',
                 u'failed': False,
                 u'hosts': [u'web-fde56f-0.example.net']},
                {u'changed': False,
                 u'dummy_file': u'/tmp/dummy.hosts',
                 u'failed': False,
                 u'hosts': [u'web2-fde56f-0.example.net',
                            u'web2-fde56f-1.example.net']}],
u'aws_res': [{u'changed': true,
                    'xxx': xxx
                   }] # just for idea 

}
(Pdb) pprint.pprint(targets[name]["outputs"])
{u'resources': [{u'changed': False,
                 u'dummy_file': u'/tmp/dummy.hosts',
                 u'failed': False,
                 u'hosts': [u'web-fde56f-0.example.net']},
                {u'changed': False,
                 u'dummy_file': u'/tmp/dummy.hosts',
                 u'failed': False,
                 u'hosts': [u'web2-fde56f-0.example.net',
                            u'web2-fde56f-1.example.net']},
                {u'changed': False,
                 u'hosts': [u'ec2ip.x.x.x'] }                          # just for the idea
                  ]}
```
 However, topology_outputs consolidates everything into a single key resources regardless of their type. topology_outputs will be helpful for the future use cases when users would like to have only a specific cloud provider type of inventory. 
You can say on can identify the resource from "outputs" by writing up logic to identify schema of each output resource which seems a difficult tasks and hard to maintain since the output formats changes can break the code anytime. 
There might be a better way to do this . But adding a key felt like a easier way.

linchpin/linchpin.constants
```diff
@@ -11,10 +11,12 @@ rundb_conn_type = file
...
                 "cfgs": [],
                 "start": "",
                 "end": "",
                 "rc": 0,
+                "inventory_path": "",
```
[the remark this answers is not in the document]
samvarankashyap: @herlo : 
inventory_path is generated with respect to per target run . I felt that should be at a higher level since it is independent (and can change frequently for future case where user would like to rerun and update records).

linchpin/linchpin.constants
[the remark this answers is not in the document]
herlo: @samvarankashyap Actually, I'm not saying that it isn't a separate item at all. I was just saying that since it's of 'output' type, it should be stored in the 'output' key. There can be *many* items in the output area, each of them can be different. I realize that resources and inventory (or topology outputs as you call them) are different. In fact, in another comment on this PR, I suggested a solution to separate out the 'topology' outputs from the 'resource' outputs. However, they are both outputs, just different kinds. Does this make better sense?

linchpin/shell/__init__.py
```diff
@@ -323,6 +328,34 @@ def fetch(ctx, fetch_type, remote, root):
...
+              default=None, type=click.Path(),
+              help='Output file path to be written \
+                    if not mentioned will be redirected to stdout')
+@click.option('-f', '--format', metavar='format', default="cfg",
+              help='Inventory output format')
+@click.option('-t', '--tx-id', metavar='tx_id', type=int, default=None,
+              help='Transaction ID to be used to generate inventory')
+@click.option('-inventory', '--inventory', metavar='format',
+              is_flag=True, required=False,
+              help='Type of resource to be generated supports inventory only')
```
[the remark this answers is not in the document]
samvarankashyap: Agreed ! Output file is much cleaner way to go.

linchpin/cli/__init__.py
samvarankashyap: we could always make it do that.

herlo: And I guess in the STDOUT case, it might be better to put that with the [_handle_results function](https://github.com/CentOS-PaaS-SIG/linchpin/blob/develop/linchpin/shell/__init__.py#L24) as we discussed below.

herlo:
@samvarankashyap be sure to rebase against develop when you have a moment so the tests will run properly. Thanks!

p3ck:
this is currently failing ci with the following error:
14:37:00 [fedora27] Action 'up' on Target 'libvirt-new' is complete
14:37:00 [fedora27] Error: option keys must be strings

I'm not quite sure where that is coming from.

samvarankashyap:
[test]

p3ck:
18:17:31 [fedora26] Action 'up' on Target 'aws-ec2-new' is complete
18:17:31 [fedora26] Error: 'layout_data'

linchpin/cli/__init__.py
```diff
@@ -119,6 +119,54 @@ def lp_init(self, providers=['libvirt']):
...
+        all_inventories = []
+        try:
+            for t_id in latest_run_data:
+                targets = latest_run_data[t_id]["targets"][0]
+                # if there are multiple targets mentioned in pinfile
+                # the multiple inventory files are being generated
+                inv_file_count = 0 if len(targets) > 1 else False
+                for name in targets:
+                    try:
+                        lt_data = targets[name]["inputs"]["layout_data"]
```
herlo: This function should verify that layout_data exists before running this function. It seems this is the cause of that failed test previously. Maybe it's already been fixed, however...

linchpin/shell/__init__.py
```diff
@@ -323,6 +328,38 @@ def fetch(ctx, fetch_type, remote, root):
...
+            click.echo(inventory)
+    else:
+        try:
+            click.echo(ret_bool[int(pos)])
+        except IndexError as e:
+            click.echo("Invalid index of inventory generated")
+            click.echo(e.message)
+    return ret_bool
+
+
```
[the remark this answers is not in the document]
samvarankashyap: @herlo : Currently linchpin's rundb does not store inventory. Further , I have designed linchpin generate such that user can have a option to generate provider specific inventory by mentioning --provider-type in future . Currently linchpin generate has multiple options of specifying --inventory-format , --inventory-type and --outputfile Adding these three options to linchpin journal in addition to the ones which already have felt overwhelming it with many options.
^^^ Any thoughts?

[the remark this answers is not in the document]
herlo: @samvarankashyap what does 'generate' mean here? It feels like it 'generates' something, but what? Adding an new subcommand means documenting it and detailing what is involved with the 'generate'.  If we were to do this, I'd argue that at a minimum, the subcommand should be named appropriately. Something like 'inventory' makes more sense in that case. Additionally, If the inventory were stored, other parameters would not be possible, so I see why you chose to make it a separate subcommand. 

On the other hand, 'journal' provides a 'a transactional record' for each task, so we're just taking the data from the record and displaying it in different ways. In this way, the journal is useful to a user.  As you stated in your comment above, the inventory doesn't need to be stored in the rundb, only the topology and layout. Using the topology, layout, and possibly other dynamic data, one could also create an inventory using the 'journal' subcommand. However, one could also obtain the linchpin.latest, or create a shell file that could be sourced to create environment variables. The possibilities are fairly unlimited. 

The vision I had for this was something akin to a single subcommand that could use either a tx_id or run_id/target combination, pass in a few options, and get an output to either STDOUT or a file with exactly the data a user would want/need. Something like the following:

```
$ linchpin journal --tx-id 82 --template-data @./domain-data.yml \
   --output-type=json-inventory
```
The above command would print json to STDOUT. Alternatively, one might want a linchpin.latest with shell environment values inside.

```
$ linchpin journal --target=aws-ec2-cluster --output-file /tmp/tmpfile.2.env \
   --output-type env-latest

```
The list is not limited here.

```
$ linchpin journal -tx-id 84 --target=aws-ec2 --output-type json-distilled \
   -o /tmp/tmpfile.3.distilled --fields='aws_ec2=instances.id,instances.tags:name'
```

What you can see here is that while at runtime, the `linchpin up` command should be able to output these sorts of items, and `linchpin journal` should be able to replicate them. Naming the subcommand `generate` or `journal` doesn't really matter per se. Adding a new subcommand seems like overkill to me, since journal already does the heavy lifting to look up the record data.

The only real adjustment needed would be to take what the generate subcommand does now and adapt it to the above. 

If you have a better name for 'journal', I'd be open to leaving what is there and adjusting to that name. Documentation aside, I believe this functionality is all related to looking up a record, and returning some sort of output a user (or application) can use.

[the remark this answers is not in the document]
herlo: @samvarankashyap I did have one other thought for you. I was looking at the difference in outputs in rundb. Between 'resources' and 'topology_outputs', the only real difference I can see is that topology_outputs adds the structure around it for each different resource definition. In my mind, that's a lot of possible duplicate data. 

Would it be advantageous to change the 'resources' component to have the same structure as topology outputs? It might mean changing the lookups a little in the API, but then we wouldn't need to duplicate data in the rundb.

A snippet to help point the issue might be useful here:

```
"resources": [
    {
        "changed": true,
        "failed": false,
        "instance_ids": [
            "i-0881adb4d84c4e470",
            "i-0de04151d162add7b",
            "i-0e320bf9ab05092bd"
        ],
        "instances": [
            {
                "ami_launch_index": "2",
                "architecture": "x86_64",
                "block_device_mapping": {
                    "/dev/sda1": {
                        "delete_on_termination": true,
                        "status": "attached",
                        "volume_id": "vol-032b73e03f8458215"
                    },
                    "/dev/sdb": {
                        "delete_on_termination": true,
                        "status": "attached",
                        "volume_id": "vol-0ca7d8e99e5c33976"
```
vs
```
"topology_outputs": {
   "aws_cfn_res": [],
   "aws_ec2_key_res": [],
   "aws_ec2_res": [
       {
           "changed": true,
           "failed": false,
           "instance_ids": [
               "i-0881adb4d84c4e470",
               "i-0de04151d162add7b",
               "i-0e320bf9ab05092bd"
           ],
           "instances": [
               {
                   "ami_launch_index": "2",
                   "architecture": "x86_64",
                   "block_device_mapping": {
                       "/dev/sda1": {
                           "delete_on_termination": true,
                           "status": "attached",
                           "volume_id": "vol-032b73e03f8458215"
```
I'm not sure why it matters here as it seems maybe you could use the 'resources' section for this, and I know we discussed this previously. However, duplication of data seems a waste for space. Why not consolidate this into one structure? The reason the resources is a list is so we can include multiple different resource topologies in one target, but I'm not opposed to having the data in the structure you use for topology_outputs (just resources reformatted).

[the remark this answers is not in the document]
samvarankashyap: @herlo : @p3ck 
After a code walk through I realized following: 
resources --> format is being used by distiller and linchpin.latest file generation ( which I see referenced in couple of independent function calls). 
topology_outputs --> format is being is used by inventory generation. 
consolidating both  of the formats involves re-writing logic for ( distiller, linchpin.latest ) or inventory generation. In my opinion it would be definitely be advantageous to switch to topology_outputs structure since it has more data w.r.to providers. 

In order to keep the code changes minimal . I had to sought for duplication of the data. 
If we would like to go ahead and following topology_outputs format ( which I personally think is easier to implement) . I might take a couple of days of effort. To make sure its not breaking any of distiller/linchpin.latest file. Would we mind if the PR is delayed with the requested changes ?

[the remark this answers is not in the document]
herlo: The only qualm I have is that we keep the name 'resources', as topology_outputs is a long way of saying the same thing. Beyond that, I don't know which is better. Merging this PR, then making changes to another one is also fine as this one works. @p3ck do you have any suggestions on this? It does feel like the 'generate' -> 'journal' functionality would need to be in this PR. But since we just started a new sprint and aren't releasing anytime soon, I'm could go either way with it as long as the functionality is adjusted.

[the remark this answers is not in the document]
p3ck: I think a little delay is acceptable.  It sounds like the right approach to me.

samvarankashyap:
Dropped linchpin generate and integrated inventory generation into linchpin journal as follows:
```
linchpin journal --view inventory --output-format json --output-file /tmp/testjournal.txt --tx-id 130
```
Further , consolidated topology_outputs and resources inside rundb to have a uniform data entry

samvarankashyap:
@herlo @p3ck : Any idea why the build is failing ?

p3ck:
I think it's here:
`14:04:14 [fedora27] TASK [duffy : Teardown Duffy Resources] ****************************************
14:04:14 [fedora27] fatal: [localhost]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: {{ topo_output_resources[0]['ssid'] }}: dict object has no element 0\n\nThe error appears to have been in '/usr/lib/python2.7/site-packages/linchpin/provision/roles/duffy/tasks/teardown_res_defs.yml': line 2, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n---\n- name: \"Teardown Duffy Resources\"\n  ^ here\n\nexception type: <class 'ansible.errors.AnsibleUndefinedVariable'>\nexception: {{ topo_output_resources[0]['ssid'] }}: dict object has no element 0"}`

samvarankashyap:
[test]

samvarankashyap:
[test]

samvarankashyap:
[test]

linchpin/provision/rundb.yml
```diff
@@ -19,7 +19,7 @@
         conn_str: "{{ rundb_path }}"
         operation: init
         table: "{{ rundb_table }}"
-        value: {'inputs': [], 'outputs': [], 'start': '', 'end': '', 'rc': 0, 'action': ''}
+        value: {'inputs': [], 'outputs': [], 'topology_ouputs': [], 'start': '', 'end': '', 'rc': 0, 'action': '', 'inventory_path': ''}
```
herlo: topology_outputs is spelled incorrectly here. Plus, shouldn't this be `resources`? If we're going to be backward compatible, we should at least use the same reference point in the code and data store.

linchpin/cli/__init__.py
```diff
@@ -119,6 +119,54 @@ def lp_init(self, providers=['libvirt']):
...
+                        layout = lt_data["inventory_layout"]
+                        i_path = targets[name]["outputs"]["inventory_path"][0]
+                        if inv_path and inv_file_count is not False:
+                            i_path = inv_path + str(inv_file_count)
+                        # t_o -> topology_outputs
+                        t_o = targets[name]["outputs"]["resources"]
```
herlo: You specify resources here, but other places, topology_outputs. I think there may be a few places that consolidation is necessary still. In my mind, resources > topology_outputs because backward compatibility and length.

linchpin/provision/dummy.yml
```diff
@@ -3,27 +3,18 @@
...
   hosts: localhost
-  gather_facts: True
+  gather_facts: False
   roles:
     - { role: 'dummy' }
 
 - name: "Writing contents to file"
   hosts: localhost
-  gather_facts: True
+  gather_facts: False
```
herlo: Be careful about setting `gather_facts: False`. I've run into more than one issue where having gather_facts whas helpful, especially during the gather_resources role as it actually uses some of the details for libvirt (possibly others).

linchpin/provision/roles/aws/tasks/teardown_aws_ec2.yml
```diff
@@ -1,4 +1,8 @@
 ---
+- name: "Debug"
+  debug:
+    msg: "{{ tp_out }}"
+
```
herlo: This needs to be commented out at a minimum because it will print account information.

linchpin/provision/roles/gather_resources/tasks/main.yml
```diff
@@ -59,3 +57,26 @@
...
+  set_fact:
+    topology_name: "{{ topo_data['topology_name'] }}"
+
+- name: "Updating uhash-based inventory_path with the absolute path"
+  set_fact:
+    inventory_path: "{{ inventory_path | default( workspace +'/'+inventories_folder+'/'+
+    topology_name.replace(' ', '_').lower() ) + '-' + uhash }}.inventory"
+  when: enable_uhash
+- name: "Updating inventory_path with the absolute path"
+  set_fact:
```
herlo: Nit: formatting (add a new line)

linchpin/shell/__init__.py
```diff
@@ -466,10 +482,22 @@ def journal(ctx, targets, fields, count, view, tx_id):
                 output += '\n==================NO TRANSACTIONS======'
                 output += '==========\n'
 
-
-        # PRINT OUTPUT RESULTS HERE
         ctx.log_state(output)
 
+    if view == "inventory":
```
herlo: As I stated above, I don't think 'inventory' is a view type, but an output type. We look at transactions, (or target/run_id combination) as views. Maybe allow output-format or some such to handle the details. Here's some code that may describe it better: 

```
if output_format == 'json-inventory':
    .. snip ..
elif output_format == 'yaml-inventory':
    .. snip ..
elif output_format == 'cfg-inventory':
   .. snip ..
```
And add other options here. Clearly this is probably not the best way to use the inventory format, but I think it gets the point across.

linchpin/shell/__init__.py
```diff
@@ -425,6 +440,7 @@ def journal(ctx, targets, fields, count, view, tx_id):
             print('No targets available for journal.'
                   ' Please provision something. :)', file=sys.stderr)
         print('\n')
+        ctx.log_state(output)
```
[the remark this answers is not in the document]
samvarankashyap: @herlo : Did not understand what you are trying to convey . Shouldnt be sent to stdout ?  you mean.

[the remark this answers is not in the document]
herlo: After reviewing I think I may have been wrong that it prints to STDOUT. It looks like the `log_state()` method logs to STDERR. 

What I was implying, however, is that we want all output, EXCEPT the generated inventory, printed to STDERR. The reason for doing it this way is to allow for dynamic inventories to work when using linchpin in non-verbose mode in the future. 

I hope that clears it up, but you can safely ignore this comment, unless you find that it does indeed print to STDOUT.

samvarankashyap:
@herlo : requested changes made !

linchpin/shell/__init__.py
```diff
@@ -376,6 +380,22 @@ def journal(ctx, targets, fields, count, view,
     (Default: action, uhash, rc. Additional fields: start, end)
     """
 
+    if output_type == "inventory":
+        inventories = lpcli._write_to_inventory(inv_path=output_file,
+                                             inv_format=output_format)
+        if target == "all":
+            click.echo("By default all targets inventories are displayed to stdout\
+                        For specific target please use --target option")
```
herlo: Probably ought to put this warning below all of the targets to STDERR.

linchpin/shell/__init__.py
```diff
@@ -340,19 +340,23 @@ def fetch(ctx, fetch_type, remote, root):
               metavar='TX_ID', required=False,
               help="Display a specific transaction by ID (tx_id)."
                    " Only works with '--view=tx'")
-@click.option('-of', '--output-format', type=str, default="cfg",
+@click.option('--output-format', type=str, default="cfg",
               metavar='output_format', required=False,
               help="Inventory output format")
-@click.option('-outputfile', '--output-file', type=str,
+@click.option('--output-file', type=str,
               metavar='output_file', required=False,
```
herlo: I'm still wondering why this option can't be a general option as I pointed out previously?

samvarankashyap: @herlo : Output-file is a optional parameter by default the inventory_path from the rundb is considered if not mentioned . It is used when users would like to write inventory file to external path.

herlo: @samvarankashyap I agree and understand the purpose of the `--output-file` option. I posted this previously [here](https://github.com/CentOS-PaaS-SIG/linchpin/pull/601#discussion_r195868811).

What I am suggesting is that `--output-pinfile` and `--output-file` serve the same purpose for different verbs (subcommands). In this case, `linchpin up` and `linchpin journal`.

Using `--output-file` in one place reduces duplication. See this example:

```
linchpin -vp PinFile.libvirt-mi.template --template-data @Data.libvirt-mi.yml \
--output-pinfile /tmp/PinFile.libvirt-mi.generated up
```
Adding the `--output-file` option to the above command would be an equivalent of `--output-pinfile`:

```
linchpin -vp PinFile.libvirt-mi.template --template-data @Data.libvirt-mi.yml \
--output-file /tmp/PinFile.libvirt-mi.generated up
```
Equally, this could be applied to linchpin journal in this way:

```
linchpin --output-file /tmp/tester.txt journal -t 143 --output-format json \
--output-type inventory --target oso_target
```
The location of `--output-file` is important here. It's a general option so it would come before the `journal` verb. We could do this for `--output-format` as well, since a PinFile could be json or yaml. 

Does this make more sense?

samvarankashyap:
[test]

samvarankashyap:
[test]

linchpin/shell/__init__.py
```diff
@@ -336,8 +340,23 @@ def fetch(ctx, fetch_type, remote, root):
               metavar='TX_ID', required=False,
               help="Display a specific transaction by ID (tx_id)."
                    " Only works with '--view=tx'")
+@click.option('--output-format', type=str, default="cfg",
+              metavar='output_format', required=False,
+              help="Inventory output format")
+@click.option('--output-file', type=str,
```
herlo: I've said this a couple times previously, but we should move this up into the generic options. It should include `--output-pinfile` with this option `--output-file`.

linchpin/shell/__init__.py
```diff
@@ -336,8 +340,23 @@ def fetch(ctx, fetch_type, remote, root):
               metavar='TX_ID', required=False,
               help="Display a specific transaction by ID (tx_id)."
                    " Only works with '--view=tx'")
+@click.option('--output-format', type=str, default="cfg",
+              metavar='output_format', required=False,
+              help="Inventory output format")
```
herlo: What's the difference between this and `--inventory-format` from above?

herlo: What is the difference between this and `--inventory-format` in the `up()` method? Seems these options should be combined and put into a general option. I'd go with `--output-format`. 

The one challenge with this is that there is already a `--output-pinfile` general option, and the standard output for that is 'json'. It seems to me that this would be a good time to move the default of `--output-format` to 'json'. While this might not be backward compatible in terms of format, it should be in terms of functionality.

linchpin/shell/__init__.py
```diff
@@ -336,8 +340,23 @@ def fetch(ctx, fetch_type, remote, root):
...
+@click.option('--output-file', type=str,
+              metavar='output_file', required=False,
+              help="inventory output file path")
```
herlo: I have pointed this out a couple of times, but `--output-file` should be combined with the general option `--output-pinfile` in the general options. You can combine these two options to have something like:

```
@click.option('-o', '--output-file', '--output-pinfile', metavar='OUTPUT_PINFILE',
              help='Write out PinFile to provided location')
```
This way, we continue to use DRY pattern. This goes for combining `--output-format` and `--inventory-format` as well, which I pointed out elsewhere in this review.

linchpin/shell/__init__.py
```diff
@@ -336,8 +340,23 @@ def fetch(ctx, fetch_type, remote, root):
...
+@click.option('--output-type', type=str,
+              metavar='output_type', required=False,
+              help="inventory output file path")
```
herlo: If we genericize this option, we could have `--output-type` be 'inventory', 'pinfile', 'distilled', 'latest', etc. If no `--output-type` is passed, each method should determine the default.

linchpin/shell/__init__.py
samvarankashyap: @herlo I tried combining as you mentioned however , click is not allowing multiple aliases to same option.
```(env_vars) [srallaba@laptopy workspace]$ linchpin journal -t 153 --output-type inventory --output-file /tmp/testinv.txt --output-format json
Traceback (most recent call last):
  File "/home/srallaba/.virtualenvs/env_vars/bin/linchpin", line 11, in <module>
    load_entry_point('linchpin', 'console_scripts', 'linchpin')()
  File "/home/srallaba/.virtualenvs/env_vars/lib/python2.7/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/home/srallaba/.virtualenvs/env_vars/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/home/srallaba/.virtualenvs/env_vars/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/srallaba/.virtualenvs/env_vars/lib/python2.7/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/srallaba/.virtualenvs/env_vars/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/home/srallaba/.virtualenvs/env_vars/lib/python2.7/site-packages/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args[1:], **kwargs)
  File "/home/srallaba/.virtualenvs/env_vars/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
TypeError: journal() got an unexpected keyword argument 'output_pinfile'
```
The moment we add --output-pinfile as option as follows the above error trace occurs . 
```python
@click.option('-o', '--output-file', '--output-pinfile', type=str,
              metavar='OUTPUT_FILE', required=False,
              help="Write output-file to provided location")
@click.option('--output-type', type=str,
              metavar='output_type', required=False,
              help="inventory output file path")
@click.option('--target', metavar='target', default="all",
              help='If multiple targets are mentioned \
                    takes parameter for target to be used.\
                    By default all are displayed\
                    displayed')
@pass_context
def journal(ctx, targets, fields, count, view,
            tx_id, output_format, output_file, output_type, target):
```

linchpin/shell/__init__.py
samvarankashyap: @herlo : ^^^ Totally agree with the direction .

linchpin/shell/__init__.py
herlo: @samvarankashyap I was able to make this work without any other modifications. I'm confused as to why you are getting that error on your end. Here's the changes I made.

```
$ git diff linchpin/shell/__init__.py
diff --git a/linchpin/shell/__init__.py b/linchpin/shell/__init__.py
index 83a6e8a..6b67047 100644
--- a/linchpin/shell/__init__.py
+++ b/linchpin/shell/__init__.py
@@ -114,7 +114,8 @@ def _handle_results(ctx, results, return_code):
                    "If template data is from a file, it must be"
                    " prepended with an '@' character."
               )
-@click.option('-o', '--output-pinfile', metavar='OUTPUT_PINFILE',
+@click.option('-o', '--output-file', '--output-pinfile', 'outfile',
+              metavar='OUTPUT_FILE',
               help='Write out PinFile to provided location')
 @click.option('-w', '--workspace', type=click.Path(), envvar='WORKSPACE',
               help='Use the specified workspace. Also works if the'
@@ -127,7 +128,7 @@ def _handle_results(ctx, results, return_code):
               help='Use the specified credentials path. Also works'
                    ' if CREDS_PATH environment variable is set')
 @pass_context
-def runcli(ctx, config, pinfile, template_data, output_pinfile,
+def runcli(ctx, config, pinfile, template_data, outfile,
            workspace, verbose, version, creds_path):
     """linchpin: hybrid cloud orchestration"""
 
@@ -141,8 +142,8 @@ def runcli(ctx, config, pinfile, template_data, output_pinfile,
     # if the pinfile is a template, data will be passed here
     ctx.pf_data = template_data
 
-    if output_pinfile:
-        ctx.set_cfg('tmp', 'output_pinfile', output_pinfile)
+    if outfile:
+        ctx.set_cfg('tmp', 'output_pinfile', outfile)
 
     ctx.pinfile = None
```
When I run `linchpin journal`, or `linchpin up`, I get the expected output. Even the PinFile in JSON format is output when using the `linchpin up` command.

```
$ linchpin -o x journal dummy-new

Target: dummy-new
run_id	    action	     uhash	        rc	
--------------------------------------------------
553    	       up	   1278c4	        0	
552    	       up	   475870	        0	
551    	       up	   f57b00	        0	

$ linchpin --output-file x journal dummy-new

Target: dummy-new
run_id	    action	     uhash	        rc	
--------------------------------------------------
553    	       up	   1278c4	        0	
552    	       up	   475870	        0	
551    	       up	   f57b00	        0	

$ linchpin --output-pinfile x journal dummy-new
 
Target: dummy-new
run_id	    action	     uhash	        rc	
--------------------------------------------------
553    	       up	   1278c4	        0	
552    	       up	   475870	        0	
551    	       up	   f57b00	        0	

$ linchpin -o x up dummy-newAction 'up' on Target 'dummy-new' is complete
-------
start hook shell:hello
hello

end hook shell:hello
-------

ID: 963
Action: up

dummy-new           	   554	39bd76	        0

```
The output files don't work yet on the journal bit (as I didn't implement that logic to test this functionality), but it should not be difficult to make work.

samvarankashyap: @herlo : I see where i am going wrong . I was trying to change journal cli instead of runcli command. Fixing it right now .

samvarankashyap:
@herlo I see tests on this PR are successful however , its reported failure to GitHub . I wonder why ? Further , I don't see any Failures on the console.log too .

herlo:
@samvarankashyap it was not the Jenkins build that failed, but the travis-ci build. See https://travis-ci.org/CentOS-PaaS-SIG/linchpin/builds/396459356?utm_source=github_status&utm_medium=notification

herlo:
@samvarankashyap I believe this patch will fix the test that is causing the problem.

```
diff --git a/config/Dockerfiles/tests.d/combined/01_beaker_openstack b/config/Dockerfiles/tests.d/combined/01_beaker_openstack
index b6cff43..52b9a83 100755
--- a/config/Dockerfiles/tests.d/combined/01_beaker_openstack
+++ b/config/Dockerfiles/tests.d/combined/01_beaker_openstack
@@ -21,7 +21,7 @@ echo "${DESCRIPTION}"
 
 function clean_up {
     set +e
-    linchpin -w . -p "${PINFILE}" -v destroy "${TARGET}"
+    linchpin -w . -p "${PINFILE}" --creds-path=../../../../keys/openstack -v destroy "${TARGET}"
     rm ${TMP_FILE}
 }
 trap clean_up EXIT SIGHUP SIGINT SIGTERM
```

If you apply that, the tests should pass. Then I can take a look at the UI and hopefully approve this PR. :+1:

herlo:
I deleted the previous comment because it was missing a line. Here's the updated diff

```
diff --git a/config/Dockerfiles/tests.d/combined/01_beaker_openstack b/config/Dockerfiles/tests.d/combined/01_beaker_openstack
index b6cff43..52b9a83 100755
--- a/config/Dockerfiles/tests.d/combined/01_beaker_openstack
+++ b/config/Dockerfiles/tests.d/combined/01_beaker_openstack
@@ -21,7 +21,7 @@ echo "${DESCRIPTION}"
 
 function clean_up {
     set +e
-    linchpin -w . -p "${PINFILE}" -v destroy "${TARGET}"
+    linchpin -w . -p "${PINFILE}" --creds-path=../../../../keys/openstack -v destroy "${TARGET}"
     rm ${TMP_FILE}
 }
 trap clean_up EXIT SIGHUP SIGINT SIGTERM
```

If you apply that, the tests should pass. Afterward, we can do a full review of the features and hopefully approve. :+1

herlo:
Looks like there's a fundamental change that needs to happen in credential handling that isn't going to work  here. Let's just disable that test for now.

```
diff --git a/config/Dockerfiles/tests.d/combined/01_beaker_openstack b/config/Dockerfiles/tests.d/combined/01_beaker_openstack
index b6cff43..2e5fd6d 100755
--- a/config/Dockerfiles/tests.d/combined/01_beaker_openstack
+++ b/config/Dockerfiles/tests.d/combined/01_beaker_openstack
@@ -1,8 +1,8 @@
 #!/bin/bash -xe
 
 # Provision both beaker and openstack with one PinFile
-# distros.exclude: none
-# providers.include: beaker
+# distros.exclude: fedora27 fedora26 centos7
+# providers.include: none
 # providers.exclude: none
 #
 # NOTE: The providers.include only sets beaker in an effort to run only the
```
I'm going to fixup the testing framework in the next sprint, probably anyway.

config/Dockerfiles/tests.d/combined/01_beaker_openstack
```diff
@@ -1,8 +1,8 @@
 #!/bin/bash -xe
 
 # Provision both beaker and openstack with one PinFile
-# distros.exclude: none
-# providers.include: beaker
+# distros.exclude: fedora27 fedora26 centos7
+# providers.include: none
```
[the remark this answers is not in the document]
herlo: @p3ck it will. As discussed in IRC, the test has a broken condition where the openstack credentials are not yet extracted if the beaker tests are run first. That needs to be resolved. I've added #620 to address the issue.

herlo:
[test]

herlo:
A couple things I noticed when looking through the last CI run for this ticket.

Why does the following output show up after a destroy? There is not a need for a layout in a destroy action. `Error: 'layout_data' :Not found` 

Second, I saw that the os-sg-new target failed to destroy (directly below this message). If the security group does not exist, destroy ignores this and reports success, since the state is idempotent. Additionally, even though we run a destroy for security groups, I'm pretty sure this did nothing as we didn't delete any security groups.

https://jenkins-continuous-infra.cloud.continuous-infra.upshift.redhat.com/job/ci-linchpin/26/artifact/linchpin/fedora27_logs/openstack.log

The libvirt error on the upstream jenkins is a known race condition. It may work on the next test run(s).

herlo:
@samvarankashyap I did notice that the race condition for libvirt testing caused failures again as well. Probably can just rerun the tests after fixing the above issue. However, I did notice outputs that look like this throughout the code when something went wrong. 

```
Error: 'outputs' :Not found
```
This one in particular is in https://jenkins-continuous-infra.apps.ci.centos.org/job/ci-linchpin/208/artifact/linchpin/centos7_logs/libvirt.log. I wonder if those are necessary?

samvarankashyap:
@herlo 
I can't find 
```
Error: 'layout_data' :Not found
```
in destroy operation. All the three occurrences are found  up operations
https://jenkins-continuous-infra.cloud.continuous-infra.upshift.redhat.com/job/ci-linchpin/26/artifact/linchpin/fedora27_logs/openstack.log

Further, 
The error is logged due to following code :  
https://github.com/samvarankashyap/linchpin/blob/f33679e5666f5b53d5f742b884bdfd1f3bf8d52a/linchpin/cli/__init__.py#L160
Since there can be multiple targets its looping through all targets which might have the layout. And logs out as Error if layout not found . 
However, I feel that it should be Ignoring layout Instead of error.

herlo:
@samvarankashyap currently, the jenkins on Centos CI is having issues, so I can't paste and updated location. However, the point here is that if there is NOT a layout to utilize it should be ignored. The error I pointed out is during a destroy action, and should never print that error.

herlo:
[test]

herlo:
[test]

herlo:
The duffy resources were having challenges, along with openstack resources being exhausted. I ran the upstream tests again (without libvirt due to the network failure) and it passes.

```
+ printf '\n==== TEST summary ====\nSUCCESS\tcentos7/dummy/01_dummy-template-data\nSUCCESS\tcentos7/dummy/02_dummy-template-data-file\nSUCCESS\tcentos7/aws/01_general\nSUCCESS\tcentos7/dummy/01_general\nSUCCESS\tcentos7/aws/01_template_inventory\n'

==== TEST summary ====
SUCCESS	centos7/dummy/01_dummy-template-data
SUCCESS	centos7/dummy/02_dummy-template-data-file
SUCCESS	centos7/aws/01_general
SUCCESS	centos7/dummy/01_general
SUCCESS	centos7/aws/01_template_inventory
+ exit 0
```
Just waiting on the downstream tests to pass again and I'll approve.

herlo:
[test]

herlo:
@samvarankashyap  Please rebase this PR against develop and run test.

herlo:
[test]

samvarankashyap:
@herlo Duffy fails !

herlo:
@samvarankashyap If you actually look closer, it's not duffy, but aws that fails. In https://jenkins-continuous-infra.apps.ci.centos.org/job/ci-linchpin/251/artifact/linchpin/centos7_logs/aws.log, you can see the failure during the first test here:

```
TASK [aws : Provisioning AWS_EC2 Resource when not async] **********************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "wait for instances running timeout on Thu Jul 26 21:28:52 2018"}
	to retry, use: --limit @/usr/lib/python2.7/site-packages/linchpin/provision/aws.retry

localhost                  : ok=16   changed=1    unreachable=0    failed=1   


ID: 9
Action: up

aws-ec2-new         	     1	660010	        2
```

This looks to me like the aws account we use has a quota limit, and maybe we're hitting it? Unsure exactly.

herlo:
@samvarankashyap I was tasked with figuring out a bug today with two resource groups in a topology and a layout structure which should generate an inventory. Here's the PinFile:

```
---
libvirt-dummy:
  topology:
    topology_name: libvirt-dummy
    resource_groups:
      - resource_group_name: libvirt-new
        resource_group_type: libvirt
        resource_definitions:
          - role: libvirt_node
            name: centos71
            uri: qemu:///system
            count: 1
            memory: 2048
            vcpus: 1
            arch: x86_64
            ssh_key: libvirt
            networks:
              - name: default
            additional_storage: 10G
            cloud_config:
              users:
                - name: herlo
                  gecos: Clint Savage
                  groups: wheel
                  sudo: ALL=(ALL) NOPASSWD:ALL
                  ssh-import-id: gh:herlo
                  lock_passwd: true
            count: 1
      - resource_group_name: dummy-topology
        resource_group_type: dummy
        resource_definitions:
          - name: "{{ distro | default('') }}test"
            role: "dummy_node"
  layout:
    inventory_layout:
      vars:
        hostname: __IP__
      hosts:
        libvirt-node:
          count: 1
          host_groups:
            - dummy-layout
        dummy-node:
          count: 1
          host_groups:
            - aws-layout
      host_groups:
        all:
          vars:
            ansible_python_interpreter: /usr/bin/python3
            ansible_user: root
            ansible_ssh_common_args: -o StrictHostKeyChecking=no
```

When I run this topology through linchpin, I get provisioned resources, but only the latest resource group is parsed into the inventory file. It should allow for *all* resources to be added to the inventory until the number of resources is greater than the number requested in the inventory.

As you can see, I get the following inventory
```
$ cat /home/herlo/Projects/centos/git/linch-pin/docs/source/examples/workspace/inventories/libvirt-dummy-ebe107.inventory
[all:vars]
ansible_ssh_common_args = -o StrictHostKeyChecking=no
ansible_python_interpreter = /usr/bin/python3
ansible_user = root

[aws-layout]

[dummy-layout]
test-ebe107-0.example.net hostname=test-ebe107-0.example.net

[all]
test-ebe107-0.example.net hostname=test-ebe107-0.example.net
```

cc @p3ck to address his request from earlier today.

samvarankashyap:
@herlo : I found the bug but i am not sure how to fix it yet. I see its due to the gather_data role , because every time it provisions new resource groups its not updating the topology_outputs. Making run db contain only latest provisioned resource group. Its at the playbook level . 
### Pinfile
```yaml
---
dummy_target:
  topology: dummy-topology.yml
  layout: dummy-layout.yml
```
### topology
```
---
topology_name: "dummy"
resource_groups:
    resource_group_type: "dummy"
    resource_definitions:
      - role: "dummy_node"
        name: "web"
        count: 1
  - resource_group_name: "dummy2"
    resource_group_type: "dummy"
    resource_definitions:
      - role: "dummy_node"
        name: "interf"
        count: 1
```
### layout
```yaml
---
inventory_layout:
      vars:
        hostname: __IP__
      hosts:
        libvirt-node:
          count: 1
          host_groups:
            - dummy-layout
        dummy-node:
          count: 1
          host_groups:
            - aws-layout
      host_groups:
        all:
          vars:
            ansible_python_interpreter: /usr/bin/python3
            ansible_user: root
            ansible_ssh_common_args: -o StrictHostKeyChecking=no
```

The following is the rundb file generated 
```json
{
  "dummy_target": {
    "1": {
      "inputs": [
        {
          "topology_data": {
            "topology_name": "dummy",
            "resource_groups": [
              {
                "resource_group_name": "dummy",
                "resource_definitions": [
                  {
                    "count": 1,
                    "role": "dummy_node",
                    "name": "web"
                  }
                ],
                "resource_group_type": "dummy"
              },
              {
                "resource_group_name": "dummy2",
                "resource_definitions": [
                  {
                    "count": 1,
                    "role": "dummy_node",
                    "name": "interf"
                  }
                ],
                "resource_group_type": "dummy"
              }
            ]
          }
        },
        {
          "layout_data": {
            "inventory_layout": {
              "hosts": [
                {
                  "count": 1,
                  "name": "libvirt-node",
                  "host_groups": [
                    "dummy-layout"
                  ]
                },
                {
                  "count": 1,
                  "name": "dummy-node",
                  "host_groups": [
                    "aws-layout"
                  ]
                }
              ],
              "host_groups": {
                "all": {
                  "vars": {
                    "ansible_ssh_common_args": "-o StrictHostKeyChecking=no",
                    "ansible_python_interpreter": "\/usr\/bin\/python3",
                    "ansible_user": "root"
                  }
                }
              },
              "vars": {
                "hostname": "__IP__"
              }
            }
          }
        }
      ],
      "end": "1532962191.71",
      "cfgs": [
        
      ],
      "uhash": "581094",
      "start": "1532962185.29",
      "rc": 0,
      "action": "up",
      "outputs": [
        {
          "resources": {
            "ovirt_vms_res": [
              
            ],
            "rax_server_res": [
              
            ],
            "aws_ec2_res": [
              
            ],
            "os_server_res": [
              
            ],
            "os_keypair_res": [
              
            ],
            "aws_ec2_key_res": [
              
            ],
            "gcloud_gce_res": [
              
            ],
            "aws_s3_res": [
              
            ],
            "duffy_res": [
              
            ],
            "os_sg_res": [
              
            ],
            "dummy_res": [
              {
                "failed": false,
                "changed": true,
                "hosts": [
                  "web-581094-0.example.net"
                ],
                "dummy_file": "\/tmp\/dummy.hosts"
              }
            ],
            "beaker_res": [
              
            ],
            "aws_cfn_res": [
              
            ],
            "os_heat_res": [
              
            ],
            "os_obj_res": [
              
            ],
            "libvirt_res": [
              
            ],
            "os_volume_res": [
              
            ]
          }
        },
        {
          "inventory_path": [
            "\/home\/srallaba\/workspace\/lp_ws\/env_vars\/workspace\/inventories\/dummy-581094.inventory"
          ]
        },
        {
          "resources": {
            "ovirt_vms_res": [
              
            ],
            "rax_server_res": [
              
            ],
            "aws_ec2_res": [
              
            ],
            "os_server_res": [
              
            ],
            "os_keypair_res": [
              
            ],
            "aws_ec2_key_res": [
              
            ],
            "gcloud_gce_res": [
              
            ],
            "aws_s3_res": [
              
            ],
            "duffy_res": [
              
            ],
            "os_sg_res": [
              
            ],
            "dummy_res": [
              {
                "failed": false,
                "changed": true,
                "hosts": [
                  "interf-581094-0.example.net"
                ],
                "dummy_file": "\/tmp\/dummy.hosts"
              }
            ],
            "beaker_res": [
              
            ],
            "aws_cfn_res": [
              
            ],
            "os_heat_res": [
              
            ],
            "os_obj_res": [
              
            ],
            "libvirt_res": [
              
            ],
            "os_volume_res": [
              
            ]
          }
        },
        {
          "inventory_path": [
            "\/home\/srallaba\/workspace\/lp_ws\/env_vars\/workspace\/inventories\/dummy-581094.inventory"
          ]
        }
      ]
    }
  },
  "linchpin": {
    "1": {
      "action": "up",
      "targets": [
        {
          "dummy_target": {
            "1": {
              "uhash": "581094",
              "rc": 0
            }
          }
        }
      ]
    }
  }
}
```

Thus, Due to errors in rundb , it generates wrong inventory !

herlo:
The above code is *not* nested, but two entries into the outputs field called resources. They just need to be combined or looped over.

I think I'd like to see a single list in outputs, called 'resources', kind of like what was there before. I know there were reasons for adding *all* of the resource outputs, but it seems to be more painful and overly verbose. When adding outputs, linchpin should just update the resources section with the new resource and keep what is already there. A simple list seems easier, like so:

```
           "outputs": [
                {
                    "resources": [
                        {
                            "changed": true,
                            "dummy_file": "/tmp/dummy.hosts",
                            "failed": false,
                            "hosts": [
                                "web-be07-0.example.net",
                                "web-be07-1.example.net",
                                "web-be07-2.example.net"
                            ]
                        },
                        {
                            "changed": true,
                            "dummy_file": "/tmp/dummy.hosts",
                            "failed": false,
                            "hosts": [
                                "test-be07-0.example.net"
                            ]
                        }
                    ]
                }
            ],
```

If you feel that more data is necessary, let's discuss it and get it right. I'm also happy to help you get the updating features right within the playbooks if they aren't actually doing what they're supposed to be doing.

samvarankashyap:
@herlo : latest commit should fix the invalid inventory generation. I have come up with simple and more efficient fix. I have verified it with dummy provider. Could you please verify it with other providers.

p3ck:
@samvarankashyap I'll test this with the beaker-openstack topology

herlo:
@samvarankashyap Interesting fix. Could you provide what the rundb records look like now? I'm curious how they differ from the previous structure without changes to the gather_resources playbook.

samvarankashyap:
@herlo 
### topology
```yaml
---
topology_name: "dummy"
resource_groups:
  - resource_group_name: "dummy1"
    resource_group_type: "dummy"
    resource_definitions:
      - role: "dummy_node"
        name: "web"
        count: 4
  - resource_group_name: "dummy2"
    resource_group_type: "dummy"
    resource_definitions:
      - role: "dummy_node"
        name: "interf"
        count: 3
  - resource_group_name: "dummy3"
    resource_group_type: "dummy"
    resource_definitions:
      - role: "dummy_node"
        name: "extern"
        count: 2
```

### layout 
```yaml
---
inventory_layout:
      vars:
        hostname: __IP__
      hosts:
        libvirt-node:
          count: 2
          host_groups:
            - dummy-layout
        dummy-node:
          count: 3
          host_groups:
            - aws-layout
      host_groups:
        all:
          vars:
            ansible_python_interpreter: /usr/bin/python3
            ansible_user: root
            ansible_ssh_common_args: -o StrictHostKeyChecking=no
```

### rundb
```json
    "dummy_target": {
        "1": {
            "action": "up",
            "cfgs": [],
            "end": "1533063270.73",
            "inputs": [
                {
                    "topology_data": {
                        "resource_groups": [
                            {
                                "resource_definitions": [
                                    {
                                        "count": 4,
                                        "name": "web",
                                        "role": "dummy_node"
                                    }
                                ],
                                "resource_group_name": "dummy1",
                                "resource_group_type": "dummy"
                            },
                            {
                                "resource_definitions": [
                                    {
                                        "count": 3,
                                        "name": "interf",
                                        "role": "dummy_node"
                                    }
                                ],
                                "resource_group_name": "dummy2",
                                "resource_group_type": "dummy"
                            },
                            {
                                "resource_definitions": [
                                    {
                                        "count": 2,
                                        "name": "extern",
                                        "role": "dummy_node"
                                    }
                                ],
                                "resource_group_name": "dummy3",
                                "resource_group_type": "dummy"
                            }
                        ],
                        "topology_name": "dummy"
                    }
                },
                {
                    "layout_data": {
                        "inventory_layout": {
                            "host_groups": {
                                "all": {
                                    "vars": {
                                        "ansible_python_interpreter": "/usr/bin/python3",
                                        "ansible_ssh_common_args": "-o StrictHostKeyChecking=no",
                                        "ansible_user": "root"
                                    }
                                }
                            },
                            "hosts": [
                                {
                                    "count": 2,
                                    "host_groups": [
                                        "dummy-layout"
                                    ],
                                    "name": "libvirt-node"
                                },
                                {
                                    "count": 3,
                                    "host_groups": [
                                        "aws-layout"
                                    ],
                                    "name": "dummy-node"
                                }
                            ],
                            "vars": {
                                "hostname": "__IP__"
                            }
                        }
                    }
                }
            ],
            "outputs": [
                {
                    "resources": {
                        "aws_cfn_res": [],
                        "aws_ec2_key_res": [],
                        "aws_ec2_res": [],
                        "aws_s3_res": [],
                        "beaker_res": [],
                        "duffy_res": [],
                        "dummy_res": [
                            {
                                "changed": true,
                                "dummy_file": "/tmp/dummy.hosts",
                                "failed": false,
                                "hosts": [
                                    "web-2c8411-0.example.net",
                                    "web-2c8411-1.example.net",
                                    "web-2c8411-2.example.net",
                                    "web-2c8411-3.example.net"
                                ]
                            },
                            {
                                "changed": true,
                                "dummy_file": "/tmp/dummy.hosts",
                                "failed": false,
                                "hosts": [
                                    "interf-2c8411-0.example.net",
                                    "interf-2c8411-1.example.net",
                                    "interf-2c8411-2.example.net"
                                ]
                            },
                            {
                                "changed": true,
                                "dummy_file": "/tmp/dummy.hosts",
                                "failed": false,
                                "hosts": [
                                    "extern-2c8411-0.example.net",
                                    "extern-2c8411-1.example.net"
                                ]
                            }
                        ],
                        "gcloud_gce_res": [],
                        "libvirt_res": [],
                        "os_heat_res": [],
                        "os_keypair_res": [],
                        "os_obj_res": [],
                        "os_server_res": [],
                        "os_sg_res": [],
                        "os_volume_res": [],
                        "ovirt_vms_res": [],
                        "rax_server_res": []
                    }
                },
                {
                    "inventory_path": [
                        "/home/srallaba/workspace/lp_ws/env_vars/workspace/inventories/dummy-2c8411.inventory"
                    ]
                }
            ],
            "rc": 0,
            "start": "1533063261.53",
            "uhash": "2c8411"
        }
    },
    "linchpin": {
        "1": {
            "action": "up",
            "targets": [
                {
                    "dummy_target": {
                        "1": {
                            "rc": 0,
                            "uhash": "2c8411"
                        }
                    }
                }
            ]
        }
    }
```

samvarankashyap:
#### Running on two os_server resource groups: Inventory generates flawlessly
@p3ck @herlo

samvarankashyap:
@p3ck @herlo : 
#### Test: Dummy + openstack : works

samvarankashyap:
@herlo @p3ck : 
#### Test: dummy + openstack + aws_ec2 : Works
^^^ I created a topology which has three of the above providers together in different resource groups and was successfully able to create inventory out of it !

p3ck:
@samvarankashyap beaker+beaker generates the correct inventory in version 1.5.4 so clearly we have a regression here and can't release 1.5.5 until this is fixed or rolled back.

samvarankashyap:
@herlo @p3ck : With latest set of commits the beaker bug should be fixed . 
could you please verify . Thanks!

p3ck:
@samvarankashyap I verified beaker+beaker works as well.  beaker+openstack works as well, but the layout is backwards...  for example:

topology requests beaker and openstack in that order.  The layout has the following:
  layout:
    inventory_layout:
      vars:
        hostname: __IP__
      hosts:
        beaker-node:
          count: 1
          host_groups:
            - beaker-layout
        openstack-node:
          count: 1
          host_groups:
            - openstack-layout

But the inventory generated is:
[beaker-layout]
10.8.248.226 hostname=10.8.248.226

[openstack-layout]
dell-per410-01.khw.lab.eng.bos.redhat.com hostname=dell-per410-01.khw.lab.eng.bos.redhat.com

I believe the order is just reversed?

samvarankashyap:
@p3ck order should depend on order in which the topology groups are defined , 
could you share beaker openstack topology for reference ?

herlo:
@samvarankashyap, it looks like this is still failing. However, it's not the more complex beaker+beaker, etc. topologies, but a simpler one. 

```
21:01:14 [fedora26] TASK [beaker : Get topology output data from RunDB] ****************************
21:01:14 [fedora26] changed: [localhost] => {"changed": true, "output": [{"bkr_id_values": [{"id": "2661383", "url": "https://beaker.engineering.redhat.com/jobs/2661383"}]}, {"resources": {"aws_cfn_res": [], "aws_ec2_key_res": [], "aws_ec2_res": [], "aws_s3_res": [], "beaker_res": [{"arch": "x86_64", "distro": "RHEL-6.5", "family": "RedHatEnterpriseLinux6", "id": "2661383", "name": "rhel65", "result": "Pass", "status": "Running", "system": "dell-per220-07.cci.lab.eng.bos.redhat.com", "url": "https://beaker.engineering.redhat.com/jobs/2661383", "variant": "Client"}], "duffy_res": [], "dummy_res": [], "gcloud_gce_res": [], "libvirt_res": [], "os_heat_res": [], "os_keypair_res": [], "os_obj_res": [], "os_server_res": [], "os_sg_res": [], "os_volume_res": [], "ovirt_vms_res": [], "rax_server_res": []}}, {"inventory_path": ["/workDir/workspace/ci-linchpin/linchpin/docs/source/examples/workspace/inventories/bkr-new-de7e82.inventory"]}]}
21:01:14 [fedora26] 
21:01:14 [fedora26] TASK [beaker : set topo_output_resources fact] *********************************
21:01:14 [fedora26] fatal: [localhost]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'dict object' has no attribute 'resources'\n\nThe error appears to have been in '/usr/lib/python2.7/site-packages/linchpin/provision/roles/beaker/tasks/teardown_resource_group.yml': line 12, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: \"set topo_output_resources fact\"\n  ^ here\n"}
21:01:14 [fedora26] 	to retry, use: --limit @/usr/lib/python2.7/site-packages/linchpin/provision/beaker.retry
21:01:14 [fedora26] 
21:01:14 [fedora26] PLAY RECAP *********************************************************************
21:01:14 [fedora26] localhost                  : ok=7    changed=1    unreachable=0    failed=1   
21:01:14 [fedora26] 
21:01:14 [fedora26] 
21:01:14 [fedora26] ID: 10
21:01:14 [fedora26] Action: destroy
21:01:14 [fedora26] 
21:01:14 [fedora26] Target              	Run ID	uHash	Exit Code 
21:01:14 [fedora26] -------------------------------------------------
21:01:14 [fedora26] beaker-new          	     2	de7e82	        2
21:01:14 [fedora26] 
```
This error is consistent across all distribution tests. Here's the console log if it helps.  https://jenkins-continuous-infra.cloud.continuous-infra.upshift.redhat.com/job/ci-linchpin/81/consoleFull

samvarankashyap:
[test]

samvarankashyap:
[test]

samvarankashyap:
@p3ck @herlo At last .....  success !

p3ck:
+1 good job!!

samvarankashyap:
[merge]

PR #608, opened 2018-06-18
ks_meta in recipesets is valid schema.

p3ck:
While trying to provision a resource in beaker using the restraint harness I found that linchpin doesn't accept the ks_meta param in topology.


p3ck:
[merge]

PR #609, opened 2018-06-20
Run one resource at a time

herlo:
During the run of _invoke_playbooks, the resources were set to all resources, but only the current resource is needed. This one line adjustment removes these failures.


herlo:
[merge]

Issue #304, opened 2017-07-12
Maintain a list of versions on dockerhub


vi-patel:
Instead of just having a latest tag, can a list of significant releases also be maintained as tags on dockerhub.

Just for the reason of if I know my tests with a specific version, I can keep using that tagged container, and move up to later releases when ready, instead of possibly running into issues if i keep trying to use the latest tag.

herlo:
@samvarankashyap, what say you?

herlo:
For LinchPin 2.0 (likely, but possibly earlier), we're going to be building a few docker containers for running linchpin. With the Fedora Modularity and Factory 2.0 initiatives in full swing, we're likely to build the containers and storing them as part of the new Fedora tools.

herlo:
This will be addressed by #613

PR #613, opened 2018-07-02
Linchpin Container Build

### LINKED ISSUE #602: Containerized linchpin not working in Openshift
I've successfully installed linchpin inside a container running in a pod on openshift, however executing linchpin up, without any scc modifications, results in the following error:

```
PLAY [schema check and Pre Provisioning Activities on topology_file] ***********

TASK [Gathering Facts] *********************************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: KeyError: 'getpwuid(): uid not found: 1000060000'
fatal: [localhost]: FAILED! => {"msg": "Unexpected failure during module execution.", "stdout": ""}
 [WARNING]: Could not create retry file '/usr/lib/python2.7/site-
packages/linchpin/provision/aws.retry'.         [Errno 13] Permission denied:
'/usr/lib/python2.7/site-packages/linchpin/provision/aws.retry'
```

This appears to be related to the ```default``` user that openshift uses when executing containers and permissions?

### LINKED ISSUE #623: There should be a way in linchpin.conf to allow absolute path to resources
While testing the container build for linchpin, I discovered that when writing out data for linchpin.distilled and linchpin.latest files, it always writes to `/path/to/workspace/resources/`. While this is generally useful outside of a container, I'd like to put all of my outputs into one place, not the workspace. 

Maybe a setting like `resources_path = /path/to/resources`. Similar to modifying the inventory location with `inventory_path = /path/to/inventories/my.inventory`

herlo:
Initial Jenkins Pipeline for building the official LinchPin container.

The container images currently built using the Dockerfile/Jenkinsfile are tagged and exist [here](https://hub.docker.com/r/herlo/linchpin-container/tags/)

To use the container image, it can be run by starting the container with the `-it` option, or by starting the container and then running a `docker exec` on the running container. Here is a simple example to get the container to run:

```
docker run -it -v /path/to/workspace/:/workdir/ herlo/linchpin-container \
-- linchpin -c /etc/linchpin.conf -vvvp PinFile.dummy.yml \
-w /workdir up dummy-topo
```

**NOTE**: The `--` allows the entire linchpin command to be passed as an argument to `docker run`

Some things to consider/discuss:

* To build this image, the jenkins slave container needs to include buildah, which is used to build containers without a docker daemon. This has been added to the Dockerfile in this PR.
* This container won't work for a local libvirt
  * The specific case that might work would be a remote libvirt connection and using the remote bits of a libvirt build, which has not been tested
* The name of the container may change. Currently, it's 'linchpin-container'
  * Having an upstream release, and then building the 'linchpin-executor' container (on which this build is [based](https://github.com/herlo/contra-env-sample-project/blob/master/config/dockerfiles/linchpin/Dockerfile)) may be desirable. This container could then work wherever, and another layer could add the stuff for running in openshift, adding containerized libvirt support, etc. It currently feels too heavy to do that for the official container. It feels like the container image should be lightweight and flexible.
* Using the linchpin-container:latest on a machine without the UID provided in the Dockerfile may cause the container to fail
  * This can be remedied by running the docker container as a different user
  * This seems like a blocker for users who just want to use the container unmodified
  * A README is probably needed to explain these items, or adjustments to the Dockerfile are needed
  * As stated above, simplifying this Dockerfile to a minimalistic version would then allow others to base LinchPin container builds on this one.

Fixes #623 
Fixes #602


config/Dockerfiles/JenkinsfileContainerBuild
```diff
@@ -0,0 +1,209 @@
...
+                                      extensions                       : [],
+                                      submoduleCfg                     : [],
+                                      userRemoteConfigs                : [
+                                              [refspec:
+                                                       '+refs/heads/*:refs/remotes/origin/*  +refs/pull/*:refs/remotes/origin/pr/* ',
+                                               url: "https://github.com/${env.ghprbGhRepository}"]
+                                      ]
+                            ])
+                            def creds = com.cloudbees.plugins.credentials.CredentialsProvider.lookupCredentials(
+                              com.cloudbees.plugins.credentials.Credentials.class
```
p3ck: Is creds used for anything here?

config/Dockerfiles/linchpin/Dockerfile
```diff
@@ -0,0 +1,55 @@
+FROM fedora:latest
+LABEL name="contrainfra/linchpin-executor" \
+      maintainer="https://github.com/CentOS-PaaS-SIG/linchpin" \
+      version="0.0.5" \
+      description="LinchPin executable container running Fedora 28"
```
p3ck: should we just say latest fedora here since the FROM is fedora:latest?

config/Dockerfiles/linchpin/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+        net-tools \
+        python-krbV \
+        libxslt-python \
+        libxml2-python \
+        krb5-workstation \
+        libselinux-python \
+        && dnf clean all; \
+        pip install -U pip; \
+        pip install -U setuptools; \
+        pip install linchpin; \
```
p3ck: I know you query the version out of the container later but this ties us to only release containers of what has been pushed to pypi.  What if you want to build a rc or pre-release container?  I think this should install from the git checkout that the jenkinsfile did.

config/Dockerfiles/linchpin/bin/run
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+id
+whoami
+exec tail -f /dev/null
```
p3ck: do we really want this container to stick around if they don't specify a command?

config/Dockerfiles/linchpin/conf/linchpin.conf
```diff
@@ -0,0 +1,296 @@
...
+# If desired, one could overwrite the location of the generated inventory path
+#inventory_path = {{ workspace }}/{{inventories_folder}}/happy.inventory
+inventory_path = /tmp/{{ topology_name }}-{{ uhash }}.inventory
+
+# Libvirt images can be stored almost anywhere (not /tmp).
+# Unprivileged users need not setup sudo to manage a path to which they have rights.
+# The following are specific settings to manage libvirt images and instances
+
+# the location to store generated ssh keys and the like
+#default_ssh_key_path = ~/.ssh
```
p3ck: I wonder if this should be using the workdir/ssh ?  Then the user can pass in their own keys.

config/Dockerfiles/linchpin/conf/ansible.cfg
```diff
@@ -0,0 +1,5 @@
+[defaults]
```
p3ck: I'd rather see this read from the workdir or at least allow the user to override it there.  I may want key checking

config/Dockerfiles/linchpin/Dockerfile
robnester-rh: There is an argument that [containers shouldn't be tagged to "latest"](https://developers.redhat.com/blog/2016/02/24/10-things-to-avoid-in-docker-containers/)

config/Dockerfiles/linchpin/conf/ansible.cfg
robnester-rh: @p3ck - To be fair, I don't think you would, as that would require you to be logged into the container while SSH'ing.

config/Dockerfiles/JenkinsfileContainerBuild
herlo: For pushing up to the docker hub? I'm not sure if this is the credentials store in Jenkins, but that's what I believe it is fore.

config/Dockerfiles/linchpin/Dockerfile
herlo: Yeah. I plan to change that to the latest current version of Fedora. In this case that should say `fedora:28` in the FROM section. I can update the description here. 

I also wanted to change the name to just linchpin (or linchpin-container), but wanted to have a discussion about how this would be the upstream container image, which others could build upon. I'm not sure of the good/bad of doing that so I wanted opinions.

config/Dockerfiles/linchpin/Dockerfile
herlo: We could easily enough put in an environment variable here that allows you to use the `git+ssh://github.com...` syntax, leaving `linchpin` as default. Would that be acceptable? It would also allow jenkins to not use the the 'latest' tag option which is currently provided.

I personally think this would give us the most flexibility because my POV on this was that we'd be doing the pypi release first, then pulling it down, testing it, and running the container build. I don't know that we want 'official' builds from git like that.

config/Dockerfiles/linchpin/bin/run
herlo: This is based upon what ari et al. has done. I didn't have a better option. However, one of the things that could be done is to do a `docker exec` if the container is running in this way. Not that using `docker run -it ...` is a bad way to go. But more options seems better to me. Plus, the ability to run the container in openshift was that you'd likely run it with `docker exec` aiui from discussions with @arilivigni .

config/Dockerfiles/linchpin/conf/ansible.cfg
herlo: Environment variables could be useful here also. We can obviously discuss this further.

config/Dockerfiles/JenkinsfileContainerBuild
p3ck: I don't think creds is used at all.  You have another withCredentials down below which does look like it's used for pushing to docker.

config/Dockerfiles/linchpin/bin/run
p3ck: ok - fair enough.

config/Dockerfiles/linchpin/conf/linchpin.conf
herlo: This line is only used currently in the [libvirt provider](https://github.com/CentOS-PaaS-SIG/linchpin/blob/develop/linchpin/provision/roles/libvirt/tasks/provision_libvirt_node.yml#L54). Since this container doesn't support libvirt within it yet.

herlo:
[test]

herlo:
@p3ck could you give this another review? I want to have this be the upstream `linchpin` container, from which all other linchpin containers (including the contrainfra/linchpin-executor) builds. There may be some adjustments needed to meet that need, but I would be uncertain as to their nature.

herlo:
Fixes #304

herlo:
[test]

herlo:
@p3ck I have made fundamental changes to this PR. Can you please review it again?

config/Dockerfiles/JenkinsfileContainerBuild
```diff
@@ -163,8 +161,13 @@ podTemplate(name: podName,
                                 script: "pwd",
                                 returnStdout: true
                             ).trim()
-                            sh "buildah run -v ${cwd}/docs/source/examples/workspace:/workdir ${containerid} -- linchpin -c /etc/linchpin.conf -vvvp PinFile.dummy.yml -w /workdir up dummy-topo"
-
+                            sh "buildah copy --chown 10001:0 ${containerid} ${cwd}/config/Dockerfiles/linchpin/bin/run-linchpin /linchpin/bin/run"
```
p3ck: why is this copied here as a separate step?  why not do it in the Jenkinsfile?

also - I think the idea of the entry_user_id script is to create the user and make sure that users primary group is 0 (root) and all files should be g+rw.  This way it doesn't matter what the uid is and you shouldn't bother with this 10001 uid..

config/Dockerfiles/JenkinsfileContainerBuild
```diff
@@ -163,8 +161,13 @@ podTemplate(name: podName,
...
+                            sh "buildah commit ${containerid} linchpin-test"
+                            string testcontainerid = sh (
+                                script: "buildah from linchpin-test",
+                                returnStdout: true
+                            ).trim()
+                            sh "buildah run -v ${cwd}/docs/source/examples/workspace:/workdir ${testcontainerid} uid_entrypoint run"
```
p3ck: this doesn't see, right to me..  The whole point of an entrypoint is that you don't have to specify it.  you should only have to specify run piece

config/Dockerfiles/linchpin/bin/run-linchpin
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+id
+linchpin -c /etc/linchpin.conf -vvvp PinFile.dummy.yml -w /workdir up dummy-topo
```
p3ck: I don't see why this needs to be put into the container.  why not pass this in when you call buildah run?

config/Dockerfiles/JenkinsfileContainerBuild
herlo: To test with buildah, I added the linchpin-run command. There's not a way to run buildah exec lik there is in docker. To test the container, I had to run a linchpin command instead of `tail -f /dev/null` and I couldn't see another way.

It's actually not used in the final container image. If you look further, you may see that this creates a new container when I run commit. The copy is against a container and not the image. Furthermore, the image that is pushed to the docker registry is indeed the linchpin image, and not the linchpin-test image.

config/Dockerfiles/JenkinsfileContainerBuild
herlo: Again, this is only for testing. In a proper use of the item, you'd have docker exec (or run it in openshift) and this wouldn't be an issue.

config/Dockerfiles/linchpin/bin/run-linchpin
herlo: That's exactly what I did previously. However, you pointed out that I was breaking functionality to make this work. While this bit is really only for testing, as discussed above, it's necessary because I needed to make sure the `uid_entrypoint` functionality actually ran as the UID 10001 and not root or some other uid.

config/Dockerfiles/JenkinsfileContainerBuild
robnester-rh: @p3ck This, I think, is mimicking the approach used in the existing docker file. In this scenario, the entrypoint is set to uid_entrypoint to handle adding the nonroot user to /etc/passwd, etc. That script the runs whatever else is specified in the CMD element. "run" is an arbitrary script that keeps the docker running if no other command is specified when executing the container.

config/Dockerfiles/linchpin/bin/run-linchpin
robnester-rh: +1

config/Dockerfiles/JenkinsfileContainerBuild
```diff
@@ -0,0 +1,215 @@
...
+                ttyEnabled: false,
+                privileged: true,
+                alwaysPullImage: true,
+                args: '${computer.jnlpmac} ${computer.name}',
+                command: '',
+                workingDir: '/workDir')
+        ],
+        volumes: [
+            emptyDirVolume(mountPath: '/var/run/containers', memory: false),
+            emptyDirVolume(mountPath: '/var/lib/containers', memory: false)
```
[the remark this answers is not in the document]
herlo: @robnester-rh it's not duplicated. One is `/var/lib` the other `/var/run`. Both are needed for buildah to work in a containerized environment.

herlo:
[merge]

Issue #612, opened 2018-06-28
bkr topology: Introduce support for 'family' and 'tag' fields in bkr topology files


yashkmankad:
The Beaker (bkr) provider for linchpin currently supports using the 'distro' parameter under recipesets, but does not support 'family' and 'tag'.

Example beaker topology file with 'family' and 'tag' would look like:

```
---
topology_name: beaker
resource_groups:
  - resource_group_name: test1
    res_group_type: beaker
    job_group: your-beaker-group
    recipesets:
      - family: RedHatEnterpriseLinuxServer6
        tag: RELEASED
        arch: x86_64
        keyvalue:
          - MEMORY>1000
          - DISKSPACE>20000
        count: 1
```

Related documentation about using the family and tag options with the Beaker (bkr) command line tool:
 * https://beaker-project.org/docs/man/bkr-distros-list.html#cmdoption-bkr-distros-list-family
 * https://beaker-project.org/docs/man/bkr-job-list.html#cmdoption-bkr-job-list-family

yashkmankad:
Issue resolved with https://github.com/CentOS-PaaS-SIG/linchpin/pull/614
Closing issue.

Thanks Bill!

Issue #611, opened 2018-06-26
bkr topology: allow boolean expressions on hostrequire


wainersm:
Currently the topology for beaker provision does not allow to pass boolean expressions in hostrequire for selecting the host.

For example, suppose you want to select a host by name like in below Beaker XML:
<or><system><name op='like' value='%gigabyte%'/></system><system><name op='like' value='%thunderx%'/></system><system><name op='like' value='%hawker%'/></system></or>

So I need the ability to pass such as query to hostrequire. It can even be a parameter that you can inject snippets of Beaker XML.

herlo:
Fixed by #614

PR #614, opened 2018-07-02
Beaker enhancement

p3ck:
Includes the following:

reserve_duration: can be specified so that resources can be held onto for longer than the default 24 hours
#612 is supported in this PR
#611  is supported in this PR.


herlo:
I think I see what was going on here. You had 'false' for both of those in the interim before finding the excludes option. [This](https://github.com/CentOS-PaaS-SIG/linchpin/pull/614/commits/013a0169ae027d8ff65659a310f7ff83201b7573) is why I thought that was the case.

p3ck:
any idea why travis exited non-zero?  the test cases look good.

anascko:
I apologize for the intervention
But we tested your changes regarding family and tags 
and have some issue:
Is it the correct format of schema.json for the tag?
https://github.com/beaker-project/beaker/blob/develop/Client/src/bkr/client/__init__.py#L218-L222
And if we used family instead of a distro and tag not specified then beaker will be used default tag

p3ck:
I have tested this morning using the following in the schema
          recipesets:
            - family: "Fedora28"
              tags:
                - released

Which works as expected.  The latest schema update will require family or distro.

anascko:
It works because it uses default tag, 
if you looked on beaker xml of this job, you don't find your provided tag

p3ck:
@anascko I don't understand your statement.  Here is the xml snippet from the job I submitted via linchpin.
`      <distroRequires>
        <and>
          <distro_family op="=" value="Fedora28"/>
          <distro_tag op="=" value="released"/>
          <distro_variant op="=" value="Server"/>
          <distro_arch op="=" value="x86_64"/>
        </and>
      </distroRequires>`

You can see the distro_tag is included.  I agree it would have been missing before my last commit.  So please update and try again.

anascko:
@p3ck Thanks,
I apologize, I didn't update the local repository with last commit afc0209773da56f215928637074479a762b297e7, that is why I still had a problem
Now tag works, thanks for the fix

herlo:
@p3ck  Travis CI failed here: https://travis-ci.org/CentOS-PaaS-SIG/linchpin/builds/400885093#L645

config/Dockerfiles/tests.d/beaker/02_beaker_distro
```diff
@@ -0,0 +1,23 @@
...
+TMP_FILE=$(mktemp)
+
+function clean_up {
+    set +e
+    linchpin -w . -p "${PINFILE}" -v destroy "${DTARGET}" "${FTARGET}"
+}
+trap clean_up EXIT SIGHUP SIGINT SIGTERM
+
+pushd docs/source/examples/workspace
+linchpin -w . -p "${PINFILE}" -v up "${DTARGET}" "${FTARGET}"
```
herlo: How is this test different from https://github.com/p3ck/linchpin/blob/23729fbe23f430fb0e9f4c4e28a5b1bc59009a77/config/Dockerfiles/tests.d/general/01_general ?

herlo: @p3ck nvm, I see the target data.

p3ck:
[merge]

p3ck:
#615 is fixed by this PR

PR #621, opened 2018-07-17
Replace PyYAML, python-requests, python-ipaddress with RPMs (centos7)

herlo:
The CentOS7 container would not succecssfully build the linchpin executable after some sort of change recently. This update uses custom built rpms from a COPR repository, along with a CentOS7 (openstack-pike) repository to get linchpin to install again.

The reason this is a problem is that pip will not uninstall certain distutils packages (the above named are the ones dealt with here) using pip. RPM dependencies cause things to be uninstalled which are needed, and CentOS7 core doesn't have new enough RPMs.


herlo:
[merge]

PR #624, opened 2018-07-19
Allow absolute path to resources. Fixes issue #623

samvarankashyap:
Refer #623

### DIFF
--- a/linchpin/cli/__init__.py
+++ b/linchpin/cli/__init__.py
@@ -186,15 +186,19 @@ class LinchpinCli(LinchpinAPI):
 
         dist_roles = self.get_cfg('distiller')
 
-        resources_path = self.get_evar('resources_folder')
-        if os.path.isabs(resources_path):
-            context_file = '{0}/{1}'.format(resources_path,
-                                            'linchpin.distilled')
-        else:
-            context_path = '{0}/{1}'.format(self.workspace, resources_path)
-            if not os.path.exists(context_path):
-                os.makedirs(context_path)
-            context_file = '{0}/{1}'.format(context_path, 'linchpin.distilled')
+        resources_folder = self.get_evar('resources_folder',
+                                         default='resources')
+        context_path = '{0}/{1}'.format(self.workspace, resources_folder)
+
+        resources_path = os.path.expanduser(
+            self.get_evar('default_resources_path'))
+
+        if (os.path.isabs(resources_path)):
+                context_path = resources_path
+        if not os.path.exists(context_path):
+            os.makedirs(context_path)
+
+        context_file = '{0}/{1}'.format(context_path, 'linchpin.distilled')
 
         roles = []
         dist_data = {}
--- a/linchpin/cli/__init__.py
+++ b/linchpin/cli/__init__.py
@@ -187,10 +187,14 @@ class LinchpinCli(LinchpinAPI):
         dist_roles = self.get_cfg('distiller')
 
         resources_path = self.get_evar('resources_folder')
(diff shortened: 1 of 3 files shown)


samvarankashyap:
@herlo : once this PR is merged, I will be re basing and updating the other PR's accordingly since , linchpin.latest is part of inventory gen PR #601 .

herlo:
I was wondering if a more general solution would be applicable in that case. Specifically, if we created a variable called `resources_path` as described in #623 which would be picked up if set? In this way, the other items could use that same location, but wouldn't have to affect the logic in all places in code. 

Essentially, rather than checking to see if the path is absolute (also it would need to be expanded in some cases), just check to see if the variable `resources_path` is set, and if so, use that value.

linchpin/cli/__init__.py
```diff
@@ -187,10 +187,14 @@ def _write_distilled_context(self, run_data):
...
-        if not os.path.exists(context_path):
-            os.makedirs(context_path)
-        context_file = '{0}/{1}'.format(context_path, 'linchpin.distilled')
+        if os.path.isabs(resources_path):
+            context_file = '{0}/{1}'.format(resources_path,
+                                            'linchpin.distilled')
+        else:
+            context_path = '{0}/{1}'.format(self.workspace, resources_path)
+            if not os.path.exists(context_path):
+                os.makedirs(context_path)
```
herlo: It's possible this `resources_path` may not exist. Might be good to make sure it's created before putting anything into that path as well.

herlo:
So, I started digging further into this idea, and got a little confused. I noticed a few things that weren't going to work (os.makedirs for only the `else` means that the `resources_path` might not get created). Digging further, I started to wonder why not use a different component in the `linchpin.conf`. I then dug into [`linchpin.constants`](https://github.com/herlo/linchpin/blob/linchpin_container/linchpin/linchpin.constants#L62) and noticed that there was already a `default_resources_path` key provided. 

After thinking about it a minute, I rewrote this method above so that it would do the same thing, but allow for overriding these values, and would create the proper directory if it didn't exit. Let me know what you think:

```
        resources_folder = self.get_evar('resources_folder', default='resources')
        context_path = '{0}/{1}'.format(self.workspace, resources_folder)

        resources_path = self.get_evar('default_resources_path')
        if (resources_path.startswith('/')):
                context_path = resources_path
        if not os.path.exists(context_path):
            os.makedirs(context_path)

        context_file = '{0}/{1}'.format(context_path, 'linchpin.distilled')
```

samvarankashyap:
@herlo The above code block looks more refined to me . 
My only question is what is the difference between 
```
os.path.isabs(resources_path) 
```
```
resources_path.startswith('/')
```
in above code block. Which one would be a better approach. 
I feel os.path.isabs might cover some edge cases too.

herlo:
@samvarankashyap I see your point. Actually, I think the fix is something like this:

```
        resources_path = os.path.expanduser(
                                self.get_evar('default_resources_path'))
        if (os.path.isabs(resources_path)):
```
I've already implemented everything in this [commit](https://github.com/CentOS-PaaS-SIG/linchpin/pull/613/commits/d6baa17f431721fb10e110f597ab5c97b7d89c4e) since I needed it for testing. Feel free to cherry-pick it into your PR. 

Note: If you don't use my commit, let me know so I can remove that commit from my PR.

- Let us not forget to expand the tildes (~). There's a lot of meaning in that one little character. :)

herlo:
Closing PR per conversations with @samvarankashyap - The fix is in #613

PR #632, opened 2018-07-25
Rework Credentials 

### LINKED ISSUE #620: The combined/01_beaker_openstack test fails on certain conditions
When running the downstream tests, the 01_beaker_openstack test will fail if the beaker provider is run before the openstack provider. In that case, the credentials are not expanded for the openstack provider. 

Simple solution, extract all credentials for all providers in one place prior to running any tests.

herlo:
Fixes #620

### DIFF
--- a/config/Dockerfiles/Jenkinsfile
+++ b/config/Dockerfiles/Jenkinsfile
@@ -34,7 +34,7 @@ OPENSHIFT_SERVICE_ACCOUNT = env.OPENSHIFT_SERVICE_ACCOUNT ?: 'jenkins'
 executionID = UUID.randomUUID().toString()
 
 // Pod name to use
-podName = 'linchpin' + executionID
+podName = 'linchpin_container-' + executionID
 
 library identifier: "ci-pipeline@master",
         retriever: modernSCM([$class: 'GitSCMSource',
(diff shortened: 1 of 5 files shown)


herlo:
[test]

herlo:
[test]

herlo:
[test]

config/Dockerfiles/Jenkinsfile
```diff
@@ -19,7 +19,7 @@ DOCKER_REPO_URL = env.DOCKER_REPO_URL ?: '172.30.254.79:5000'
 // if it's not defined you will get all the possible providers
 // listed here.
 
-LPROVIDERS = env.LINCHPIN_PROVIDERS ?: 'dummy duffy libvirt aws'
```
p3ck: why do we have to disable libvirt?

herlo: @p3ck Currently, we have a race condition where the libvirt provisioning happens for each distro in a single pod, sharing a single network. As we've discovered, there will be a race condition where one of the three distros will get the libvirt default network and the other two will then be unable to get IP addresses and timeout eventually. See #634 for more detail.

herlo:
[merge]

PR #640, opened 2018-08-02
Check that virt-install exists

14rcole:
When provisioning libvirt, the step to "Install VM" is set to ignore errors.  If virt-install does not exist on the system, there will not be an error until the VM attempts to get a dhcp ip address a few steps later. This fix checks that virt-install exists first so that the error is reported immediately.

### DIFF
--- a/config/Dockerfiles/centos7/Dockerfile
+++ b/config/Dockerfiles/centos7/Dockerfile
@@ -26,7 +26,7 @@ RUN yum -y install python-pip \
     #pip install -U 'requests>=2.14.2'; \
     #pip install -U 'PyYAML>=3.12'; \
     #pip uninstall -y urllib3; \
-    #yum -y install virt-install \
+    yum -y install virt-install \
     #echo 'pip freeze | grep requests'; \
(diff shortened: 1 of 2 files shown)


herlo:
I don't think this is a necessary fix. The centos7 container specifically didn't have this installed previously. Either I missed it, or it was never there (I probably missed it in the latest build). 

I'm working on a PR to go with the separate_pods branch and I'm testing it [here](https://github.com/herlo/linchpin/pull/1/commits/dcddc5beb55d85557f21350a3a528cc0b369ccf7#diff-e9765686542eccec4207d32f0d0170ceR19
). Should have something in a day or to to for virt-install on the container. 

In terms of user installation, they should be using the new linchpin container [here](https://github.com/CentOS-PaaS-SIG/linchpin/blob/develop/config/Dockerfiles/linchpin/Dockerfile).

I do need to get this container into the general public's hands though. That shouldn't take too long. It will land [here](https://hub.docker.com/r/contrainfra/linchpin/) when it does release.

14rcole:
Sounds good to me

Issue #511, opened 2018-03-21
auth_driver leaks secret keys


p3ck:
The auth_driver leaks secret keys when verbose mode is on.  Adding no_log: true to these tasks should prevent that.

TASK [aws : Get creds from auth driver] ****************************************

changed: [localhost] => {"changed": true, "gid": 0, "group": "root", "mode": "0755", "output": {"default": {"aws_access_key_id": "SECRETID", "aws_secret_access_key": "SECRETKEY"}}, "owner": "root", "params": {"cred_path": "/workDir/workspace/ci-pipeline-linchpin/linchpin/keys/aws-ec2-new", "cred_type": "aws", "driver": "file", "filename": "aws.key"}, "path": "/workDir/workspace/ci-pipeline-linchpin/linchpin/keys/aws-ec2-new", "size": 21, "state": "directory", "uid": 0}

TASK [aws : Set auth_var] ******************************************************

ok: [localhost] => {"ansible_facts": {"auth_var": {"aws_access_key_id": "SECRETID", "aws_secret_access_key": "SECRETKEY"}}, "changed": false}

There could be other tasks as well...

herlo:
AWS ones were cleaned up today in #520. However, a more formal clean up is necessary for all providers.

samvarankashyap:
@14rcole : 
I have been testing the above issue against my latest PR and foundout that , on failure of provisioning the instances( by giving incorrect parameters to the provision task ) the it still leaks the passwords and keys. 
Please make sure that you add ```no_log: true``` to provision tasks to prevent that from happening  
#631

14rcole:
@samvarankashyap pushed up a fix: #641

PR #641, opened 2018-08-06
add no_log to ansible tasks that may leak secret keys

14rcole:
Added no_log to some additional tasks.  Should fix anything left over from #511


14rcole:
@samvarankashyap PTAL

14rcole:
[merge]

PR #642, opened 2018-08-08
Update version to 1.6.0-beta

herlo:
Currently 1.6.0b1 since PR #601 is released. Need to get a few more
items in and run tests (which this update will cause)


samvarankashyap:
@herlo approved

PR #647, opened 2018-08-10
Add linchpin validate command

14rcole:
`linchpin validate` validates topologies and provides a list of errors if the topology does not successfully validate.  This PR also fixes some schema errors that prevent validation and updates cerberus to version 1.2, which is required in order to validate lists inside of schemas.

### DIFF
--- a/docs/source/advanced/validator.rst
+++ b/docs/source/advanced/validator.rst
@@ -1,0 +1,151 @@
+Validate Command
+-----------------
+
+The purpose of the validate command is to determine whether topologies are
+syntactically valid and, if not, provide a list of errors that occured during
+validation
+
+Using the validate command
+```````````````````````````
+
+The command `linchpin validate` looks at the topology file for each target in
+a given PinFile. If the topology is not valid under the current schema, it will
+attempt to convert the topology to an older schema and try again. If the
+topology is still invalid, the command will report the topology and a list of
+errors found.
+
+Invalid Topologies
+++++++++++++++++++
+
+Here is a simple PinFile and topology file. The topology file has some errors
+and will not validate.
+
+.. code-block:: yaml
+
+   ---
+   libvirt-new:
+      topology: libvirt-new.yml
+      layout: libvirt.yml
+
+   libvirt:
+      topology: libvirt.yml
+      layout: libvirt.yml
+
+   libvirt-network:
+      topology: libvirt-network.yml
+
+
+.. code-block:: yaml
+
+   ---
+   topology_name: libvirt-new
+   resource_groups:
+     - resource_group_name: libvirt-new
+       resource_group_type: libvirt
+       resource_definitions:
+         - role: libvirt_node
+           uri: qemu:///system
+           count: "1"
+           image_src: http://cloud.centos.org/centos/7/images/CentOS-7-x86_64-GenericCloud-1608.qcow2.xz
+           memory: 2048
+           vcpus: 1
+           arch: x86_64
+           ssh_key: libvirt
+           networks:
+             - name: default
+               additional_storage: 10G
+           cloud_config:
+             users:
+               - name: herlo
+                 gecos: Clint Savage
+                 groups: wheel
+                 sudo: ALL=(ALL) NOPASSWD:ALL
+                 ssh-import-id: gh:herlo
+                 lock_passwd: true
+
+.. code-block:: bash
+
+   $ linchpin validate
+   topology for target 'libvirt-network' is valid
+
+   Topology for target 'libvirt-new' does not validate
+   topology: 'OrderedDict([('topology_name', 'libvirt-new'), ('resource_groups', [OrderedDict([('resource_group_name', 'libvirt-new'), ('resource_group_type', 'libvirt'), ('resource_definitions', [OrderedDict([('role', 'libvirt_node'), ('uri', 'qemu:///system'), ('image_src', 'http://cloud.centos.org/centos/7/images/CentOS-7-x86_64-GenericCloud-1608.qcow2.xz'), ('memory', 2048), ('vcpus', '1'), ('arch', 'x86_64'), ('ssh_key', 'libvirt'), ('networks', [OrderedDict([('name', 'default'), ('hello', 'world')])]), ('additional_storage', '10G'), ('cloud_config', OrderedDict([('users', [OrderedDict([('name', 'herlo'), ('gecos', 'Clint Savage'), ('groups', 'wheel'), ('sudo', 'ALL=(ALL) NOPASSWD:ALL'), ('ssh-import-id', 'gh:herlo'), ('lock_passwd', True)])])])), ('count', 1)])])])])])'
+   errors:
+         res_defs[0][count]: value for field 'count' must be of type 'integer'
+         res_defs[0][networks][0][additional_storage]: field 'additional_storage' could not be recognized within the schema provided
+         res_defs[0][name]: field 'name' is required
+
+   topology for target 'libvirt' is valid under old schema
+   topology for target 'libvirt-network' is valid
+
+
+The `linchpin validate` command can also provide a list of errors against the
+old schema with the `--old-schema` floag
+
+.. code-block:: bash
+
+   $ linchpin validate --old-schema
+   
+   Topology for target 'libvirt-new' does not validate
+   topology: 'OrderedDict([('topology_name', 'libvirt-new'), ('resource_groups', [OrderedDict([('resource_group_name', 'libvirt-new'), ('resource_group_type', 'libvirt'), ('resource_definitions', [OrderedDict([('role', 'libvirt_node'), ('uri', 'qemu:///system'), ('image_src', 'http://cloud.centos.org/centos/7/images/CentOS-7-x86_64-GenericCloud-1608.qcow2.xz'), ('memory', 2048), ('vcpus', '1'), ('arch', 'x86_64'), ('ssh_key', 'libvirt'), ('networks', [OrderedDict([('name', 'default'), ('hello', 'world')])]), ('additional_storage', '10G'), ('cloud_config', OrderedDict([('users', [OrderedDict([('name', 'herlo'), ('gecos', 'Clint Savage'), ('groups', 'wheel'), ('sudo', 'ALL=(ALL) NOPASSWD:ALL'), ('ssh-import-id', 'gh:herlo'), ('lock_passwd', True)])])])), ('count', 1)])])])])])'
+   errors:
+         res_defs[0][networks][0][additional_storage]: field 'additional_storage' could not be recognized within the schema provided
+         res_defs[0][name]: field 'name' is required
+
+   topology for target 'libvirt' is valid under old schema
+   topology for target 'libvirt-network' is valid
+
+As you can see, validation under both schemas result in an error stating that
+the field `additional_storage` could not be recognized.  In this case, there is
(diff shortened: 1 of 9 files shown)


linchpin/error_handlers.py
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env python
```
herlo: The linchpin/exceptions namespace is where exceptions/errors go.

docs/source/examples/workspace/PinFile.complex-new.yml
```diff
@@ -0,0 +1,90 @@
+---
```
herlo: What is the purpose of creating a new pinfile? Can you put these into the existing ones? PinFiles here are generally for the purpose of testing and documentation. Adding/changing them should be to make more tests. Do we need a new pinfile for this functionality? Or can you use an existing one without modification?

docs/source/examples/workspace/PinFile.dummy.yml
```diff
@@ -3,6 +3,6 @@ dummy-new:
   topology: dummy-new.yml
   layout: dummy-new.yml
 
-dummy-topo:
-  topology: dummy-new.yml
-  layout: dummy.yml
+#dummy-topo:
+#  topology: dummy-new.yml
+#  layout: dummy.yml
```
herlo: I use this for the linchpin container builds, please don't remove it.

docs/source/examples/workspace/topologies/libvirt-new.yml
```diff
@@ -7,7 +7,7 @@ resource_groups:
       - role: libvirt_node
         name: centos71
         uri: qemu:///system
-        count: 1
+        count: 2
         image_src: http://cloud.centos.org/centos/7/images/CentOS-7-x86_64-GenericCloud-1608.qcow2.xz
         memory: 2048
```
herlo: Best to reduce the memory here to 1024 (unless there's a specific need for more)

linchpin/__init__.py
```diff
@@ -6,8 +6,12 @@
 import time
 import errno
 import hashlib
+import pprint
 
-from cerberus import Validator
+import error_handlers
```
herlo: See the comment about the namespace for exceptions/errors

linchpin/cli/__init__.py
```diff
@@ -432,7 +455,8 @@ def lp_destroy(self, targets=(), run_id=None, tx_id=None):
                                     tx_id=tx_id)
 
 
-    def _execute_action(self, action, targets=(), run_id=None, tx_id=None):
+    def _execute_action(self, action, targets=(), run_id=None, tx_id=None,
+                        old_schema=False):
```
herlo: What's the purpose of adding `old_schema` here? The util.dataparser does the conversion from the old schema to the new without indicating anything here. I wonder if maybe having the validator run *after* the conversion might be better?

linchpin/cli/__init__.py
```diff
@@ -603,7 +634,7 @@ def _build(self, pf, pf_data=None):
 
 
     def _execute(self, provision_data, targets,
-                 action='up', run_id=None, tx_id=None):
+                 action='up', run_id=None, tx_id=None, old_schema=False):
```
herlo: Same question as previous, why do we need to know if it's the old_schema?

linchpin/cli/__init__.py
```diff
@@ -405,6 +405,29 @@ def lp_up(self, targets=(), run_id=None, tx_id=None, inv_f="cfg"):
...
+        :param old_schema
+            Denotes whether schema should be validated with the old schema
+            rather than the new one!/usr/bin/env python
+        """
+
+        # Prep input data
+
+        # Execute prepped data
+        return_code, return_data = self._execute_action('validate', targets,
+                                                        old_schema=old_schema)
```
herlo: I see where you are going here, but I don't consider 'validate' an action. There are only two actions at the moment, 'up' and 'destroy'. Validation would be considered a component of either of these actions. Maybe we should figure out a better way to perform validation and not modify the execute_action function? I'd think we would want the validation to happen *before* executing an up or down. It is its own method also, so calling lp_validate here makes logically more sense.

docs/source/examples/workspace/topologies/libvirt-new.yml
```diff
@@ -7,7 +7,7 @@ resource_groups:
       - role: libvirt_node
         name: centos71
         uri: qemu:///system
-        count: 1
+        count: 2
```
herlo: Changing this count fundamentally changes the tests. If there's a good reason to change this, please explain it here. Thx

linchpin/shell/__init__.py
```diff
@@ -504,6 +504,38 @@ def journal(ctx, targets, fields, count, view,
...
+    try:
+        return_code, results = lpcli.lp_validate(targets=targets,
+                                                 old_schema=old_schema)
+        for target, result in results.iteritems():
+            if result == "valid":
+                result = "topology for target '{0}' is valid". format(target)
+            elif result == "valid with old schema":
+                result = "topology for target '{0}' is valid under old schema"\
+                    .format(target)
+            print(result)
```
herlo: please use log_state/log_info/log_debug here rather than print.

linchpin/validator.py
```diff
@@ -0,0 +1,53 @@
+from cerberus import Validator
```
herlo: I don't think this file should reside in the api root. It's likely going to land in the util namespace. What do you think of that?

docs/source/examples/workspace/PinFile.dummy.yml
14rcole: I just used this for testing and accidentally committed it.  I'll revert it

docs/source/examples/workspace/topologies/libvirt-new.yml
14rcole: Same thing as the one above

14rcole: This change was the same as the one above.

docs/source/examples/workspace/PinFile.complex-new.yml
14rcole: I initially needed to be able to demonstrate validation on a PinFile with only new topologies, but I don't think that's necessary anymore

linchpin/error_handlers.py
14rcole: Sounds good, I'll move it

linchpin/cli/__init__.py
14rcole: The old_schema flag needs to make its way down to the _validate_topology method.  If a document cannot be validated with either schema, this flag determines under which schema error will be reported

linchpin/cli/__init__.py
14rcole: That's fine with me.  Without this pr, validation occurs inside of either _execute(), which is called by _execute_action() so I'll need to decouple a bit more

docs/source/examples/workspace/topologies/libvirt-new.yml
14rcole: Nope, just another change that wasn't supposed to be pushed

linchpin/validator.py
14rcole: That's fine with me

linchpin/cli/__init__.py
herlo: Yeah, adding a call in each of those to your validation seems the right approach. We could definitely decouple more. There's a lot of complexity in those two functions, so it's not super straightforward and probably should be refactored anyway.

herlo:
Also, you have some flake8 errors you should clean up. You can see the travis-ci link below to find what to fixup. flake8 and testing tooling can be installed by performing `pip install -e .[tests]` or running pytest like in the travis-ci logs.

14rcole:
@herlo PTAL

linchpin/cli/__init__.py
```diff
@@ -405,6 +405,47 @@ def lp_up(self, targets=(), run_id=None, tx_id=None, inv_f="cfg"):
...
+
+        prov_data = OrderedDict()
+
+        if len(targets):
+            for target in targets:
+                prov_data[target] = provision_data.get(target)
+        else:
+            prov_data = provision_data
+
+        print "HELLO, WORLD"
```
herlo: I'm certain this isn't needed.

herlo:
@14rcole I think it's useful to have it there, yes. I was thinking the getting started guide, too. If you want to take the time to put these things in the docs, I'm all for that also. I just figured if we had some documentation somewhere, we could move it around in the future. I'm happy either way.

docs/source/advanced/validator.rst
```diff
@@ -0,0 +1,61 @@
+Validate Command
```
14rcole: @herlo how's this?

docs/source/advanced/validator.rst
```diff
@@ -0,0 +1,61 @@
...
+-----------------
+
+New in version 1.6.7
```
herlo: Coming in version 1.6.7? Haha! However, we should figure out how we're going to promise things like this that aren't in a release yet and may not make the release. Part of why I considered just putting it in as an 'Advanced Topic'. Let's discuss this detail more and come up with a plan.

docs/source/advanced/validator.rst
```diff
@@ -0,0 +1,61 @@
...
+   
+   Topology for target 'libvirt-new' does not validate
+   topology: 'OrderedDict([('topology_name', 'libvirt-new'), ('resource_groups', [OrderedDict([('resource_group_name', 'libvirt-new'), ('resource_group_type', 'libvirt'), ('resource_definitions', [OrderedDict([('role', 'libvirt_node'), ('uri', 'qemu:///system'), ('image_src', 'http://cloud.centos.org/centos/7/images/CentOS-7-x86_64-GenericCloud-1608.qcow2.xz'), ('memory', 2048), ('vcpus', '1'), ('arch', 'x86_64'), ('ssh_key', 'libvirt'), ('networks', [OrderedDict([('name', 'default'), ('hello', 'world')])]), ('additional_storage', '10G'), ('cloud_config', OrderedDict([('users', [OrderedDict([('name', 'herlo'), ('gecos', 'Clint Savage'), ('groups', 'wheel'), ('sudo', 'ALL=(ALL) NOPASSWD:ALL'), ('ssh-import-id', 'gh:herlo'), ('lock_passwd', True)])])])), ('count', 1)])])])])])'
+   errors:
+         res_defs[0][networks][0][hello]: field 'hello' could not be recognized within the schema provided
+         res_defs[0][name]: field 'name' is required
+
+As you can see, the error stating the 'count' must be an integer is gone, because the old schema makes no assumptions about the type of 'count'.
```
herlo: In both cases above, maybe show a simple topology which is being validated? it's also hard to  know what is right and how to correct the schema. Maybe even add an update to show how the schema (just one) is corrected and passes the validator?

docs/source/advanced/validator.rst
14rcole: That's fine with me, I was largely just copying other files at this point.  I can remove this line for now until we decide when this will become part of a release

14rcole:
@herlo is this good to merge?

herlo:
@14rcole It is definitely ready to go, but want to wait until the 1.6.0 release is out.

14rcole:
[merge]

PR #650, opened 2018-08-14
Add debug mode

14rcole:
Allows users to disable no_log in order to view debugging info at the expense of leaking secret keys

### DIFF
--- a/docs/source/examples/workspace/linchpin.conf
+++ b/docs/source/examples/workspace/linchpin.conf
@@ -111,6 +111,11 @@ distill_on_error = False
 # How long to wait before failing (in seconds) for an async task.
 #async_timeout = 1000
 
+# enables debug mode, which turns on debugging outputs for certain commands.
+# Note that this feature may cause secret keys to be leaked and therefore
+# should not be used in production.
+#debug_mode = False
+
 # the uhash value will still exist, but will not be added to
 # instances or the inventory_path
 #enable_uhash = False
--- a/linchpin/provision/roles/common/tasks/main.yml
+++ b/linchpin/provision/roles/common/tasks/main.yml
@@ -12,4 +12,3 @@
 - name: "output vars"
   set_fact:
     topology_outputs: {}
-
--- a/linchpin/provision/roles/openstack/tasks/provision_os_server.yml
+++ b/linchpin/provision/roles/openstack/tasks/provision_os_server.yml
@@ -89,7 +89,7 @@
   register: res_def_output
   loop: "{{ res_def_names }}"
   when: not async and auth_var is defined
-  no_log: true
+  no_log: "{{ not debug_mode }}"
   loop_control:
     loop_var: prov_inst
 
--- a/linchpin/linchpin.constants
+++ b/linchpin/linchpin.constants
@@ -33,6 +33,7 @@ _check_mode = False
 _async = False
 async_timeout = 1000
 default_ssh_key_path = ~/.ssh
+debug_mode = False
 
 
 # NOTE: if either of these change, the value in linchpin/templates must also change
--- a/linchpin/provision/roles/aws/tasks/provision_aws_cfn.yml
+++ b/linchpin/provision/roles/aws/tasks/provision_aws_cfn.yml
@@ -21,7 +21,7 @@
       Stack: "{{ res_def['res_name'] | default(res_def['name']) }}"
   register: res_def_output
   when: not async
-  no_log: true
+  no_log: "{{ not debug_mode }}"
 
 - name: "Provision cloud formation stack"
   cloudformation:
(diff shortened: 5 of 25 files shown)


herlo:
I'd like to merge them all. But I think we're going to get a 1.6.0b2 out today no matter what has been completed. Next week, we'll hopefully get the 1.6.0 out the door as well.

linchpin/provision/roles/openstack/tasks/provision_os_server-auto_ip.yml
```diff
@@ -0,0 +1,118 @@
...
+  poll: 0
+  register: res_def_output
+  when: async and auth_var is not defined
+
+# following tasks saves the async job details
+- name: "Async:: save the job id"
+  set_fact:
+   topology_outputs_os_server: "{{ topology_outputs_os_server + [ res_def_output ] }}"
+  when: async and res_def_output['skipped'] is not defined
+
```
herlo: Why is this new code here? Is this related to another PR that isn't merged into develop?

linchpin/provision/roles/openstack/tasks/provision_os_server.yml
```diff
@@ -197,8 +197,37 @@
...
+    floating_ip_pools: "{{ res_def['fip_pool'] | default(omit) }}"
+    security_groups: "{{ res_def['security_groups'] | default(omit) }}"
+    userdata: "{{ res_def['userdata'] | default(omit) }}"
+    count: "{{ res_def['count'] }}"
+    boot_from_volume: "{{ res_def['boot_from_volume'] | default(omit) }}"
+    terminate_volume: "{{ res_def['boot_from_volume'] | default(omit) }}"
+    volume_size: "{{ res_def['volume_size'] | default(omit) }}"
+    boot_volume: "{{ res_def['boot_volume'] | default(omit) }}"
+    volumes: "{{ res_def['volumes'] | default(omit) }}"
+>>>>>>> Add debug mode
```
herlo: You have a conflict here that needs to be resolved before it should be merged.

docs/source/examples/workspace/linchpin.conf
```diff
@@ -111,6 +111,11 @@ distill_on_error = False
 # How long to wait before failing (in seconds) for an async task.
 #async_timeout = 1000
 
+# enables debug mode, which turns on debugging outputs for certain commands.
+# Note that this feature may cause secret keys to be leaked and therefore
+# should not be used in production.
+# _debug_mode = False
```
herlo: For consistency, you'll notice the actual values (even though they are commented) do not have a space, while the descriptions do. Adjusting this will help.

linchpin/provision/roles/openstack/tasks/provision_os_server-auto_ip.yml
14rcole: It looks like I just made a mistake rebasing the PR.  I'll go back and fix that.

linchpin/provision/roles/common/tasks/main.yml
```diff
@@ -13,3 +13,7 @@
   set_fact:
     topology_outputs: {}
 
+- name: "assign debug_mode value"
+  set_fact:
+    debug_mode: "{{ _debug_mode }}"
```
14rcole: This is the reason for `debug_mode` and `_debug_mode`.  This is the same as `async` and `_async` on line 2

herlo: Oh, I see. You were just trying to confuse me by being clever. :)

herlo: @14rcole I wonder if we need to set the debug mode this way. If `_debug_mode` is what is used in the evars section of linchpin.constants/linchpin.conf, maybe we should be consistent in the code throughout? It's kind of unnecessary. It is in code, however, so if you don't want to change it, it's probably fine.

14rcole: I don't mind changing it, but should the `_async` option be changed as well?  Also, is there any point in the underscore if the variable isn't being renamed here?

herlo: The reason the `_async` has an underscore is because ansible already has `async` used and would conflict. LinchPin uses `async: "{{ _async }}" in plays. I am actually fine with taking out the underscore and just having `debug_mode` since you didn't use `_debug` which may have conflicted with something in ansible.

14rcole: I see.  I'll change that

herlo:
@14rcole can you rebase this against develop and if it passes, merge?

14rcole:
Will do

14rcole:
[test]

14rcole:
[merge]

PR #655, opened 2018-08-20
limiting pip to avoid dependency errors for pip 10+

skatlapa:


samvarankashyap:
@skatlapa : Could you comment the error you faced on when using pip >= 10 ? Further can you add more details why errors occur ?

skatlapa:
@samvarankashyap Hey, I did not face any error and also, I did not try LP anytime recently. But we faced related errors in Infrared last week.
My idea was to limit pip version as there have been errors related to the installation of virtualization pkgs and related packages such as pyyaml with the newer version of pip(10.0.1, not sure of 10.0.0 ..)

skatlapa:
Let's keep this on hold for now, until we see some related errors. I just wanted to introduce the idea of limiting pip version in the tool.

samvarankashyap:
Currently I am using  version 18 to install linchpin on my virtualenv . Didnot face any problem . I believe it is safe to assume pip version does not effect pypi installation . 
```
(pip_fix) [srallaba@laptopy pip_fix]$ pip --version
pip 18.0 from /home/srallaba/.virtualenvs/pip_fix/lib/python2.7/site-packages/pip (python 2.7)
```
If any were to arise in future this PR will be reopened .

skatlapa:
If the packages are already present earlier or installed via yum we might not face any errors. 

The best way to test would be to 
1) remove pyyaml, virt-install, (vbmc, etc..) via yum (and try the tool against packages in req.txt, e.g; PyYaml is one among common)
2) manually install packages with downgrading pip and try the tool with latest pip

Anyways, I'll try it later and re-open if I find any issues, as it is not blocking now, thanks.

Issue #649, opened 2018-08-12
Replace Auth_driver plain text credential management with ansible-vault


samvarankashyap:
Currently, linchpin accepts the credential folder as one of input which stores credentials in plain-text format which can be insecure . However , one can easily ansible-vault with single password more securely. 
Some parts of auth_driver module needs to be updated using helper package ansible-vault to achieve the same.
refer: [https://pypi.org/project/ansible-vault/1.0.4/](https://pypi.org/project/ansible-vault/1.0.4/)
refer: [https://stackoverflow.com/questions/44142208/how-to-view-decrypt-ansible-vault-credentials-files-from-within-a-python-script](https://stackoverflow.com/questions/44142208/how-to-view-decrypt-ansible-vault-credentials-files-from-within-a-python-script
)

PR #663, opened 2018-08-27
Implement auth_driver to read from vault

samvarankashyap:
$subject
- Now linchpin can read credentials from vault using vault_pass which can be read from 
1. environment variables
```
export VAULT_PASSWORD=testval
```
2. From command line 
```
linchpin -vvvv --creds-path ./credentials up --ask-vault-pass
```
3. Can be specified in linchpin.conf 
```
[evars]
....
vault_pass = testval
```
- In order to enable this feature one should turn on vault_encryption = True in linchpin.conf which is by default false
```
vault_encryption = True
```
fixes: #649


samvarankashyap:
[test]

herlo:
@samvarankashyap can you rebase this against develop?

herlo:
@samvarankashyap can you please rebase this against develop?

config/Dockerfiles/tests.d/dummy/03_dummy-creds-vault
```diff
@@ -0,0 +1,34 @@
...
+# providers.exclude: none
+
+DISTRO=${1}
+PROVIDER=${2}
+
+PINFILE="PinFile"
+TARGET="dummy_target"
+WORKSPACE_PATH="docs/source/examples/workspaces/dummy-creds-vault/"
+CREDS_PATH="docs/source/examples/workspaces/dummy-creds-vault/credentials/"
+export VAULT_PASSWORD=testval
```
herlo: The dummy provider doesn't use credentials, which seems like this test would not be useful. Additionally, the default location for testing credentials are actually stored in Jenkins variables. This credential itself should probably be stored in the contra-files repository, and included in some way where it can be expanded, rather than specifying the location here.

linchpin/__init__.py
```diff
@@ -547,7 +547,8 @@ def get_pf_data_from_rundb(self, targets, run_id=None, tx_id=None):
         return pinfile
 
 
-    def do_action(self, provision_data, action='up', run_id=None, tx_id=None):
+    def do_action(self, provision_data, action='up', run_id=None, tx_id=None,
+                  vault_pass=''):
```
herlo: This functionally changes the LinchPin api when it is not needed. A better way to approach this would be to use the context and pass it in similarly to the way `outfile` is passed [here](https://github.com/CentOS-PaaS-SIG/linchpin/blob/develop/linchpin/shell/__init__.py#L145).

linchpin/__init__.py
```diff
@@ -584,6 +585,13 @@ def do_action(self, provision_data, action='up', run_id=None, tx_id=None):
 
 
         return_code = 99
+        if vault_pass == '':
+            self.set_evar('vault_pass',
+                          self.ctx.get_cfg('evars',
+                                           'vault_pass',
+                                           default=''))
+        else:
+            self.set_evar('vault_pass', vault_pass)
```
herlo: `do_action` should take whatever the value is in `get_evar` since it would be set as described above. Also, no need to use `get_cfg` here, `get_evar` exists to pull evar values more easily.

linchpin/cli/__init__.py
```diff
@@ -374,7 +374,8 @@ def lp_down(self, pinfile, targets=(), run_id=None):
         pass
 
 
-    def lp_up(self, targets=(), run_id=None, tx_id=None, inv_f="cfg"):
+    def lp_up(self, targets=(), run_id=None, tx_id=None, inv_f="cfg",
+              vault_pass=''):
```
herlo: See the reference above for `do_action`. `vault_pass` should not need to be passed, use the context object instead.

linchpin/cli/__init__.py
```diff
@@ -471,7 +472,7 @@ def lp_validate(self, targets=(), old_schema=False):
         return self.do_validation(prov_data, old_schema=old_schema)
 
 
-    def lp_destroy(self, targets=(), run_id=None, tx_id=None):
+    def lp_destroy(self, targets=(), run_id=None, tx_id=None, vault_pass=''):
```
herlo: See the reference above for `do_action`. `vault_pass` should not need to be passed, use the context object instead.

linchpin/cli/__init__.py
```diff
@@ -495,14 +496,15 @@ def lp_destroy(self, targets=(), run_id=None, tx_id=None):
...
+                                       vault_pass=vault_pass)
         if ('post' in self.pb_hooks) and (self.__meta__ == "CLI"):
             self.hook_state = '{0}{1}'.format('post', 'destroy')
         return outputs
 
 
-
-    def _execute_action(self, action, targets=(), run_id=None, tx_id=None):
+    def _execute_action(self, action, targets=(), run_id=None, tx_id=None,
+                        vault_pass=''):
```
herlo: See the reference above for `do_action`. `vault_pass` should not need to be passed, use the context object instead.

linchpin/cli/__init__.py
```diff
@@ -674,7 +678,7 @@ def _build(self, pf, pf_data=None):
 
 
     def _execute(self, provision_data, targets,
-                 action='up', run_id=None, tx_id=None):
+                 action='up', run_id=None, tx_id=None, vault_pass=''):
```
herlo: As demonstrated, this variable is basically being passed around. The context variable is there for that purpose, use it instead. See the reference above for `do_action`.

linchpin/cli/__init__.py
```diff
@@ -703,7 +707,8 @@ def _execute(self, provision_data, targets,
         return self.do_action(prov_data,
                               action=action,
                               run_id=run_id,
-                              tx_id=tx_id)
+                              tx_id=tx_id,
+                              vault_pass=vault_pass)
```
herlo: See the reference above for `do_action`. `vault_pass` should not need to be passed, use the context object instead.

linchpin/provision/roles/dummy/tasks/provision_resource_group.yml
```diff
@@ -1,7 +1,38 @@
...
+    msg: "Auth var not set"
+  when: res_grp['credentials'] is defined and auth_var == ""
+
+- name: "provisioning resource definitions of current group"
+  include: provision_res_defs.yml res_def={{ res_item.0 }} res_grp_name={{ res_item.1 }}
+  with_nested:
+    - "{{ res_grp['res_defs'] | default(res_grp['resource_definitions']) }}"
+    - ["{{ res_grp['resource_group_name'] }}"]
+  loop_control:
+    loop_var: res_item
```
herlo: I would almost not want to see the credentials passed in the dummy provider. Yet, we did discuss this as a possibly useful thing in the future. However, this doesn't really authenticate against anything. Maybe it should actually validate against a simple file somewhere to prove it actually works.

linchpin/provision/roles/gcloud/tasks/provision_resource_group.yml
```diff
@@ -10,10 +10,12 @@
 
 - name: "Get creds from auth driver"
   auth_driver:
-    filename: "{{ res_grp['credentials']['filename']  }}"
-    cred_type: "gcloud"
+    filename: "{{ res_grp['credentials']['filename'] }}"
+    cred_type: "dummy"
```
herlo: This is the gcloud provider, but you are specifying the `dummy` cred type?

linchpin/provision/roles/openstack/tasks/provision_resource_group.yml
```diff
@@ -25,13 +25,16 @@
 
 - name: "Get creds from auth driver"
   auth_driver:
-    filename: "{{ cred_filename }}" # don't include the .yaml
+    filename: "{{ cred_filename }}"
```
herlo: Why remove the useful comment? Did the .yaml get added back to the cred_filename?

linchpin/shell/__init__.py
```diff
@@ -208,8 +208,9 @@ def init(ctx):
               cls=MutuallyExclusiveOption, mutually_exclusive=["run_id"])
 @click.option('-if', '--inventory-format', default="cfg",
               help="Inventory format can be cfg or json")
+@click.option('--ask-vault-pass', is_flag=True)
 @pass_context
-def up(ctx, targets, run_id, tx_id, inventory_format):
+def up(ctx, targets, run_id, tx_id, inventory_format, ask_vault_pass):
```
herlo: See the reference above for `do_action`. `vault_pass` should not need to be passed, use the context object instead.

linchpin/shell/__init__.py
```diff
@@ -267,8 +274,9 @@ def up(ctx, targets, run_id, tx_id, inventory_format):
 @click.option('-t', '--tx-id', metavar='tx_id', type=int,
               help='Destroy resources using the transaction ID (tx-id)',
               cls=MutuallyExclusiveOption, mutually_exclusive=["run_id"])
+@click.option('--ask-vault-pass', is_flag=True)
 @pass_context
-def destroy(ctx, targets, run_id, tx_id):
+def destroy(ctx, targets, run_id, tx_id, ask_vault_pass):
```
herlo: See the reference above for `do_action`. `vault_pass` should not need to be passed, use the context object instead.

herlo:
Also, put `--ask-vault-pass` as a generic option *before* the `up/destroy` action.

config/Dockerfiles/tests.d/dummy/03_dummy-creds-vault
samvarankashyap: @herlo : Dummy provider does not use credentials but the above changes simulates it use of credentials when mentioned in topology . The purpose of the current PR is to check whether the ansible vault is working or not . Further , The dummy credentials used are mock values I would not want them to mix up with the working credentials That is one of the reasons i chose to keep it within the workspace.

samvarankashyap:
@herlo changes made as per review suggestions

herlo:
[test]

config/Dockerfiles/tests.d/dummy/03_dummy-creds-vault
herlo: Ok, thanks for the explanation.

linchpin/shell/__init__.py
```diff
@@ -127,9 +127,11 @@ def _handle_results(ctx, results, return_code):
 @click.option('--creds-path', type=click.Path(), envvar='CREDS_PATH',
               help='Use the specified credentials path. Also works'
                    ' if CREDS_PATH environment variable is set')
+@click.option('--ask-vault-pass', is_flag=True, default=False,
+              help='Prompts for vault password')
```
herlo: :+1:

samvarankashyap:
[merge]

PR #672, opened 2018-08-30
Schema update to libvirt cloud_config

herlo:
The new validation structures require a dict for this schema instead of just a list. Cerberus 1.2 is a bit more strict about these things.

The rest of the updates were to match the [cloud_config documentation](https://cloudinit.readthedocs.io/en/latest/topics/examples.html#including-users-and-groups) for users and groups.


14rcole:
LGTM

herlo:
[merge]

PR #673, opened 2018-08-31
Libvirt: Read ip address from correct registered variable

joanbe:
When uri_hostname != localhost we should read the ip address from the
registered variable extract_ip_address_result_remote and not
extract_ip_address_result.


herlo:
[test]

linchpin/provision/roles/libvirt/tasks/provision_libvirt_node.yml
```diff
@@ -351,7 +351,7 @@
 
 - name: "Append ip_addresses and node_data to topology_outputs_libvirt_nodes"
```
herlo: Can you update the name to include '(remote)' or somesuch in it to differentiate from the play just above it?

herlo:
[merge]

Issue #677, opened 2018-09-04
linchpin validate fails when using 'type' instead of 'role'


herlo:
In the old version of the topology schema 'type' was an allowed value. When running linchpin validate against it, an error occurs.

```
$ linchpin -p /tmp/workspace/PinFile validate 
Traceback (most recent call last):
  File "/home/herlo/Projects/virtualenvs/v1.6.1-release/bin/linchpin", line 11, in <module>
    load_entry_point('linchpin', 'console_scripts', 'linchpin')()
  File "/home/herlo/Projects/virtualenvs/v1.6.1-release/lib/python2.7/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/home/herlo/Projects/virtualenvs/v1.6.1-release/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/home/herlo/Projects/virtualenvs/v1.6.1-release/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/herlo/Projects/virtualenvs/v1.6.1-release/lib/python2.7/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/herlo/Projects/virtualenvs/v1.6.1-release/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/home/herlo/Projects/virtualenvs/v1.6.1-release/lib/python2.7/site-packages/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args[1:], **kwargs)
  File "/home/herlo/Projects/virtualenvs/v1.6.1-release/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/home/herlo/Projects/centos/git/linch-pin/linchpin/shell/__init__.py", line 525, in validate
    old_schema=old_schema)
  File "/home/herlo/Projects/centos/git/linch-pin/linchpin/cli/__init__.py", line 452, in lp_validate
    return self.do_validation(prov_data, old_schema=old_schema)
  File "/home/herlo/Projects/centos/git/linch-pin/linchpin/__init__.py", line 822, in do_validation
    self._validate_topology(topology_data)
[...]

PR #680, opened 2018-09-04
Adjust validation so that 'type' does not create an error

14rcole:
Solves #677

### DIFF
--- a/linchpin/__init__.py
+++ b/linchpin/__init__.py
@@ -665,7 +665,7 @@ class LinchpinAPI(object):
             # if validation fails the first time, convert topo from old -> new
             try:
                 resources = self._validate_topology(topology_data)
-            except SchemaError:
+            except (SchemaError, KeyError):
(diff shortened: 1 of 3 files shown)


herlo:
When the topology validates against the 'old' schema, it says so. 

`topology for target 'simple' is valid under old schema`

When the topology validates against the 'new' schema, it doesn't say so.

I'm not sure this is possible, but is there a way to say something like `'type' is old, use 'role'`? Or maybe suggest updating? Thoughts?

14rcole:
We can do that.  Old schema is guaranteed to use 'type' instead of 'role', right?

14rcole:
I can do that. I actually think that instead of printing out the message about replacing directives for every topology that validates only under the old schema, we should have the message print out at the end (only if some topologies validate under the old schema).  It looks cleaner that way

14rcole:
[merge]

PR #689, opened 2018-09-13
added --no-hooks and --ignore-failed-hooks flags to commandline

samvarankashyap:
$subject 
--no-hooks ==> skips running hooks 
--run-hooks-on-failure ==> continues with hooks run even one fails . By default its false from linchpin.conf
usage:
```
linchpin -vvvv --creds-path ./credentials/ up --no-hooks

linchpin -vvvv --creds-path ./credentials/ destroy --no-hooks

linchpin -vvvv --creds-path ./credentials/ up --run-hooks-on-failure

linchpin -vvvv --creds-path ./credentials/ destroy --run-hooks-on-failure

```

Example: Workspace : https://github.com/samvarankashyap/linchpin_hooks_openstack_ws

### DIFF
--- a/linchpin/__init__.py
+++ b/linchpin/__init__.py
@@ -44,8 +44,6 @@ class LinchpinAPI(object):
 
         self.ctx = ctx
 
-        self.flags = {}
-
         self.hook_state = None
         self._hook_observers = []
         self.playbook_pre_states = self.get_cfg('playbook_pre_states',
@@ -549,8 +547,7 @@ class LinchpinAPI(object):
         return pinfile
 
 
-    def do_action(self, provision_data, action='up', run_id=None, tx_id=None,
-                  flags={}):
+    def do_action(self, provision_data, action='up', run_id=None, tx_id=None):
         """
         This function takes provision_data, and executes the given
         action for each target within the provision_data disctionary.
@@ -572,7 +569,6 @@ class LinchpinAPI(object):
                    to perform an idempotent reprovision, or destroy provisioned
                    resources.
         """
-        self.flags = flags
 
         ansible_console = False
         if self.ctx.cfgs.get('ansible'):
@@ -731,9 +727,7 @@ class LinchpinAPI(object):
             self.ctx.log_debug('calling: {0}{1}'.format('pre', action))
 
             if (('pre' in self.pb_hooks) and
-                not flags.get('no_hooks',
-                              ast.literal_eval(self.get_cfg('hooks',
-                                                            'no_hooks')))):
+                not self.get_cfg('hook_flags', 'no_hooks')):
                 self.hook_state = '{0}{1}'.format('pre', action)
 
             # FIXME need to add rundb data for hooks results
@@ -754,16 +748,12 @@ class LinchpinAPI(object):
 
             # add post provision hook for inventory generation
             if (('inv' in self.pb_hooks) and
-                not flags.get('no_hooks',
-                              ast.literal_eval(self.get_cfg('hooks',
-                                                            'no_hooks')))):
+                not self.get_cfg('hook_flags', 'no_hooks')):
                 self.hook_state = 'postinv'
 
             if (('post' in self.pb_hooks) and
                     (self.__meta__ == "API") and
-                    not flags.get('no_hooks',
-                                  ast.literal_eval(self.get_cfg('hooks',
-                                                                'no_hooks')))):
+                    not self.get_cfg('hook_flags', 'no_hooks')):
                 self.hook_state = '{0}{1}'.format('post', action)
 
             end = time.time()
--- a/linchpin/__init__.py
+++ b/linchpin/__init__.py
@@ -44,6 +44,8 @@ class LinchpinAPI(object):
 
         self.ctx = ctx
 
+        self.flags = {}
+
         self.hook_state = None
         self._hook_observers = []
         self.playbook_pre_states = self.get_cfg('playbook_pre_states',
@@ -547,7 +549,8 @@ class LinchpinAPI(object):
         return pinfile
 
 
-    def do_action(self, provision_data, action='up', run_id=None, tx_id=None):
+    def do_action(self, provision_data, action='up', run_id=None, tx_id=None,
+                  flags={}):
         """
         This function takes provision_data, and executes the given
         action for each target within the provision_data disctionary.
@@ -569,6 +572,7 @@ class LinchpinAPI(object):
                    to perform an idempotent reprovision, or destroy provisioned
                    resources.
         """
+        self.flags = flags
 
         ansible_console = False
         if self.ctx.cfgs.get('ansible'):
@@ -726,7 +730,10 @@ class LinchpinAPI(object):
             self.pb_hooks = self.get_cfg('hookstates', action)
             self.ctx.log_debug('calling: {0}{1}'.format('pre', action))
 
-            if 'pre' in self.pb_hooks:
+            if (('pre' in self.pb_hooks) and
+                not flags.get('no_hooks',
+                              ast.literal_eval(self.get_cfg('hooks',
+                                                            'no_hooks')))):
                 self.hook_state = '{0}{1}'.format('pre', action)
 
             # FIXME need to add rundb data for hooks results
@@ -746,10 +753,17 @@ class LinchpinAPI(object):
             # return an error code at the end.
 
             # add post provision hook for inventory generation
-            if 'inv' in self.pb_hooks:
+            if (('inv' in self.pb_hooks) and
+                not flags.get('no_hooks',
+                              ast.literal_eval(self.get_cfg('hooks',
+                                                            'no_hooks')))):
                 self.hook_state = 'postinv'
 
-            if ('post' in self.pb_hooks) and (self.__meta__ == "API"):
+            if (('post' in self.pb_hooks) and
+                    (self.__meta__ == "API") and
+                    not flags.get('no_hooks',
+                                  ast.literal_eval(self.get_cfg('hooks',
+                                                                'no_hooks')))):
                 self.hook_state = '{0}{1}'.format('post', action)
 
             end = time.time()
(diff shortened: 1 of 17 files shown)


linchpin/shell/__init__.py
```diff
@@ -208,8 +208,11 @@ def init(ctx):
               cls=MutuallyExclusiveOption, mutually_exclusive=["run_id"])
 @click.option('-if', '--inventory-format', default="cfg",
               help="Inventory format can be cfg or json")
+@click.option('-rhof', '--run-hooks-on-failure', is_flag=True, default=None)
+@click.option('-nh', '--no-hooks', is_flag=True, default=None)
```
herlo: Any long switches must have two dashes (eg. `--nh`)

linchpin/shell/__init__.py
```diff
@@ -208,8 +208,11 @@ def init(ctx):
...
 @pass_context
-def up(ctx, targets, run_id, tx_id, inventory_format):
+def up(ctx, targets, run_id, tx_id, inventory_format,
+       run_hooks_on_failure, no_hooks):
```
herlo: Can this flag not be obtained from the `ctx.get_evar()` or `ctx.get_cfg()` method? Why does this need to be passed?

linchpin/cli/__init__.py
```diff
@@ -355,7 +355,7 @@ def lp_down(self, pinfile, targets=(), run_id=None):
         pass
 
 
-    def lp_up(self, targets=(), run_id=None, tx_id=None, inv_f="cfg"):
+    def lp_up(self, targets=(), run_id=None, tx_id=None, inv_f="cfg", flags={}):
```
herlo: Why not use the context (self.ctx) here? It seems `flags` is just another thing to pass. The point of the ctx is that it is the ONE thing that is available in most (if not all) classes and should contain this data. In fact, the `inv_f` thing is likely something that could go through the ctx as well.

Plus, as I see above, the settings are already in the linchpin.constants/linchpin.conf. Adding this flag to the [runcli method](https://github.com/CentOS-PaaS-SIG/linchpin/blob/develop/linchpin/shell/__init__.py#L145) as shown in this link. Then you can just access the ctx values as necessary.

linchpin/__init__.py
```diff
@@ -547,7 +549,8 @@ def get_pf_data_from_rundb(self, targets, run_id=None, tx_id=None):
         return pinfile
 
 
-    def do_action(self, provision_data, action='up', run_id=None, tx_id=None):
+    def do_action(self, provision_data, action='up', run_id=None, tx_id=None,
+                  flags={}):
```
herlo: Why is flags necessary? See my other comments for details.

linchpin/__init__.py
```diff
@@ -569,6 +572,7 @@ def do_action(self, provision_data, action='up', run_id=None, tx_id=None):
                    to perform an idempotent reprovision, or destroy provisioned
                    resources.
         """
+        self.flags = flags
```
herlo: You could get the flags by doing something like `flags.append(ctx.get_cfg('hooks', 'no_hooks', default='False'))` for each of the flags, or have a way to load them from the config/cli.

linchpin/cli/__init__.py
```diff
@@ -476,14 +482,20 @@ def lp_destroy(self, targets=(), run_id=None, tx_id=None):
...
+                          ast.literal_eval(self.get_cfg('hooks',
+                                                        'no_hooks')))):
             self.hook_state = '{0}{1}'.format('post', 'destroy')
         return outputs
 
 
 
-    def _execute_action(self, action, targets=(), run_id=None, tx_id=None):
+    def _execute_action(self, action, targets=(), run_id=None, tx_id=None,
+                        flags={}):
```
herlo: Same comment as others, why use `flags` when `ctx` already exists?

linchpin/shell/__init__.py
```diff
@@ -227,11 +230,18 @@ def up(ctx, targets, run_id, tx_id, inventory_format):
     run-id:     Use the data from the provided run_id value
     """
 
+    flags = {}
+    if run_hooks_on_failure:
+        flags["run_hooks_on_failure"] = run_hooks_on_failure
+    if no_hooks:
+        flags["no_hooks"] = no_hooks
+
```
herlo: I kind of gave an example elsewhere of how to do this, but here's a simple way to get this data into context (or update it if it's changing) instead of using another variable (`flags` in this case).

`ctx.set_cfg('no_hooks', no_hooks)`
`self.ctx.set_cfg('run_hooks_on_failure', run_hooks_on_failure)`

When you need to create the flags object in the cli library, say, just get the data out there.

linchpin/shell/__init__.py
```diff
@@ -267,8 +278,10 @@ def up(ctx, targets, run_id, tx_id, inventory_format):
 @click.option('-t', '--tx-id', metavar='tx_id', type=int,
               help='Destroy resources using the transaction ID (tx-id)',
               cls=MutuallyExclusiveOption, mutually_exclusive=["run_id"])
+@click.option('--rhof', '--run-hooks-on-failure', is_flag=True)
+@click.option('-nh', '--no-hooks', is_flag=True)
```
herlo: Again, you need to have two dashes (`--nh`) to conform to the linchpin standard options.

linchpin/shell/__init__.py
```diff
@@ -282,11 +295,15 @@ def destroy(ctx, targets, run_id, tx_id):
     the appropriate PinFile will be destroyed.
 
     """
+    flags = {}
+    flags["run_hooks_on_failure"] = run_hooks_on_failure
+    flags["no_hooks"] = no_hooks
```
herlo: See the example in the up() method for how to handle this without passing more things around.

linchpin/cli/__init__.py
samvarankashyap: Totally agree with you . But I would not encourage adding flag to runcli method since the command structure might change with would effect usability of flags
``` 
linchpin -vvvv --creds-path --run-hooks-on-failure up 
```
The below looks more intutive and easy to remember since hooks happen after provisioning
```
linchpin -vvvv --creds-path up --run-hooks-on-failure
```
However, as you have suggested one can always set the context e_vars or cfg vars within the function. which can be done. And also avoids passing another arguments
Any thoughts ?

linchpin/__init__.py
samvarankashyap: @herlo My thoughts while implementation were there might be multiple flags coming up in future which might need to be accomodated . However, passing these as context variables seems much logical solution at this point.

linchpin/shell/__init__.py
herlo: @samvarankashyap  Actually, no they would not have to use what's in the config files. My point was exactly that you can modify the ctx to set the configs in the up/down methods which would use the flags from the cli. In turn, it would override the ones from linchpin.conf/linchpin.constants.

linchpin/cli/__init__.py
herlo: Good point. I may have misspoke about where to put the code. I do think in the up/down methods is more likely the better location. I'm happy with wherever you think is best, just that in the latter case, you'd have to add the options twice.

linchpin/shell/__init__.py
```diff
@@ -267,8 +275,10 @@ def up(ctx, targets, run_id, tx_id, inventory_format):
 @click.option('-t', '--tx-id', metavar='tx_id', type=int,
               help='Destroy resources using the transaction ID (tx-id)',
               cls=MutuallyExclusiveOption, mutually_exclusive=["run_id"])
+@click.option('--ifh', '--ignore-failed-hooks', is_flag=True)
+@click.option('--nh', '--no-hooks', is_flag=True)
```
herlo: These should have a help kwarg with a good short description.

linchpin/shell/__init__.py
```diff
@@ -208,8 +208,11 @@ def init(ctx):
               cls=MutuallyExclusiveOption, mutually_exclusive=["run_id"])
 @click.option('-if', '--inventory-format', default="cfg",
               help="Inventory format can be cfg or json")
+@click.option('--ifh', '--ignore-failed-hooks', is_flag=True, default=None)
+@click.option('--nh', '--no-hooks', is_flag=True, default=None)
```
herlo: These should have a help kwarg with a good short description.

samvarankashyap:
[test]

samvarankashyap:
[test]

samvarankashyap:
[test]

samvarankashyap:
[test]

samvarankashyap:
[test]

samvarankashyap:
[test]

samvarankashyap:
[test]

samvarankashyap:
[test]

samvarankashyap:
[test]

linchpin/shell/__init__.py
herlo: @samvarankashyap I updated your code with the `--nh` and `--ifh` and they work just fine. I'm not sure why they don't work for you. Please change this back.

linchpin/shell/__init__.py
herlo: @samvarankashyap I updated your code with the `--nh` and `--ifh` and they work just fine. I'm not sure why they don't work for you. Please change this back.

samvarankashyap:
@herlo it failed in build environment : https://jenkins-continuous-infra.apps.ci.centos.org/job/ci-linchpin/566/
when i change the notation to --

linchpin/shell/__init__.py
herlo: Still need `--help` option here

linchpin/shell/__init__.py
herlo: See the other comment I just made above. Same, but this one is a bit more specific.

samvarankashyap:
[merge]

PR #691, opened 2018-09-14
Remove print statement that caused AWS inventory gen tests to fail

14rcole:


14rcole:
@herlo if this passes tests, it should probably be pulled in for 1.6.2. Since it broke integration tests, it may also break someone else's script

14rcole:
[merge]

PR #696, opened 2018-09-18
Updated linchpin inventory generation to use relative/absolute paths

### LINKED ISSUE #617: inventory_file relational path can break linchpin
Setting `inventory_file` with relational path such as `beaker.yml`  will overwrite lincpin file `provision/beaker.yml`.

Linchpin version: 1.5.4
Example:
```
---
inventory_layout:
  inventory_file: beaker.yml
  vars:
    hostname: __IP__
    ansible_user: root
  hosts:
    example_shit:
      count: 1
      host_groups:
        - example
```
Running `linchpin up` will overwrite `~/.virtualenvs/linchpin/lib/python2.7/site-packages/linchpin/provision/beaker.yml` with the result inventory and thus breaking linchpin.

Expected result:
Workspace will be the relational path and `./beaker.yml` will be create or modified.

samvarankashyap:
Updated linchpin inventory generation to use relative/absolute paths if mentioned in layoutfile 
fixes #617 

Example workspace demonstrating the same:
https://github.com/samvarankashyap/linchpin_dummy_ws_demo_rel_path

### DIFF
--- a/linchpin/cli/__init__.py
+++ b/linchpin/cli/__init__.py
@@ -144,8 +144,10 @@ class LinchpinCli(LinchpinAPI):
                             # if its not absolute path make path
                             # relative to workspace
                             if not os.path.isabs(i_path):
-                                i_path = "{0}/{1}".format(self.ctx.workspace,
-                                                          i_path)
+                                i_path = "{0}/{1}/{2}".format(
+                                        self.ctx.get_cfg('evars','workspace'),
+                                        self.ctx.get_cfg('evars','inventories_folder'),
+                                        i_path)
                         else:
                             i_path = targets[name]["outputs"]
                             i_path = i_path["inventory_path"][0]
--- a/linchpin/cli/__init__.py
+++ b/linchpin/cli/__init__.py
@@ -11,6 +11,7 @@ import hashlib
 from random import randint
 from distutils import dir_util
 from collections import OrderedDict
+from jinja2 import Environment
 
 from linchpin import LinchpinAPI
 from linchpin.fetch import FETCH_CLASS
@@ -144,10 +145,13 @@ class LinchpinCli(LinchpinAPI):
                             # if its not absolute path make path
                             # relative to workspace
                             if not os.path.isabs(i_path):
-                                i_path = "{0}/{1}/{2}".format(
-                                        self.ctx.get_cfg('evars','workspace'),
-                                        self.ctx.get_cfg('evars','inventories_folder'),
-                                        i_path)
+                                # get default variable for inventories_path
+                                d_p = self.ctx.get_evar("default_inventories"
+                                                        "_path")
+                                d_p = Environment().from_string(d_p).render(
+                                    workspace=self.ctx.get_cfg('evars',
+                                                               'workspace'))
+                                i_path = "{0}/{1}".format(d_p, i_path)
                         else:
                             i_path = targets[name]["outputs"]
                             i_path = i_path["inventory_path"][0]
--- a/linchpin/cli/__init__.py
+++ b/linchpin/cli/__init__.py
@@ -138,7 +138,18 @@ class LinchpinCli(LinchpinAPI):
                         lt_data = targets[name]["inputs"]["layout_data"]
                         t_data = targets[name]["inputs"]["topology_data"]
                         layout = lt_data["inventory_layout"]
-                        i_path = targets[name]["outputs"]["inventory_path"][0]
+                        # check whether inventory_file is mentioned in layout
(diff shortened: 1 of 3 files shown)


linchpin/cli/__init__.py
```diff
@@ -138,7 +138,18 @@ def _write_to_inventory(self, tx_id=None, inv_path=None, inv_format="cfg"):
...
+                        if layout.get("inventory_file", None):
+                            i_path = layout["inventory_file"]
+                            # if its not absolute path make path
+                            # relative to workspace
+                            if not os.path.isabs(i_path):
+                                i_path = "{0}/{1}".format(self.ctx.workspace,
+                                                          i_path)
+                        else:
+                            i_path = targets[name]["outputs"]
+                            i_path = i_path["inventory_path"][0]
```
p3ck: Is this correct?  it looks to me like you meant to build upon the previous i_path definition.

samvarankashyap: @p3ck : Flake 8 tests were failing due to the length of line was exceeding 80 characters when I tried to accomodate everything into single line. I had to break it down into two lines as above to get the flake8 test pass.

p3ck: thanks for the explanation

linchpin/cli/__init__.py
```diff
@@ -138,7 +138,18 @@ def _write_to_inventory(self, tx_id=None, inv_path=None, inv_format="cfg"):
...
                         layout = lt_data["inventory_layout"]
-                        i_path = targets[name]["outputs"]["inventory_path"][0]
+                        # check whether inventory_file is mentioned in layout
+                        if layout.get("inventory_file", None):
+                            i_path = layout["inventory_file"]
+                            # if its not absolute path make path
+                            # relative to workspace
+                            if not os.path.isabs(i_path):
+                                i_path = "{0}/{1}".format(self.ctx.workspace,
+                                                          i_path)
```
herlo: I think we need to make this use the configuration settings in linchpin.conf, not just `workspace/inventory_path`.  See https://github.com/CentOS-PaaS-SIG/linchpin/blob/develop/linchpin/linchpin.constants#L54

herlo: @samvarankashyap Good catch on that one. Maybe we should scan the rest of the code and see where 'default_inventories_path' is used? I'm thinking it's not used much outside of the playbooks, if at all. We could adjust it to use a better path. 

BTW, the 'defaults' bit is probably wrong anyway. :smile:

linchpin/cli/__init__.py
```diff
@@ -144,8 +144,10 @@ def _write_to_inventory(self, tx_id=None, inv_path=None, inv_format="cfg"):
                             # if its not absolute path make path
                             # relative to workspace
                             if not os.path.isabs(i_path):
-                                i_path = "{0}/{1}".format(self.ctx.workspace,
-                                                          i_path)
+                                i_path = "{0}/{1}/{2}".format(
+                                        self.ctx.get_cfg('evars','workspace'),
+                                        self.ctx.get_cfg('evars','inventories_folder'),
+                                        i_path)
```
herlo: :+1:

herlo: I actually thought this would be the `default_inventory_path`. This does provide a similar location, but seems to have missed the point. A user would now have to know that the location defaults to the values of `<workspace>/<inventories_folder>`, rather than a single value of `<default_inventory_folder>`. Which should be used from the configuration?

samvarankashyap:
[test]

samvarankashyap:
[test]

linchpin/cli/__init__.py
```diff
@@ -138,7 +139,23 @@ def _write_to_inventory(self, tx_id=None, inv_path=None, inv_format="cfg"):
...
+                            # if its not absolute path make path
+                            # relative to workspace
+                            if not os.path.isabs(i_path):
+                                # get default variable for inventories_path
+                                d_p = self.ctx.get_evar("default_inventories"
+                                                        "_path")
+                                d_p = Environment().from_string(d_p).render(
+                                    workspace=self.ctx.get_cfg('evars',
+                                                               'workspace'))
+                                i_path = "{0}/{1}".format(d_p, i_path)
```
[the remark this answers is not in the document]
samvarankashyap: That's a good point in that case we need to have a system in place which renders whole config file with the variables available to us . Since its applicable for the all other config variables. In which case we move away from config parser interpolation to something else. 
We will be needed to device a solution for that .

samvarankashyap:
[merge]

PR #697, opened 2018-09-18
Make changes to InventoryFilters trigger Jenkins tests

14rcole:
Previously, changes to Inventory Generation would only trigger the default libvirt and dummy tests, which led to a few errors being missed in PRs.  This PR makes Jenkins check for changes to files in linchpin/InventoryFilters so that the relevant providers are tested if the files are changed


14rcole:
[merge]

PR #702, opened 2018-09-24
Allow users to configure inventory generation

14rcole:
Add a config section to PinFiles that allows users to define variables. These variables can then be used to customize the output of inventories


14rcole:
There are a number of files here (everything in docs/source/examples/workspace) that are listed as changed, but don't show up on my local branch when I run `git diff HEAD^ HEAD`.  Why is that?  Is there a way to reconcile the two lists?

PR #721, opened 2018-10-04
updated inventory_format flag to --if

### LINKED ISSUE #715: Inventory Format short option should have two dashes (--if vs -if)
As a standard for long options is always two dashes (even though it can be done otherwise), the (inventory format (`-if`) [option](https://github.com/CentOS-PaaS-SIG/linchpin/blob/develop/linchpin/shell/__init__.py#L209)  is incorrect. It should be corrected.

samvarankashyap:
fixes #715

### DIFF
--- a/linchpin/shell/__init__.py
+++ b/linchpin/shell/__init__.py
@@ -206,8 +206,8 @@ def init(ctx):
 @click.option('-t', '--tx-id', metavar='tx_id', type=int,
               help='Provision resources using the Transaction ID (tx-id)',
               cls=MutuallyExclusiveOption, mutually_exclusive=["run_id"])
-@click.option('-if', '--inventory-format', default="cfg",
-              help="Inventory format can be cfg or json")
+@click.option('--inventory-format', '--if', default="cfg",
+              metavar='INVENTORY_FORMAT', help="Inventory format cfg/json")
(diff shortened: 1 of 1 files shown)


linchpin/shell/__init__.py
```diff
@@ -206,7 +206,7 @@ def init(ctx):
 @click.option('-t', '--tx-id', metavar='tx_id', type=int,
               help='Provision resources using the Transaction ID (tx-id)',
               cls=MutuallyExclusiveOption, mutually_exclusive=["run_id"])
-@click.option('-if', '--inventory-format', default="cfg",
+@click.option('--inventory-format', '--if', default="cfg",
               help="Inventory format can be cfg or json")
```
herlo: Nit. For consistency sake, would it be better to have the shortcut (but long option) be after or before? I noticed that `--ifh` (and `--nh`) are the first of the two options. 

One thing that can be done, is use the third argument as the variable. See https://github.com/CentOS-PaaS-SIG/linchpin/blob/develop/linchpin/shell/__init__.py#L117. I like this option since it helps make things cleaner. Let me know what you think about that option.

samvarankashyap: Updated meta_var and & made the shortcut option first :+1:

samvarankashyap:
[merge]

PR #662, opened 2018-08-27
Add ability to validate layouts

14rcole:
Causes layouts to be validated when `linchpin up` is run.  As of now, layouts cannot be validated with `linchpin validate`, but I can add that to this PR if #647 is merged first, otherwise I'll submit a separate PR once that is merged

PR #722, opened 2018-10-04
Make validation functions public

14rcole:
Make _validate_topology and _validate_layout public so that they can be used via linchpin as an API.  Dependent of #662

### DIFF
--- a/linchpin/__init__.py
+++ b/linchpin/__init__.py
@@ -863,8 +863,7 @@ class LinchpinAPI(object):
                     self._convert_topology(topology_data)
                     self.validate_topology(topology_data)
                 except SchemaError as s:
-                    error = "[ERROR]"
-                    error += """Topology for target '{0}' does not validate
+                    error = """Topology for target '{0}' does not validate
 topology: '{1}'
 errors:
 """.format(target, topology_data)
--- a/linchpin/__init__.py
+++ b/linchpin/__init__.py
@@ -398,7 +398,7 @@ class LinchpinAPI(object):
         return layout_json
 
 
-    def _validate_topology(self, topology):
+    def validate_topology(self, topology):
         """
         Validate the provided topology against the schema
 
@@ -473,7 +473,7 @@ class LinchpinAPI(object):
             return msg
 
 
-    def _validate_layout(self, layout):
+    def validate_layout(self, layout):
         """
         Validate the provided layout against the schema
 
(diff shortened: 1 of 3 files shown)


14rcole:
@herlo I chose not to combine these functions because each function basically has three phases:

1. prepare inputs for validation
2. call shared `validate()` function
3. handle outputs

1 and 3 are pretty different for each function and 2 calls the same function for each.  Combining anything else would require the processing for 1 and 3 to be moved to the functions that call `_validate_topology()` or `_validate_layout()`, which seems to defeat the purpose

linchpin/__init__.py
```diff
@@ -688,6 +708,12 @@ def do_action(self, provision_data, action='up', run_id=None, tx_id=None):
 
             if provision_data[target].get('layout', None):
                 l_data = provision_data[target]['layout']
+                try:
+                    self._validate_layout(l_data)
```
herlo: should this be not using the `_`?

linchpin/__init__.py
```diff
@@ -847,11 +874,32 @@ def do_validation(self, provision_data, old_schema=False):
...
             else:
-                results[target] = "valid"
+                results[target] = "topology valid"
+
+            # validate layout
+            if provision_data[target].get('layout', None):
+                l_data = provision_data[target]['layout']
+
+                try:
+                    self._validate_layout(l_data)
```
herlo: should this be not using the `_`?

linchpin/__init__.py
14rcole: I accidentally branched off of my branch for #662 instead of develop.  Now that that's merged, I'll go back and clean things up

linchpin/__init__.py
14rcole: Same as above

14rcole:
[test]

herlo:
@14rcole A curious thing occurs when I run `linchpin validate`. I get `[ERROR] topology valid`, which seems counterintuitive. Any thoughts on this? Here's a gist to show the weirdness: https://gist.github.com/herlo/c6cea4c3a2f8a3a6289ec6e69b22f44f

14rcole:
@herlo that's fixed by #733

14rcole:
@herlo done

14rcole:
[merge]

PR #728, opened 2018-10-09
Make docs build in python3

### LINKED ISSUE #701: Docs do not build in Python 3
Attempting to "make html" in the docs directory fails. There are import path issues with linchpin in Python 3 that cause sphinx to fail at parsing the library.

14rcole:
Some of the code was not compatible with python 3.  Most of this was just updating import paths

Fixes #701


14rcole:
This should resolve #701

14rcole:
[merge]

PR #729, opened 2018-10-10
Remove duplicate task in dummy provider

samvarankashyap:
$subject
The above causes regression in integration tests


samvarankashyap:
[merge]

PR #730, opened 2018-10-11
Linchpin container release process - rework

herlo:
Continuous Delivery of the LinchPin Container. Depends on Jenkins multibranch pipelines with both branches and tags.

If a new change is made to the particular branch, develop in our case, a job will launch and build and test a container. If successful, it will be tagged 'latest'. If a tag is added to the branch, a job will launch and build the container also, upon success this will tag 'latest' and the version associated with the tag. (eg. v1.6.3 will get tagged '1.6.3'). In this way, there will be a 'development' container called 'latest', and regular release containers with proper versions.

### DIFF
--- a/config/Dockerfiles/JenkinsfileBuilder
+++ b/config/Dockerfiles/JenkinsfileBuilder
@@ -8,14 +8,6 @@
 
 import com.cloudbees.plugins.credentials.Credentials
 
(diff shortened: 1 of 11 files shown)


herlo:
I found a few bugs along the way that caused 'linchpin fetch' to fail. These fixes are in this PR also.

herlo:
[merge]

PR #733, opened 2018-10-12
Fix validate output on success

14rcole:
When linchpin validate successfully validates, it still puts [ERROR] at the beginning of the output.  Fix this


14rcole:
[test]

14rcole:
including this in #722

PR #734, opened 2018-10-15
Fix indentation of aws-sg-new target

14rcole:
This was causing some strange validation errors due to the indentation problems


14rcole:
[merge]

PR #735, opened 2018-10-15
Fix spelling errors in linchpin up output

14rcole:
If topology does not validate, the `linchpin up` output telling the user to run `linchpin validate` had some spelling errors


14rcole:
[merge]

PR #743, opened 2018-10-17
Make 'linchpin init' use 'linchpin fetch'

herlo:
For the purposes of documenting the new tutorial, I wanted to use linchpin init with some specific arguments. In this case, it was easier to modify `linchpin init` to allow it to use `lp_fetch()` to clone the git repository and use the workspaces that already exist there.

The `linchpin init` command will create a subdirectory, clone this repository (master branch by default) and copy the appropriate workspace to the subdirectory.

This PR allows a user to do the following:

```
[/tmp]$ linchpin init
Created destination workspace: /tmp/dummy

[/tmp]$ ls dummy
PinFile  PinFile.json  README.rst

[/tmp]$ cat dummy/PinFile
$ cat dummy/PinFile
---
dummy-new:
  topology:
    topology_name: "dummy_cluster" # topology name
    resource_groups:
        resource_group_type: "dummy"
        resource_definitions:
          - name: "{{ distro | default('') }}web"
            role: "dummy_node"
            count: 3
          - name: "{{ distro | default('') }}test"
            role: "dummy_node"
            count: 1
  layout:
   .. snip ..
```

If someone wants to use a different workspace from this repository, that is also possible:

```
$ linchpin init libvirt
Created destination workspace: /tmp/libvirt

$ cd /tmp/libvirt
$ ls
PinFile  README.rst
$ cat PinFile
---
libvirt-new:
  topology:
    topology_name: libvirt-new
    resource_groups:
      - resource_group_name: libvirt-new
        resource_group_type: libvirt
        resource_definitions:
          - role: libvirt_node
            name: centos71
            uri: qemu:///system
            count: 1
            memory: 2048
            vcpus: 1
            arch: x86_64
            ssh_key: libvirt
            networks:
              - name: default
            additional_storage: 10G
            cloud_config:
              users:
                - name: herlo
                  gecos: Clint Savage
                  groups: wheel
                  sudo: ALL=(ALL) NOPASSWD:ALL
                  ssh_import_id: gh:herlo
                  lock_passwd: true
  layout:
    inventory_layout:
      vars:
        hostname: __IP__
      hosts:
        example-node:
          count: 1
          host_groups:
            - example

.. snip ..
```

### DIFF
--- a/linchpin/shell/__init__.py
+++ b/linchpin/shell/__init__.py
@@ -183,20 +183,41 @@ def help(ctx):
 
 
 @runcli.command('init', short_help='Initializes a linchpin project.')
+@click.argument('provider', metavar='PROVIDER', required=False, nargs=1)
 @pass_context
-def init(ctx):
+def init(ctx, provider):
     """
     Initializes a linchpin project, which generates an example PinFile, and
     creates the necessary directory structure for topologies and layouts.
 
+    Utilizes lp_fetch with the following parameters:
+
+    remote:         git://github.com/CentOS-PaaS-SIG/linchpin
+    root:           workspaces
+    fetch_type:     workspace
+    fetch_protocol: FetchGit
+    fetch_ref:      master
+    dest_ws:        Unused, will use workspace/provider
+    nocache:        True
+
+
     """
 
-    # add a providers option someday
-    providers = None
+    remote = ctx.get_cfg('init', 'remote',
+                         default='git://github.com/CentOS-PaaS-SIG/linchpin')
+    root = ctx.get_cfg('init', 'root', default='workspaces/dummy')
+    fetch_type = ctx.get_cfg('init', 'fetch_type', default='workspace')
+    fetch_proto = 'FetchGit'
+    fetch_ref = ctx.get_cfg('init', 'fetch_ref', default='master')
+    nocache = ast.literal_eval(ctx.get_cfg('init', 'nocache', default='True'))
+
+    if provider:
+        root = 'workspaces/{0}'.format(provider)
 
     try:
-        # lpcli.lp_init(pf_w_path, targets) # TODO implement targets option
-        lpcli.lp_init(providers=providers)
+        lpcli.lp_fetch(remote, root=root, fetch_type=fetch_type,
+                       fetch_protocol=fetch_proto, fetch_ref=fetch_ref,
+                       dest_ws=None, nocache=nocache)
     except LinchpinError as e:
         ctx.log_state(e)
         sys.exit(1)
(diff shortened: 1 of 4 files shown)


samvarankashyap:
I love this concept !

samvarankashyap:
I have a question . 
So how do we handle the systems which do not have access to github . 
I am pretty sure we might face similar issue when provisioning duffy or internal openstack .

samvarankashyap:
Since everything is packaged with linchpin we can do a local fetch rather than a remote fetch .

herlo:
@samvarankashyap To answer your question about github access, I'd argue that if they wish, the user could pull down the git repo to a local place. The settings can be adjusted in the linchpin.conf to point to *any* git repository, including a local one.

herlo:
@samvarankashyap To address the remove vs local fetch question. The point of using git for `linchpin init` is because the 'local' functionality of linchpin fetch no longer exists/works. In terms of the amount of weight this has, pulling from github is a one time process, performed on the first fetch. After that, only fetch/pull is performed. 

I get the desire for a local init, however, the purpose is really just to introduce the basic concepts of linchpin in documentation. Maybe we can add that functionality in the future if it's desired.

14rcole:
is there currently an example in the docs for `linchpin fetch` from a local repo?

herlo:
@14rcole there are no examples for `linchpin fetch` that I know of, actually. There are vague references to it and the CLI help page. Here's the search I did in the documentation. https://linchpin.readthedocs.io/en/latest/search.html?q=fetch&check_keywords=yes

linchpin/shell/__init__.py
```diff
@@ -183,20 +183,41 @@ def help(ctx):
...
+    fetch_proto = 'FetchGit'
+    fetch_ref = ctx.get_cfg('init', 'fetch_ref', default='master')
+    nocache = ast.literal_eval(ctx.get_cfg('init', 'nocache', default='True'))
+
+    if provider:
+        root = 'workspaces/{0}'.format(provider)
 
     try:
-        # lpcli.lp_init(pf_w_path, targets) # TODO implement targets option
-        lpcli.lp_init(providers=providers)
```
samvarankashyap: @herlo : Everything looks good to me.
Just to maintain the backward compatibility for older linchpin init (using jinja )
what are your thoughts on introducing --local flag. when used uses older jinja templating for generating files ?

herlo: @14rcole yeah, I can see pulling an empty workspace (or rather, just making init create a directory with an empty PinFile should be simple enough. Let me add that to the code. In terms of not having internet access, I explained this previously above [here](https://github.com/CentOS-PaaS-SIG/linchpin/pull/743#issuecomment-431045428).

herlo:
[merge]

PR #744, opened 2018-10-19
Fix error with aws teardown

14rcole:
A variable in teardown_resource_group was being set to some_dict['content']['aws_ec2_res'][0], then later on, ansible would attempt to read variable['aws_ec2_res'].  I changed the variable to be equal to some_dict['content'], but if it's better practice to change the way the variable is being read later on, I can do that instead


14rcole:
@herlo PTAL.  This is blocking #732

14rcole:
[merge]

14rcole:
[merge]

14rcole:
[merge]

Issue #379, opened 2017-10-17
Multiple ansible hooks defined within the PinFile not executed


cmoulliard:
The following [pinfile](https://gist.github.com/cmoulliard/3f55ce132d2bdefb806088d6dbd8f02e#file-gistfile1-txt-L407-L425) 

```
---
os-single:
  topology: openstack-single.yml
  layout: openstack-single.yml
  hooks:
    postup:
      - name: test
         type: ansible
         context: True
         actions:
           - playbook: ping.yaml
             extra_vars: {"ansible_ssh_private_key_file": "/home/fedora/provision_vm_linchpin/my_key"}
      - name: osoos
         type: ansible
         context: True
         actions:
           - playbook: install_packages.yaml
           - playbook: git_clone.yaml
```

when executed doesn't show any error 

```
DEBUG 2017-10-17 12:17:12,126 from_api: True
DEBUG 2017-10-17 12:17:12,129 calling: preup
DEBUG 2017-10-17 12:17:12,129 hook preup initiated
DEBUG 2017-10-17 12:17:26,940 STATE - Action 'up' on Target 'os-single' is complete
DEBUG 2017-10-17 12:17:26,940 hook postup initiated
DEBUG 2017-10-17 12:17:26,941 running postup hooks
DEBUG 2017-10-17 12:17:26,941 STATE - -------
start hook ansible:test
DEBUG 2017-10-17 12:17:27,892 STATE - end hook ansible:test
-------
DEBUG 2017-10-17 12:17:27,892 STATE - -------
start hook ansible:osoos
DEBUG 2017-10-17 12:17:33,970 STATE - end hook ansible:osoos
-------

PLAY RECAP *********************************************************************************************************************************************************************************************************************
localhost                  : ok=30   changed=4    unreachable=0    failed=0

Action 'up' on Target 'os-single' is complete
hook postup initiated
running postup hooks
-------
start hook ansible:test

PLAY [osoos] *******************************************************************************************************************************************************************************************************************

TASK [Gathering Facts] *********************************************************************************************************************************************************************************************************
ok: [10.8.246.91]

TASK [ping] ********************************************************************************************************************************************************************************************************************
ok: [10.8.246.91]

PLAY RECAP *********************************************************************************************************************************************************************************************************************
10.8.246.91                : ok=2    changed=0    unreachable=0    failed=0

end hook ansible:test
-------
-------
start hook ansible:osoos
```

[...]

Issue #748, opened 2018-10-22
Hooks - ansible doesn't support mutliple playbooks


p3ck:
If supplying a hook and you have multiple playbooks to execute in one directory it will only execute the first one.

Please add a test case to the dummy provider to cover this.

samvarankashyap:
@p3ck My bad , The issue was already reported as a part of #379 
I think its still open

PR #752, opened 2018-10-23
Enables multi-file ansible hook in linchpin 

samvarankashyap:
fixes: #379 #748 
Summary:
Fixes Ansible hooks action_manager
Added integration test for multiple playbook runs for dummy provider
@p3ck


p3ck:
Looks good.  Can we change the hooks to create some file which you can check for?  That way we can be sure the hooks actually ran.

docs/source/examples/workspaces/dummy-multi-file-hook/linchpin.conf
```diff
@@ -0,0 +1,308 @@
...
+# The name of the state which occurs after (pre) teardown (destroy)
+#postdestroy = postdestroy
+
+# The name of the state which occurs after (post) inventory is generated (inv)
+#postinv = inventory
+#
+#[hookflags]
+#no_hooks = False
+#ignore_failed_hooks = False
+
```
herlo: Unless this file differs from the one in ../linchpin.conf, why not make it a symlink. It works fine with fetch and the testing afaik.

p3ck:
@samvarankashyap How do you know your second hook ran?  I want these tests to be able to verify that your hooks are running.  We shouldn't have to do manually review the tests to know if it worked.  Unless I'm missing something.

samvarankashyap:
@p3ck Updated the hook test create a file and verify it with a cat command. 
@herlo : Removed unnecessary  linchpin.conf

config/Dockerfiles/tests.d/dummy/05-dummy-multi-file-hook
```diff
@@ -0,0 +1,38 @@
...
+
+if [ -e /tmp/dummy.hosts ]; then
+    rm /tmp/dummy.hosts
+fi
+
+# testing no-hooks flag
+echo "Testing multi file hook"
+linchpin -w "${WORKSPACE_PATH}" -p "${PINFILE}" -v up "${TARGET}"
+
+cat /tmp/ex_hook.txt
```
p3ck: We're getting closer...  Can you actually check the output of the file to verify it contains the data that should have been written out by the second hook?  And if it doesn't exit non-zero.

samvarankashyap:
@p3ck Updated as per your suggestions

samvarankashyap:
[merge]

PR #753, opened 2018-10-24
Openshift examples for provisioning in openstack and beaker.

p3ck:
Looking for feedback on how I should structure the README.rst.

I would also like to add a make libvirt example as well which would be fully self contained.

### DIFF
--- a/docs/source/examples/workspaces/openshift-on-beaker/Makefile
+++ b/docs/source/examples/workspaces/openshift-on-beaker/Makefile
@@ -1,0 +1,14 @@
+credentials/beaker.conf:
+	@echo "No beaker.conf found! Please configure and try again"
+	exit 1
+
+hooks/ansible/openshift-ansible:
+	pushd hooks/ansible && \
+		git clone -b release-3.9 https://github.com/openshift/openshift-ansible
+
+openshift-on-beaker: hooks/ansible/openshift-ansible credentials/beaker.conf
+	BEAKER_CONF=$(PWD)/credentials/beaker.conf linchpin -v up openshift-on-beaker
+
+destroy-openshift-on-beaker: credentials/beaker.conf
+	BEAKER_CONF=$(PWD)/credentials/beaker.conf linchpin -v \
+                   destroy openshift-on-beaker
--- a/docs/source/examples/workspaces/openshift-on-beaker/PinFile
+++ b/docs/source/examples/workspaces/openshift-on-beaker/PinFile
@@ -1,0 +1,21 @@
+---
+openshift-on-beaker:
+  topology: openshift-on-beaker.yml
+  layout: openshift.yml
+  hooks:
+    postup:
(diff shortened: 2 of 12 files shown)


14rcole:
The README.rst seems pretty good to me.  I would link to the hooks documentation if possible

docs/source/examples/workspaces/openshift/hooks/ansible/openshift-install-rpmrepo/install_openshift_repo.yml
```diff
@@ -0,0 +1,33 @@
...
+                host: "{{ inventory_hostname }}"
+                search_regex: OpenSSH
+                port: 22
+                delay: 65
+                timeout: 900
+            delegate_to: localhost
+          - name: Install openshift
+            yum:
+                    name:
+                            - centos-release-openshift-origin39
```
samvarankashyap: Indentation little off at these lines

docs/source/examples/workspaces/openshift/hooks/ansible/openshift-install-rpmrepo/install_openshift_repo.yml
```diff
@@ -0,0 +1,33 @@
...
+                            - NetworkManager
+                    state: installed
+          - name: Enable NetworkManager
+            service:
+                    name: NetworkManager
```
samvarankashyap: Indentation little off at these lines

docs/source/examples/workspaces/openshift/PinFile
```diff
@@ -1,39 +1,42 @@
 ---
-openshift-new:
```
samvarankashyap: @p3ck I am not sure why this Pinfile and workspace is modified . 
The intention of older Pinfile is to use oepnshift as a resource provider . 
According to that logic this workspace should go under openstack folder . 
@p3ck @herlo Any thoughts ???

herlo: @samvarankashyap the workspaces bits are not being used in documentation yet, but that would be the intent eventually. 

@p3ck I do somewhat agree with @samvarankashyap here that this workspace should maybe be its own, rather than lumping it in with another. Thoughts?

p3ck:
Yeah - I agree that this is the wrong directory for this.  I wonder if I should break this into an example to be placed in the openstack directory and the beaker directory?  The alternative is to create an openshift-deployment dir?

herlo:
The directories don't have to specifically be for a provider. Rather, a use-case version would seem appropriate, yes? If it were me, I'd call the workspace something like 'openshift-cluster' or 'mp-openshift-cluster' (mp stands for multi-provider).

samvarankashyap:
[test]

samvarankashyap:
@p3ck needs rebasing !

p3ck:
[merge]

PR #703, opened 2018-09-24
Add variables in inventory generation

14rcole:
Allow users to define variables in `cfgs` section of the PinFile, which can be used to customize inventory output

PR #756, opened 2018-10-25
Adjust code style to reflect new flake8 rules

14rcole:
I originally included this in #732 but it appears its also causing #703 to fail (and potentially others) so I thought it would be wise to split it out


14rcole:
[merge]

Issue #757, opened 2018-10-25
LinchPin Tutorial - SPIKE


herlo:
Create Your Own Adventure Style tutorial

* Introduction (Getting Started)
* Tutorial Landing Page (with a selection of providers, features, etc) for the user to choose.
* Coverage for each Provider and Feature
  * Dummy
  * Openstack
  * AWS
  * Libvirt
  * GCE
  * Beaker
  * Duffy
  * Openshift
  * Ovirt**
  * Rackspace**
  * Inventory Generation
  * Hooks
  * Journal
  * Distiller / External Context / Linchpin Latest

** - If desired as no known demand exists

maneeshmehra:
Remaining work per following JIRA items that were recently scrubbed:

1. Tutorial for Dummy (https://projects.engineering.redhat.com/browse/CONTRA-673)
2. Tutorial for Duffy (https://projects.engineering.redhat.com/browse/CONTRA-672)
3. Tutorial for OpenShift (https://projects.engineering.redhat.com/browse/CONTRA-671)
4. Tutorial for GCE (https://projects.engineering.redhat.com/browse/CONTRA-669)

samvarankashyap:
Few of the video tutorials are added to the linchpin project's GitHub wiki. 
If needed more separate issues can be created.

PR #758, opened 2018-10-25
Initial getting started tutorial

herlo:
This is the introduction bit of the Create-Your-Own Adventure Tutorial. Currently, it covers the basics of the `simple` workspace. It also uses `linchpin init`, (see #743) to generate the workspace in the tutorial. 

Addresses the `getting started` part of #757

It may be useful to checkout this PR, and run `make html` from the docs directory to view the updated documentation.


docs/source/simple_pinfile.rst
```diff
@@ -0,0 +1,56 @@
...
+
+A :term:`PinFile` takes a :term:`topology` and an optional :term:`layout`, among other options, as a combined set of configurations as a resource for provisioning. An example :term:`Pinfile` is shown.
+
+The *simple* PinFile is shown below
+
+.. code-block:: yaml
+
+    1   ---
+    2   simple:
+    3   topology:
```
samvarankashyap: I believe the indentation is off at this line after the target

14rcole:
I pulled and built the docs, they're looking great so far!

docs/source/simple_up.rst
```diff
@@ -0,0 +1,102 @@
+Up
+--
+
+It's time to provision your first LinchPin resources.
+
+.. code-block:: bash
+
+    1   [/tmp/simple]$ linchpin up
```
samvarankashyap: @herlo 
linchpin up block here has line numbers , 
but linchpin destroy does not . 
To maintain consistency could you please remove the line numbers here ! or add line numbers to linchpin destroy

herlo: Actually, the purpose for the line numbers is to explain what things are in the components. I'm not sure I need to add line numbers to linchpin destroy since everything that is explained is covered in the `linchpin up` section. If you render the documentation, you'll see superscript numbers that correspond to the line numbers. This is what the `:sup:8` is for [here](https://github.com/CentOS-PaaS-SIG/linchpin/pull/758/files#diff-05d8e36df29d8bddd5336fe0290917c1R39).

Can you tell me why it would be needed for the other examples on this page?

samvarankashyap: It might be needed if we would like to refer . Since for dummy provider destroy is self explanatory it not needed as you explained . Approving it now .

herlo:
[merge]

Issue #419, opened 2017-12-05
Detect Missing Extras When Possible


jaypoulz:
# User Story
Currently, installing linchpin for using beaker (and potentially ovirt), is less than intuitive.

For providers like OpenStack, you simply install the program like so:
```
pip install linchpin
```
Following the docs, you might copy-paste a dummy topology file and start provisioning from there.

Now, say you wanted to change that to beaker. Now that the topology layouts are the same, it's even easier to fall into this trap, since it's much easier to assume that just modifying the topology file will provision the resources as expected.

But they don't - attempting to provision here leads you to:
```
ImportError: No module named bkr.client.task_watcher
```
To solve this, you have to either look it up in the `docs` or crack open the `setup.py`.
There you'd find that there's actually an extra package for `linchpin[beaker]`.

After a quick
```
pip install linchpin[beaker]
```

You might try your provision command again, expecting it to work. Of course, now the issue is that you're missing the `linchpin[krbV]` extras.

# Problem Context
The underlying issue here is that linchpin has an internal struggle going on. It's currently being sold as a 'one-size-fits-all' tool for your cloud provisioning needs. That is, it should be able to provision anything it supports out of the box. On the other hand, there's a case to be made for why you'd want to keep individual providers external to the linchpin core - to reduce the package size and number of dependencies.

This conflict is understandable but creates a challenge for user experience.

# Proposed Solutions
1. All providers should work out of the box.
This is ideal for usability. Besides the package size concerns, this guarantees that once I have linchpin installed, it will work as expected.

2. Install provider dependencies them as you need them
Since topologies control which providers you might be using, installing the modules dependencies you need at runtime should be as easy as scanning through the topology being called. This could be a role called off the main playbook, which could be implemented by each provider for the dependencies they need.

## Minimum Viable Solution
At the very least, the user should get feedback for calls made when missing extras. The problem with some of the current extras is that they're expected to be part of the main package. So if linchpin fails on missing beaker or krbV imports, it should suggest as part of its output that you install the missing extras.

PR #759, opened 2018-10-29
Linchpin setup and Adding error handling messages beaker and libvirt

samvarankashyap:
Introducing linchpin setup command which are used to automate linchpin external dependency installations 

command: 
```
linchpin setup [dependency_list]
```
example usage: 
```
# by default linchpin setup installs all dependencies 
# test , docs , libvirt, beaker, selinux etc 
linchpin setup
```
```
# one can always specify set of dependencies
linchpin setup beaker libvirt
```

Further current PR adds debug fail messages to beaker and libvirt when something goes wrong with the provisioning process. It works both in virtualenv.
Some of the dependencies might need  sudo permissions which can either be done using 
```
linchpin setup --ask-sudo-pass 
```
or running linchpin with sudo 
```
sudo linchpin setup
```
or running linchpin under a privileged sudo user 
```
linchpin setup
```

fixes: #419

### DIFF
--- a/linchpin/provision/roles/dependencies/tasks/libvirt_deps.yml
+++ b/linchpin/provision/roles/dependencies/tasks/libvirt_deps.yml
@@ -1,0 +1,25 @@
+- name: Install dependencies
+  block:
+    - name: Install package dependencies
+      package:
+        name: "{{ libvirt_pkg }}"
+        state: latest
+      with_items:
+      - python2-dnf 
+      - libvirt-devel
+      - libguestfs-tools 
+      - python-libguestfs
+      become: true
+      loop_control: 
+        loop_var: libvirt_pkg
+    - name: Install pypi dependencies of libvirt
+      pip:
+        name: "{{ libvirt_pypi }}"
+      with_items:
+      - "libvirt-python>=3.0.0"
+      - "lxml"
+      loop_control: 
+        loop_var: libvirt_pypi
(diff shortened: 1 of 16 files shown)


linchpin/provision/roles/dependencies/tasks/libvirt_deps.yml
```diff
@@ -0,0 +1,25 @@
+- name: Install dependencies
+  block:
+    - name: Install package dependencies
+      package:
```
p3ck: Please see https://github.com/ansible/ansible/issues/24581

You should not use with_items with package module.  Pass the list directly to name: in package module.

14rcole: @samvarankashyap fyi this appears in a few places

samvarankashyap:
@14rcole : The reasoning behind the merging error handling and linchpin setup in the current PR is the error messages are suggestion to use linchpin setup before running linchpin provision.
I will try to add the documentation, However I could not find appropriate place for it in docs . Most probably I will try to write something in installation docs as an additional section how to use linchpin setup to install dependencies. 
https://linchpin.readthedocs.io/en/latest/installation.html

samvarankashyap:
@14rcole ??

samvarankashyap:
[merge]

PR #760, opened 2018-10-30
Fix Typos in Getting Started Docs

herlo:
An update to the new getting started guide. Fixes some typos, adds a couple of more examples and clarifies a few things.

After checking it out, go to the `docs` directory and run `make html`. Then open the `docs/build/html/index.html` in a browser and view the 'Getting Started` section.

PR #767, opened 2018-10-31
bugfix: Update target name integration tests w.r.to changes in examples

samvarankashyap:
$subject
Fixes error in the integration tests


samvarankashyap:
NVM, I think this is fixed in #760

Issue #771, opened 2018-10-31
If namespace is not mentioned template linchpin openshift raises an error .


samvarankashyap:
$subject

skatlapa:
just adding in @robnester-rh

PR #772, opened 2018-10-31
bugfix: Set namespace default to null string 

samvarankashyap:
Sets the namespace variable to "" when None
fixes : #771


skatlapa:
Let's wait on @herlo too

skatlapa:
same, lgtm, we could merge.

samvarankashyap:
[merge]

PR #775, opened 2018-11-05
Add inventory generation integration test

14rcole:
Add integration test to inventory generation to make sure that host ordering and variables work as expected

### DIFF
--- a/config/Dockerfiles/tests.d/inventory/02_inventory_valid
+++ b/config/Dockerfiles/tests.d/inventory/02_inventory_valid
@@ -1,0 +1,90 @@
+#!/bin/bash -xe
+
+# Verify template-based provisioning using complex template data file
+# distros.exclude: none
+# providers.include: aws
+# providers.exclude: none
+
+DISTRO=${1}
+PROVIDER=${2}
+
+PINFILE="PinFile"
+TARGET="dummy-aws"
+CONFIG_FILE="../linchpin.conf"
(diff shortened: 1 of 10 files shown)


config/Dockerfiles/tests.d/inventory/02_inventory_valid
```diff
@@ -0,0 +1,90 @@
...
+uhash=$(tail -n 1 ${TMP_FILE} | awk '{ print $3 }' )
+echo ${uhash}
+
+INVENTORY_FILE="./inventories/${TARGET}-${uhash}.inventory"
+ls -l ${INVENTORY_FILE}
+grep ${VAL2CHECK} ${INVENTORY_FILE} | grep ${VALUE}
+RC1=${?}
+
+verify_sections $INVENTORY_FILE
+
```
p3ck: I like that you are adding verification that we are generating the correct inventory.  But I think the way this is implemented might be overly complicated and hard to maintain.

The beaker code has similar tests which verify that they generate a kickstart file correctly and they simply have an expected output of the content and compare that to what they just generated pragmatically.  They then can show a diff between the versions if it doesn't match.  Makes it easy to update the test if needed.

Can you look here:
https://gerrit.beaker-project.org/c/beaker/+/6191/7/IntegrationTests/src/bkr/inttest/server/kickstarts/Fedorarawhide-scheduler-defaults.expected

I can go into more depth if needed.  Thanks!

PR #776, opened 2018-11-05
Make linchpin validate print both layout and topology success

14rcole:
Code typo made linchpin validate print only "topology valid" if both topology and layout for a given host were valid.  This fixes that bug


14rcole:
[merge]

PR #782, opened 2018-11-06
Update reqs to control ansible version

skatlapa:


skatlapa:
Please hold on merging this, as we might need 2.7 if we were to use the kube module for openshift provider https://docs.ansible.com/ansible/latest/modules/k8s_module.html?highlight=kubernetes

Issue #769, opened 2018-10-31
Inventories do not order hosts correctly


14rcole:
linchpin should collect all provisioned hosts, sort them by the order in which they're declared in the topology, and pop them off one by one according to the structure of the layout. This isn't currently happening.  When I tried to provision with [this](https://github.com/14rcole/inventory-order) PinFile, I get the inventory that appears in that repo.  the ec2 host appears in every section except the one in which it is supposed to and the host in [dummy-master] also appears in [dummy-layout]

14rcole:
1. Adding custom host groups messes with the ordering of hosts.  See https://github.com/14rcole/inventory-order/tree/correct-aws-node

14rcole:
2. Hosts *should* be parsed by the layout in the same order in which they're provisioned.  However, in [this topology](https://github.com/14rcole/inventory-order/tree/multiple-dummy), `dummy-topology` and `dummy-two` are grouped together, and `dummy-two` is parsed before `aws-topology`

14rcole:
3. Despite ordering being mostly fixed in https://github.com/14rcole/inventory-order/tree/correct-aws-node, the node in the dummy-master section is still double-dipping in the dummy-node section

14rcole:
> 3. Despite ordering being mostly fixed in https://github.com/14rcole/inventory-order/tree/correct-aws-node, the node in the dummy-master section is still double-dipping in the dummy-node section

As it turns out, this was not an error.  The hosts in the inventory are organized by *host group*, not by *host*.  I didn't realize that the `dummy-master` host was a member of both `dummy-layout` and `dummy-master` host groups

14rcole:
@samvarankashyap 1 and 3 are fixed. This was left open due to 2, which I wanted to address during the linchpin devel meeting

PR #784, opened 2018-11-06
Test inventory order in a unit test

14rcole:


p3ck:
Oops, I spoke too soon.  If you dig into the failure it's because it can't open a file.  getting close though.

14rcole:
[test]

p3ck:
@14rcole it passed the unit test, but the code hasn't been fixed yet right?
        "linchpin/tests/InventoryFilters/test_GenericInventory_pass.py::test_output_order PASSED [ 49%]",

linchpin/tests/InventoryFilters/test_GenericInventory_pass.py
```diff
@@ -5,6 +5,8 @@
 import os
 import json
 import yaml
+import json
```
samvarankashyap: @14rcole  Duplicate import

14rcole:
@p3ck I only added the test.  The code seems to be working, at least this is the expected output here.

One thing worth questioning is how similar do these topologies need to be?  I compared the files as if they needed to be identical, but really, each section just needs to match.  The sections themselves could be reordered and be fine, right?  If so, I should modify the verification to read in the expected inventory to the ConfigParser and compare sections from there

samvarankashyap:
> @p3ck I only added the test. The code seems to be working, at least this is the expected output here.
> 
> One thing worth questioning is how similar do these topologies need to be? I compared the files as if they needed to be identical, but really, each section just needs to match. The sections themselves could be reordered and be fine, right? If so, I should modify the verification to read in the expected inventory to the ConfigParser and compare sections from there

@14rcole A simpler test would be a pattern match on a file using regex for sections, instances should output the identical sections every time its ran.

p3ck:
My suggestion would be to take the example topology here and run it for real.  See if you get the same results in the real world.

Also - Don't we have a reproducer from issue from #769 ?

I really want to see the problem reproduced with this test case.

14rcole:
#769 mysteriously stopped reproducing.  I don't know if another PR accidentally fixed it. I left the issue open because of [this](https://github.com/CentOS-PaaS-SIG/linchpin/issues/769#issuecomment-434792907), which I meant to bring up at the linchpin devel meeting or not.  I'm not sure if that needs to be fixed.  Other than that, the example topology did produce the expected inventory (albeit with the sections in a different order - I had to reorder the expected output to match)

p3ck:
ok - I agree that the order of the hosts in each section shouldn't matter, as long as they end up in the correct sections.

@samvarankashyap Can you think of any other combinations that could be a problem?  I know in the beaker provider I can provide multiple recipesets:

yaml```
        recipesets:
          - distro: RHEL-7.4
            arch: x86_64
            variant: Server
            count: 1
            name: rhel75-1
            keyvalue:
              - DISKSPACE>=100000
            bkr_data: lvm-on
            hostrequires:
              - tag: arch
                op: "="
                value: "X86_64"
              - tag: memory
                op: ">="
                value: "1000"
          - distro: RHEL-7.5
            arch: x86_64
            variant: Server
            count: 1
            name: rhel75-2
            keyvalue:
              - DISKSPACE>=100000
```

Can we add this use case into the test?

14rcole:
I added this to the test.  As it turns out, we do need to compare the files verbatim. ConfigParser can't read the inventory files it's generating because the [host_group:children] sections have keys without values

p3ck:
@14rcole did you push the latest version of this patch?  You stated you added my example to the test but I don't see it.

14rcole:
@p3ck I just pushed up the most recent changes.  I spent a while wrestling with beaker

14rcole:
[merge]

PR #789, opened 2018-11-09
Add integration test to check linchpin.latest output

14rcole:
Check linchpin.latest output to make sure that the data we receive from the various providers is up to date


14rcole:
One thing to consider with this test is should we be running it with every relevant PR, or just the release PRs?

14rcole:
[test]

p3ck:
@14rcole #792 has been merged.  Remember to update providers.include to 
providers.include: __RELEASE__

PR #792, opened 2018-11-13
Allow for easy release testing

p3ck:
Introduce __RELEASE__ provider which will match on linchpin/version.py
change.  dummy still matches everything and all drivers will still
match if linchpin/version.py changes.  This just makes it so you
can do testing *ONLY* when it's a release.


p3ck:
[merge]

PR #793, opened 2018-11-13
Add linchpin.latest files to Jenkins artifacts

14rcole:


14rcole:
I don't know either, I guess we'll find out

p3ck:
@14rcole looks like it worked!  you get lots of latest. :-)

Issue #798, opened 2018-11-19
ansible functions should have unit tests


14rcole:
All of the functions in `linchpin/provision/filter_plugins` and `linchpin/provision/library` need unit tests

PR #800, opened 2018-11-20
Fix combination beaker names and hosts

14rcole:
Previously, each name was combined with each host.  This fixes the function so that each name is paired with the corresponding host


samvarankashyap:
@14rcole Could you add sample unit test with mock data to check if works or not ?

14rcole:
@p3ck @samvarankashyap ptal

14rcole:
@samvarankashyap sorry, I didn't see your comment.  I can.  This would fall under #798 though, should I just put the unit test in there with the rest or slip it in here?

samvarankashyap:
You can group it with others

14rcole:
[merge]

PR #808, opened 2018-11-28
Added wait timeout parameter to aws_ec2 module

samvarankashyap:
Add wait_timeout parameter aws resource_def to increase timeout period for tests.
Further, change the default value to 600 instead of 300
ref : https://docs.ansible.com/ansible/2.6/modules/ec2_module.html


samvarankashyap:
[merge]

PR #817, opened 2018-12-05
Fail on provider playbook failure while provisioning

samvarankashyap:
$subject


samvarankashyap:
@p3ck Do we want to log it to console ?? when destroy fails, if thats the case linchpin already does the reporting.

samvarankashyap:
@p3ck Current PR is a hard exit. However w.r.to destroy linchpin reports and continues with execution.

johnbieren:
Commenting just so I get the emails for activity on this

samvarankashyap:
[merge]

PR #839, opened 2018-12-07
use gitHubAPI to do the merge

p3ck:
This library method gives us the merge commit so we can pass that to the next job.

Tested this locally and it works with the latest updates to contra-lib


p3ck:
[merge]

Issue #853, opened 2018-12-13
Linchpin fails to run under any conditions


greg-hellings:
A new install of Linchpin fails to run under any conditions.

This affects, minimally, versions 1.6.4 and 1.6.5. The versions fail with a module import issue:

Version 1.6.5 installed into a virtualenv
```
$ linchpin
Traceback (most recent call last):
  File "/home/ghelling/venv/linchpin/bin/linchpin", line 7, in <module>
    from linchpin.shell import runcli
  File "/home/ghelling/venv/linchpin/lib/python3.7/site-packages/linchpin/shell/__init__.py", line 13, in <module>
    from linchpin.cli import LinchpinCli
  File "/home/ghelling/venv/linchpin/lib/python3.7/site-packages/linchpin/cli/__init__.py", line 16, in <module>
    from linchpin.fetch import FETCH_CLASS
  File "/home/ghelling/venv/linchpin/lib/python3.7/site-packages/linchpin/fetch/__init__.py", line 1, in <module>
    from fetch_http import FetchHttp
ModuleNotFoundError: No module named 'fetch_http'
```

Version 1.6.4 installed from a provisionally created RPM:
```
        $ linchpin
        Traceback (most recent call last):
          File "/usr/bin/linchpin", line 11, in <module>
            load_entry_point('linchpin==1.6.4', 'console_scripts',
                             'linchpin')()
          File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py",
               line 484, in load_entry_point
            return get_distribution(dist).load_entry_point(group, name)
          File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py",
               line 2714, in load_entry_point
            return ep.load()
          File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py",
               line 2332, in load
            return self.resolve()
          File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py",
               line 2338, in resolve
            module = __import__(self.module_name, fromlist=['__name__'],
                                level=0)
          File "/usr/lib/python3.7/site-packages/linchpin/shell/__init__.py",
               line 12, in <module>
            from linchpin.cli import LinchpinCli
          File "/usr/lib/python3.7/site-packages/linchpin/cli/__init__.py",
               line 16, in <module>
            from linchpin.fetch import FETCH_CLASS
          File "/usr/lib/python3.7/site-packages/linchpin/fetch/__init__.py",
               line 1, in <module>
[...]

PR #854, opened 2018-12-13
Force relative imports to be explicit

greg-hellings:
Python has long supported explicit relative imports and encouarged them
in the style guide. However, in the days of Python 3, this is now a
requirement of the language. This syntax still works in Python 2.

Addresses #853

### DIFF
--- a/linchpin/__init__.py
+++ b/linchpin/__init__.py
@@ -10,6 +10,7 @@ import hashlib
 
 from uuid import getnode as get_mac
 from collections import OrderedDict
+from six import text_type
 
 from linchpin.ansible_runner import ansible_runner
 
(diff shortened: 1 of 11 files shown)


greg-hellings:
Please don't merge yet. I'm still working on fixing the Python 3 test failures.

greg-hellings:
OK, so I'm not going to be able to get the tests all passing in Python 3. There are some changes that go deeper than my knowledge of linchpin's internals now reach. I'm attaching some fixes that do correct some of the errors in Python 3. I also cleaned up a number of code warnings, though not all, that are reported during test execution.

samvarankashyap:
[merge]

PR #874, opened 2018-12-19
Enable unique attribute for all resource roles

samvarankashyap:
Enable unique attribute for all resources types in openstack


14rcole:
What is the point of the unique attribute?

samvarankashyap:
@14rcole purpose to enable and disable uhash at topology level, if unique is false , uhash is disabled in resource provisioned. This is custom to linchpin.

PR #879, opened 2018-12-26
simple examples

abraverm:


14rcole:
As we begin adding use cases to `workspaces` that go beyond the generic provider-specific workspaces, I think we should be including a README.md in each workspace with a short description of the workspace and it's use case.  Can you add one to these workspaces?  I'll open an issue to add READMEs to existing ones

abraverm:
Sure, sounds good :+1:

14rcole:
This is great, thank you

abraverm:
[test]

abraverm:
[merge]

PR #866, opened 2018-12-18
Inventory duplication ordering

14rcole:
Fixes #801

PR #884, opened 2019-01-03
Reduce the number of jobs we keep due to storage issues.

p3ck:


14rcole:
you're getting the same version conflict I was getting in #866.  Bump the minimum version for requests to `2.21.0` in requirements.txt and it should be fixed

PR #886, opened 2019-01-04
Add 'unique' field to os_network

14rcole:
The `unique` field, which exists for all other openstack resources, was omitted from the os_network schema.  Adding this back in fixes the validation error preventing upshift tests in #866 from passing


14rcole:
[merge]

Issue #778, opened 2018-11-05
Linchpin validate fails on following Pinfile


samvarankashyap:
```
---
ember:
  topology:
    topology_name: ember
    resource_groups:
    - resource_group_name: libvirt-single
      resource_group_type: libvirt
      resource_definitions:
      - role: libvirt_node
        name: centos71
        uri: qemu:///system
        image_src: http://cloud.centos.org/centos/7/images/CentOS-7-x86_64-GenericCloud-1608.qcow2.xz
        count: 1
        networks:
          - name: default
  layout:
    inventory_layout:
      vars:
        hostname: __IP__
      hosts:
        example-node:
          count: 1
          host_groups:
            - example
```

Error:
```
(nummy) 🎩linchpin -vvv up
Processing target: ember
rundb_id: 2
uhash: e5816f
Topology 'OrderedDict([('topology_name', 'ember'), ('resource_groups', [OrderedDict([('resource_group_name', 'libvirt-single'), ('resource_group_type', 'libvirt'), ('resource_definitions', [OrderedDict([('role', 'libvirt_node'), ('name', 'centos71'), ('uri', 'qemu:///system'), ('image_src', 'http://cloud.centos.org/centos/7/images/CentOS-7-x86_64-GenericCloud-1608.qcow2.xz'), ('networks', [OrderedDict([('name', 'default')])]), ('count', 1)])])])])])' does not validate.  For more information run `linchpin validate`
(nummy) 🎩linchpin validate
Traceback (most recent call last):
  File "/home/srallaba/.virtualenvs/nummy/bin/linchpin", line 11, in <module>
    load_entry_point('linchpin', 'console_scripts', 'linchpin')()
  File "/home/srallaba/.virtualenvs/nummy/lib/python2.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/srallaba/.virtualenvs/nummy/lib/python2.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/srallaba/.virtualenvs/nummy/lib/python2.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/srallaba/.virtualenvs/nummy/lib/python2.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
[...]

PR #887, opened 2019-01-04
Move validation to separate module

14rcole:
This will make it easier to modify and expand validation in the future.  Precursor to my fi for #778


p3ck:
the unit tests are failing. looks like it can't find the new validator module

14rcole:
[merge]

Issue #844, opened 2018-12-11
beaker: allow injecting ssh keys inside topologies files


tareqalayan:
similar to libvirt where you can set ssh_key
or openstack where you can set key_pair

we would like something for beaker.

tareqalayan:
similar to libvirt where you can set ssh_key 
or openstack where you can set key_pair

we would like something for beaker.

tareqalayan:
@abraverm @arilivigni can you please help with this?

samvarankashyap:
@beerparty @p3ck : Do we have a that feature in beaker ?? Any thoughts ?

p3ck:
I can take this one.  We can do this via kickstart snippets I believe.

We're just talking about storing the public key on the system being provisioned right?

dollierp:
My 2 cents: It would be nice to expose beaker ks_append feature in linchpin. SSH key injection could be done by adding a ks %post scriplet via ks_append and it could also be used for other use cases.

abraverm:
Here is a working example using tasks:
```yaml
---
simple:
  topology:
    topology_name: custom
    resource_groups:
      - resource_group_name: custom
        resource_group_type: beaker
        resource_definitions:
          - role: bkr_server
            recipesets:
              - distro: RHEL-7.5
                arch: x86_64
                count: 1
                variant: Server
                name: rhel75
                tasks:
                  - name: '/installation/sshkey'
                taskparam:
[...]

PR #888, opened 2019-01-04
Support injecting ssh key into hosts provisioned through beaker

p3ck:
Also supports ks_append in general.

Implements RFE's from Issue #844


p3ck:
[merge]

PR #892, opened 2019-01-07
Validate high-level components of topologies

14rcole:
Topology validation only considers resource groups, which are provider-specific.  This validates the general components of the topology as well


14rcole:
Addresses #778

14rcole:
[merge]

PR #899, opened 2019-01-10
Fix lack of return code for shell hooks

14rcole:
Shell hook execution previously did not return anything, which would result in an exception being thrown even if the hooks were successful. This PR makes shell hook execution return 0, or the error code of the first error.  It also prints stderr to the console along with stdout.

Finally, this includes the integration test that was referenced in #898


14rcole:
[merge]

Issue #880, opened 2018-12-26
RFE: Beaker support for kickstart


abraverm:
Beaker allows kickstart to be written within the recipe element, this can allow simpler inline (PinFile) customization of deployments, and with [Jinja include|http://jinja.pocoo.org/docs/2.10/templates/#include| it could be reused. For example:
Pinfile:
```yaml
---
simple:
  topology:
    topology_name: simple
    resource_groups:
      - resource_group_name: simple
        resource_group_type: beaker
        resource_definitions:
          - role: bkr_server
            recipesets:
              - distro: RHEL-6.5
                arch: x86_64
                count: 1
                name: rhel65
                kickstart: |
{% macro kickstart() %}{% include 'kickstart' -%}{%- endmacro -%}
                {{ kickstart() | indent(18, True) }}
```
result:
```xml
<job id="3262259" owner="abraverm@redhat.com" result="New" status="Installing" retention_tag="scratch">
  <whiteboard>Provisioned by LinchPin</whiteboard>
  <recipeSet priority="Normal" response="ack" id="5160991">
    <recipe id="6347265" job_id="3262259" recipe_set_id="5160991" whiteboard="" role="None" kickstart_url="http://beaker.engineering.redhat.com/kickstart/5031795" ks_meta="method=nfs" kernel_options="" kernel_options_post="" start_time="2018-12-26 11:37:06" result="New" status="Installing" distro="RHEL-6.5" arch="x86_64" family="RedHatEnterpriseLinux6" variant="Client" system="lenovo-rd230-02.khw.lab.eng.bos.redhat.com">
      <autopick random="false" />
      <watchdog />
      <installation />
      <packages />
      <ks_appends />
      <roles>
        <role value="None">
          <system value="lenovo-rd230-02.khw.lab.eng.bos.redhat.com" />
        </role>
      </roles>
      <repos />
      <distroRequires>
        <and>
          <distro_name op="=" value="RHEL-6.5" />
          <distro_arch op="=" value="x86_64" />
        </and>
      </distroRequires>
      <hostRequires>
        <system_type value="Machine" />
      </hostRequires>
      <partitions />
      <logs>
        <log href="https://beaker.engineering.redhat.com/recipes/6347265/logs/console.log" name="console.log" />
      </logs>
      <task name="/distribution/dummy" role="STANDALONE" version="1.0-0" id="86228168" result="New" status="Waiting" avg_time="300">
[...]

PR #900, opened 2019-01-10
Allow kickstart option in beaker topology

dollierp:
This argument is used underlyingly by beaker client library, its value
must be set to the path of a kickstart file present on the system
running linchpin.


dollierp:
BTW it is related to #880.

docs/source/beaker.rst
```diff
@@ -106,6 +106,8 @@ describes the available recipesets options.
 +---------------------+------------+----------+-----------------------------------------+
 | ssh_key             | false      | list     | list of strings                         |
 +---------------------+------------+----------+-----------------------------------------+
+| kickstart           | false      | string   | N/A                                     |
```
p3ck: Should we specify that this should be an absolute path to a kickstart template?

dollierp: I may definitively be useful. Does the last column fit this purpose?

p3ck: yes - that would be fine.  Thank you for the patch

dollierp: updated with description

PR #905, opened 2019-01-15
Remove import of update_aws_cfn_outputs.yml

14rcole:
Remove the above dependency from the playbooks, as the file no longer exists.  Also gets rid of a long warning that prints every time linchpin is run


14rcole:
[merge]

PR #925, opened 2019-02-04
Inventory template name

14rcole:
We'll need to work to clean up the code, templates are a bit of a mess right now and I'm sure there's a weird edge case or two hiding somewhere.  But for now things seem to work roughly as expected and this won't hold up the next release

### DIFF
--- a/docs/source/examples/workspaces/dummy/PinFile
+++ b/docs/source/examples/workspaces/dummy/PinFile
@@ -30,8 +30,7 @@ dummy-new:
         all:
           vars:
             ansible_user: root
-            ansible_private_key_file: |
-              "{% raw -%}{{ lookup('env', 'TESTLP') | default('/tmp', true) }}/CSS/keystore/css-central{%- endraw %}"
+            ansible_private_key_file: "{% raw -%}{{ lookup('env', 'TESTLP') | default('/tmp', true) }}/CSS/keystore/css-central{%- endraw %}"
 
 dummy-topo:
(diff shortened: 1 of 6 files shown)


p3ck:
Not sure what is actually failing.  but I see this in the test logs:

Action 'up' on Target 'dummy-new' is complete
Error: unexpected char u'\\' at 310

When executing the 01_dummy-template-data and 02_dummy-template-data-file tests

docs/source/examples/workspaces/dummy/PinFile
```diff
@@ -30,8 +30,7 @@ dummy-new:
         all:
           vars:
             ansible_user: root
-            ansible_private_key_file: |
-              "{% raw -%}{{ lookup('env', 'TESTLP') | default('/tmp', true) }}/CSS/keystore/css-central{%- endraw %}"
+            ansible_private_key_file: "{% raw -%}{{ lookup('env', 'TESTLP') | default('/tmp', true) }}/CSS/keystore/css-central{%- endraw %}"
```
14rcole: @p3ck This was the problem.  With the template on a separate line, the string was treated as if it did not have any quotes.  Single quotes were added and the ones in the middle of the string were escaped.  These then became '\\\' while the templates were being rendered, hence the strange error

p3ck:
AWS is failing because layout_data is undefined.

14rcole:
[merge]

PR #926, opened 2019-02-04
Fix json inventory formatter configs

### LINKED ISSUE #918: json inventory filters does not support cfgs
I seem to have overlooked this in #703.  We should also have integration tests that specifically verify the json output

14rcole:
Fixes #918

### DIFF
--- a/linchpin/InventoryFilters/JSONInventoryFormatter.py
+++ b/linchpin/InventoryFilters/JSONInventoryFormatter.py
@@ -73,7 +73,7 @@ class JSONInventoryFormatter(InventoryFormatter):
                         self.config["all"]["hosts"].append(ip)
         return True
 
-    def add_common_vars(self, host_groups, layout):
+    def add_common_vars(self, host_groups, layout, config):
         # defaults cvrs to [] when they doesnot exist
         # cvrs --> common_vars
         host_groups.append("all")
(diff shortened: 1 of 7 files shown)


linchpin/InventoryFilters/JSONInventoryFormatter.py
```diff
@@ -85,11 +85,13 @@ def add_common_vars(self, host_groups, layout):
                 self.config["_meta"]["hostvars"] = {}
             for item in items:
                 for var in cvrs:
-                    self.config["_meta"]["hostvars"][item] = {}
-                    if cvrs[var] == "__IP__":
```
p3ck: We don't need this special case anymore?

linchpin/InventoryFilters/JSONInventoryFormatter.py
```diff
@@ -73,7 +73,7 @@ def add_ips_to_groups(self, inven_hosts, layout):
                         self.config["all"]["hosts"].append(ip)
         return True
 
-    def add_common_vars(self, host_groups, layout):
+    def add_common_vars(self, host_groups, layout, config):
```
p3ck: I see a new argument added here but I don't see any references to the code that calls this.

14rcole: The code that calls it was updated in #703.  It's the same as the code that calls this method in `CFGInventoryFormatter.py`

linchpin/InventoryFilters/JSONInventoryFormatter.py
14rcole: No, if `__IP__` isn't defined in the config it's added with a default value, so this case will be handled in the for loop below (`for cfg_item in config:`)

14rcole:
[test]

14rcole:
[test]

14rcole:
[test]

14rcole:
[merge]

Issue #909, opened 2019-01-19
Inventory incomplete with beaker when count > 1


dollierp:
When using a beaker topology with count > 1, the generated inventory contains only one host instead of the n hosts requested.

It is not a provisioning issue: the job sent to beaker provisions successfully all the hosts requested but the inventory generated contains only one of them.

Git bisect points out a regression from #800.

This simple topology allows to reproduce the issue:

```
---
beaker_inventory:
  topology:
    topology_name: beaker_topology
    resource_groups:
      - resource_group_name: beaker_group
        resource_group_type: beaker
        resource_definitions:
          - role: bkr_server
            whiteboard: Provisioned by LinchPin
            cancel_message: Job cancelled on account of rain
            max_attempts: 60
            attempt_wait_time: 60
            recipesets:
              - name: beaker_system
                family: RedHatEnterpriseLinux7
                tags:
                  - released
                variant: Server
                arch: x86_64
                count: 3

  layout:
    inventory_layout:
      vars:
        hostname: __IP__
      hosts:
        beaker_system:
          count: 3
          host_groups:
            - beaker_nodes
```

If (as a workaround) I declare as many recipesets with count = 1 as the number of hosts wanted, the inventory is generated correctly:
```
---
beaker_inventory:
  topology:
    topology_name: beaker_topology
    resource_groups:
      - resource_group_name: beaker_group
        resource_group_type: beaker
        resource_definitions:
          - role: bkr_server
            whiteboard: Provisioned by LinchPin
            cancel_message: Job cancelled on account of rain
            max_attempts: 60
            attempt_wait_time: 60
            recipesets:
              - &recipeset
                name: beaker_system
                family: RedHatEnterpriseLinux7
                tags:
                  - released
                variant: Server
                arch: x86_64
                count: 1
              - <<: *recipeset
              - <<: *recipeset


  layout:
    inventory_layout:
      vars:
        hostname: __IP__
      hosts:
        beaker_system:
          count: 3
          host_groups:
            - beaker_nodes
```

dollierp:
Git bisect says it broke in 5bb7cf6 from #800 which fix #795.

p3ck:
@dollierp you are correct. If I take 1.7.0 and revert that commit it works as expected.

samvarankashyap:
fixed in #914

PR #928, opened 2019-02-05
Add integration test to check for correct number of hosts in beaker

14rcole:
Tests for this will fail until #926 is merged.  This is the integration test corresponding to #909.  I can modify it so that it passes without #926 if necessary but it'll be sloppy


14rcole:
[test]

14rcole:
[merge]

PR #930, opened 2019-02-05
Specify /usr/sbin/init for fedora28

p3ck:


p3ck:
I know it doesn't look like this patch should make a difference but it comes down to the oci-systemd-hook

https://github.com/projectatomic/oci-systemd-hook

p3ck:
[merge]

PR #937, opened 2019-02-12
Intro ssh_key_file property on Beaker's topology

wainersm:
Linchpin provides the ssh_key property which allows to set (statically)
ssh key's content. However one may want to use a dynamically generated
key file, and that's what is solved on this change.

It was introduced the ssh_key_file property for recipesets, that you can set
a list of file names from where read the ssh public key. For example:
... snip ...
recipesets:
  - family: RedHatEnterpriseLinux8
    name: rhel80-x86_64
    arch: x86_64
    variant: BaseOS
    count: 1
    ssh_key_file:
      - ssh_key.pub
    hostrequires:
.. snip ..

The key files should be stored on the Linchpin's credentials directory, or
on alternative location and the ssh_keys_path property for resource_definitions
should be set accordingly.

### DIFF
--- a/docs/source/beaker.rst
+++ b/docs/source/beaker.rst
@@ -39,6 +39,9 @@ role definition, the following options are available.
 +-------------------+------------+----------+-------------------+-----------------+
 | attempt_wait_time | false      | integer  | attempt_wait_time |                 |
 +-------------------+------------+----------+-------------------+-----------------+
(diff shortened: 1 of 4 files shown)


wainersm:
@p3ck Hello Bill! I appreciate if you could review this as well. thanks!

samvarankashyap:
@p3ck @14rcole Shall we merge this ?

p3ck:
[merge] - sounds good. :-)

PR #953, opened 2019-02-18
Update Fedora containers

p3ck:
PR/945 should have updated the fedora containers


samvarankashyap:
@p3ck I think change in dockerfiles would not update containers unless there is a release or the containers are manually updated via openshift

p3ck:
@samvarankashyap It has always worked this way.  Why do you think otherwise?

p3ck:
[merge]

Issue #540, opened 2018-04-13
Inventory is not correctly populated after provisioning with libvirt


dyasny:
Here's what the IPs are:
```
# for i in `virsh list --name`; do echo $i; virsh domifaddr $i; done
master_0
 Name       MAC address          Protocol     Address
-------------------------------------------------------------------------------
 vnet0      52:54:00:a0:5b:d8    ipv4         192.168.122.12/24
 
node_0
 Name       MAC address          Protocol     Address
-------------------------------------------------------------------------------
 vnet1      52:54:00:e4:01:d7    ipv4         192.168.122.250/24
 
node_1
 Name       MAC address          Protocol     Address
-------------------------------------------------------------------------------
 vnet2      52:54:00:97:79:d4    ipv4         192.168.122.10/24
```
And here's the resulting inventory:
```
[nodes]
192.168.122.10 hostname=192.168.122.10
192.168.122.250 hostname=192.168.122.250
192.168.122.12 hostname=192.168.122.12
 
[nfs]
192.168.122.10 hostname=192.168.122.10
 
[masters]
192.168.122.10 hostname=192.168.122.10
```

As you can see above, master is 192.168.122.12 ad the assigned master is in fact a node - 192.168.122.10

samvarankashyap:
@dyasny could you please share the topology file and layout file you have used to create the above inventory ?

dyasny:
```yaml
inventory_layout:
#  inventory_file: cnv-inventory.inv
  vars:
    hostname: __IP__
  hosts:
    "master":
      count: 1
      host_groups:
        - masters
        - etcd
        - nodes
        - nfs
    "node":
      count: 2
      host_groups:
        - nodes
  host_groups:
    OSEv3:
      children:
        - masters
        - nodes
        - etcd
        - nfs
      vars:
        ansible_ssh_user: root
        openshift_schedulable: true
        openshift_node_labels: "{'region': 'infra','zone': 'default', 'node-role.kubernetes.io/compute': 'true'}"
        openshift_deployment_type: openshift-enterprise
        openshift_clock_enabled: true
        openshift_master_identity_providers: "[{'name': 'allow_all_auth', 'login': 'true', 'challenge': 'true', 'kind': 'AllowAllPasswordIdentityProvider'}]"
        openshift_disable_check: "memory_availability,disk_availability"
        openshift_enable_excluders: false
        template_service_broker_install: false
        openshift_use_manageiq: false
[...]

PR #959, opened 2019-02-20
Fix dict falling out of order in template rendering

14rcole:
The _render_template() function converted an OrderedDict to a dict so that its templates could be rendered.  This converts the OrderedDict to a string to render the template so that key order is maintained until the data can be converted back to an OrderedDict

### DIFF
--- a/linchpin/cli/__init__.py
+++ b/linchpin/cli/__init__.py
@@ -644,8 +644,8 @@ class LinchpinCli(LinchpinAPI):
                 data = strm.read()
 
         try:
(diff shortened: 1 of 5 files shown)


14rcole:
@lukas-bednar This should fix #540

samvarankashyap:
@14rcole could you remove the debug statements before you merge . Thanks !

14rcole:
@samvarankashyap Will do

14rcole:
[test]

14rcole:
[merge]

Issue #960, opened 2019-02-20
Malformed libvirt inventory with custom libvirt cfgs and count > 1 libvirt_node


mcornea:
The following Pinfile:

```raw 
---
libvirt-network:
  topology:
    topology_name: "libvirt_network"
    resource_groups:
      - resource_group_name: "libvirt_network"
        resource_group_type: "libvirt"
        resource_definitions:
          - name: external
            role: libvirt_network
            uri: qemu:///system
            ip: 192.168.123.1
            dhcp_start: 192.168.123.100
            dhcp_end: 192.168.123.150
            domain: redhat.local
            bridge: "external"
          - name: internal
            role: libvirt_network
            uri: qemu:///system
            ip: 192.168.124.1
            bridge: internal

cfgs:
  libvirt:
    __IP__: name
    __ADDRESS__: ip
 
libvirt-new:
  topology:
    topology_name: rhhi
    resource_groups:
      - resource_group_name: libvirt
        resource_group_type: libvirt
        resource_definitions:
          - role: libvirt_node
            name: master
            uri: qemu:///system
            count: 3
            image_src: file:///var/lib/libvirt/images/rhel-guest-image-7.6-215.x86_64.qcow2
            memory: 16384
            vcpus: 4
            arch: x86_64
            ssh_key: id_rsa
            networks:
              - name: external
              - name: internal
  layout:
    inventory_layout:
      vars:
        hostname: __IP__
        ansible_ssh_host: __ADDRESS__
      hosts:
        master:
          count: 3
          host_groups:
            - masters
```
outputs the inventory below:

```raw
[masters]
master_2 ansible_ssh_host=192.168.123.133 ansible_ssh_host=192.168.123.116 ansible_ssh_host=192.168.123.129 hostname=master_2 hostname=master_2 hostname=master_2

[all]
master_2 ansible_ssh_host=192.168.123.133 ansible_ssh_host=192.168.123.116 ansible_ssh_host=192.168.123.129 hostname=master_2 hostname=master_2 hostname=master_2
```
expected inventory is:

```raw
[masters]
master_0 ansible_ssh_host=192.168.123.133 hostname=master_0
master_1 ansible_ssh_host=192.168.123.116 hostname=master_1
master_2 ansible_ssh_host=192.168.123.129 hostname=master_2

[all]
master_0 ansible_ssh_host=192.168.123.133 hostname=master_0
[...]

PR #962, opened 2019-02-20
Fix issue #960

mcornea:
This change implements the following changes:

 - register res_count var as a list by using with_sequence. Previously
   the res_count var was registered as a shell task output which made
   res_count.stdout a string and didn't allow iterating over the items
   in the list.

 - store output_name var as a list allowing iteration over multiple
   elements. Previously only the last iteration item was stored in the
   output_name.

### DIFF
--- a/linchpin/provision/roles/libvirt/tasks/copy_image_src.yml
+++ b/linchpin/provision/roles/libvirt/tasks/copy_image_src.yml
@@ -89,7 +89,7 @@
     remote_src: false
   with_nested:
     - ["{{ local_image_path }}"]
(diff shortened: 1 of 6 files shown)


mcornea:
Fixes #960

samvarankashyap:
@mcornea Could you please squash commits into one commit before we merge it . Thank you :+1:

mcornea:
@samvarankashyap Commits are now squashed.

14rcole:
@mcornea Thanks for fixing this! [merge]

PR #970, opened 2019-02-21
Allow creating bridge libvirt networks

### LINKED ISSUE #969: Allow creating simple bridge libvirt networks
Currently forward mode is enforced for libvirt networks so we cannot define non-forward(bridged) networks:

https://github.com/CentOS-PaaS-SIG/linchpin/blob/develop/linchpin/provision/roles/libvirt/templates/libvirt_network.xml.j2#L6

We need to ability to define simple bridge networks, e.g:

<network connections='4' ipv6='yes'>
  <name>brovc</name>
  <uuid>66004e03-134a-499a-9bdf-a28fd1dc8d6a</uuid>
  <bridge name='brovc' stp='on' delay='0'/>
  <mac address='52:54:00:95:82:36'/>
</network>

mcornea:
- do not require the ip attribute to be set
- do not required forward mode attribute to be set

Fixes #969


mcornea:
@abraverm @samvarankashyap Hi! Could you please review this change? Thanks!

Issue #921, opened 2019-01-28
OKD Provider: Doesn't recognize <provider>.creds, clouds.yaml


skatlapa:
Atm, the okd provider doesn't recognize the values passed into provider.creds (same as clouds.yaml for openstack.creds used in HDSL). We see this error[1] when used with default profile with this format[2].

[1] - http://pastebin.test.redhat.com/703337

[2] - http://pastebin.test.redhat.com/703338

skatlapa:
@p3ck @samvarankashyap @14rcole looping in

14rcole:
what do you get when you run `linchpin validate`?

skatlapa:
Hi, @14rcole I did not run Linchpin Validate, I shall try it and let you know. However, that wouldn't help us much I think here the LP is not honoring openshift.creds for its credential file (unlike openstack provider).

Steps to re-create:
1) Create the credentials as per the format[2] and name the file openshift.creds(or *.creds)
2) Do not specify --creds-path (LP defaults to the environment variable - here is the problem)
3) You'll see the error [1]
[1,2] referenced in the first comment.

I am planning to get back to this on Monday, if you find/solve by then please let me know. Thanks.

14rcole:
@skatlapa did you try this out with the latest release of linchpin?

Issue #929, opened 2019-02-05
OKD: --creds-path if not mentioned the provider defaults to upshift.yml  and does not recognize <provider_file>.creds


skatlapa:
OKD: --creds-path if not mentioned the provider defaults to upshift.yml and does not recognize <provider_file>.creds

We need to have a way to not pass --creds-path and be using openshift.creds, just like OpenStack provider.
It was previously raised as: #770

skatlapa:
OKD: --creds-path if not mentioned the provider defaults to upshift.yml and does not recognize <provider_file>.creds

We need to have a way to not pass --creds-path and be using openshift.creds, just like OpenStack provider. 
It was previously raised as: https://github.com/CentOS-PaaS-SIG/linchpin/issues/770

samvarankashyap:
fixed by #973

PR #973, opened 2019-02-22
Search for creds in workspace when they are not found in defaults

samvarankashyap:
$subject

fixes : #921 #929

### DIFF
--- a/linchpin/provision/roles/openshift/tasks/provision_resource_group.yml
+++ b/linchpin/provision/roles/openshift/tasks/provision_resource_group.yml
@@ -36,6 +36,17 @@
   no_log: "{{ not debug_mode }}"
   when: res_grp['credentials'] is defined and res_grp['credentials']['filename'] is defined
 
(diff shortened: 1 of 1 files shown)


linchpin/provision/roles/openshift/tasks/provision_resource_group.yml
```diff
@@ -36,6 +36,17 @@
   no_log: "{{ not debug_mode }}"
   when: res_grp['credentials'] is defined and res_grp['credentials']['filename'] is defined
```
skatlapa: what do you think of adding debug task here to check the right workspace is being used:
- debug: msg="{{ lookup('env','workspace') }} is an workspace env variable"

samvarankashyap: @skatlapa I tested it on my workspace , I can add that .

skatlapa: +1

samvarankashyap:
[merge]

PR #982, opened 2019-02-26
Update issue templates

samvarankashyap:
Add issue templates for Feature request and bug reports


14rcole:
Does this have to be in the master branch in order to work? If so I say we try to get it into 1.7.2 so that it can take effect this week

samvarankashyap:
@14rcole I think it should work with the develop branch since develop is the default branch

samvarankashyap:
[merge]

Issue #956, opened 2019-02-19
Cancel to append 1 after rhos server name when set count as 1


tanfengshuang:
There is always 1 appended after rhos server name when count is 1, it's a little wired. I thinks it's better to append one number when count > 1.

And I tried to update file render_os_server_insts.py to cancel 1 appended after rhos server name when count is 1, the name is correct in task "set_fact instance filters"(rhos-server), but incorrect name in later task "openstack : generate res_def_names"(rhos-server1).
 
 ```
TASK [openstack : set_fact instance filters] *********************************************************************************************************************************
task path: /home/jenkins/.venv/lib/python2.7/site-packages/linchpin/provision/roles/openstack/tasks/provision_os_server.yml:6
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
rhos-server
[u'rhos-server']
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ok: [localhost] => {"ansible_facts": {"res_def_prov": [{"name": "rhos-server"}]}, "changed": false}
TASK [openstack : set_fact when addtional volumes] *********************************************************************************************************************************
task path: /home/jenkins/.venv/lib/python2.7/site-packages/linchpin/provision/roles/openstack/tasks/provision_os_server.yml:10
skipping: [localhost] => {"changed": false, "skip_reason": "Conditional result was False"}
TASK [openstack : Provision additional volumes] *********************************************************************************************************************************
task path: /home/jenkins/.venv/lib/python2.7/site-packages/linchpin/provision/roles/openstack/tasks/provision_os_server.yml:15
TASK [openstack : Detach additional volumes] *********************************************************************************************************************************
task path: /home/jenkins/.venv/lib/python2.7/site-packages/linchpin/provision/roles/openstack/tasks/provision_os_server.yml:28
TASK [openstack : Delete additional volumes] *********************************************************************************************************************************
task path: /home/jenkins/.venv/lib/python2.7/site-packages/linchpin/provision/roles/openstack/tasks/provision_os_server.yml:40
TASK [openstack : generate res_def_names] *********************************************************************************************************************************
task path: /home/jenkins/.venv/lib/python2.7/site-packages/linchpin/provision/roles/openstack/tasks/provision_os_server.yml:50
ok: [localhost] => (item=1) => {"ansible_facts": {"res_def_names": ["rhos-server1"]}, "changed": false, "inst_c": "1"}
```

[...]

PR #1001, opened 2019-03-05
Do not append count when there is only one instance to be provisioned

samvarankashyap:
$subject 
fixes: #956


samvarankashyap:
[merge]

PR #1002, opened 2019-03-06
Add OpenStack tutorial

14rcole:

PR #1004, opened 2019-03-07
Openstack tutorial template

14rcole:
This PR templates out the tutorial for OpenStack so that the information that will remain the same between providers stays that way.  I would have preferred to use jinja templates for even more flexibility but it doesn't look like sphinx supports that.  Note that this shouldn't be merged until after #1002.


14rcole:
[merge]

PR #1005, opened 2019-03-07
use only virsh define to bring up libvirt instances

p3ck:
Previously, depending if you defined cloud-config in your topology
the libvirt provisioner would either use virt-install or virt ansible
module. This patch eliminates the virt-install path and consolidates
on just using the ansible virt module.


config/Dockerfiles/run_tests.sh
```diff
@@ -7,7 +7,7 @@ shift
 DISTROS=$*
 #TARGETS="dummy os-server-new os-sg-new os-vol-new"
 
-PROVIDERS="dummy openstack"
+PROVIDERS="dummy libvirt"
```
14rcole: Was this supposed to be a temporary change, or was having openstack there wrong to begin with?

p3ck: I believe it was a mistake to begin with.  The PROVIDERS in run_tests.sh don't affect what we run in jenkins, only when running locally with make test.  libvirt will work locally, the openstack provider here will only work if you have a openstack clouds.yaml in the keys directory.

samvarankashyap:
@p3ck : Checking whether is changes anything related to virt-customize

samvarankashyap:
@p3ck Needs rebase .

p3ck:
[merge]

PR #986, opened 2019-02-26
Introduce name_separator attribute for libvirt resource

mcornea:
This change allows creating VMs with a custom separator between the
name and index, e.g:

$name-$index; currently only $name_$index is possible

Fixes #985

PR #1009, opened 2019-03-08
Adjust OpenStack instance naming

mcornea:
This change adjusts the OpenStack instances naming so that:
 - index starts at 0 instead of 1
 - there is a '-' separator between the resource name and index

This makes the naming consistent with the libvirt provider.

### DIFF
--- a/linchpin/provision/roles/openstack/tasks/provision_os_server.yml
+++ b/linchpin/provision/roles/openstack/tasks/provision_os_server.yml
@@ -49,8 +49,8 @@
 
 - name: generate res_def_names
   set_fact:
-    res_def_names: "{{ res_def_names  + [os_resource_name + inst_c] }}"
-  with_sequence: start=1 end={{ res_def['count'] }}
+    res_def_names: "{{ res_def_names  + [os_resource_name + '-' + inst_c] }}"
+  with_sequence: start=0 end={{ res_def['count']|int - 1 }}
   when: res_def['count'] > 1
   loop_control:
     loop_var: inst_c


samvarankashyap:
@mcornea @14rcole @p3ck : to be merged after #1001

14rcole:
This is nit, but we ordinarily use an underscore instead of a dash.  Can we keep int consistent?

mcornea:
@14rcole Underscore is not allowed in the node's hostname for our use case. I can make it configurable and use underscore as the default, similar to what I did in https://github.com/CentOS-PaaS-SIG/linchpin/pull/986  Would that be ok?

14rcole:
@mcornea don't worry about it.  If someone needs it to be underscores we can make it configurable then

samvarankashyap:
@mcornea You could use linchpin.constants evars section to make it configurable. 
https://github.com/CentOS-PaaS-SIG/linchpin/blob/develop/linchpin/linchpin.constants#L31  
Any variable mentioned in evars section would be available as extravar in provision playbooks

14rcole:
@mcornea Let us know if you want to use evars to make instance naming configurable or just leave what you have for now.  Either way, can you rebase this?

mcornea:
> @mcornea Let us know if you want to use evars to make instance naming configurable or just leave what you have for now. Either way, can you rebase this?

@14rcole I'd prefer leaving it like this and making it configurable in a subsequent patch if that's ok. Rebasing in the next minutes.

mcornea:
@rycole updated, thanks!

samvarankashyap:
[merge]

PR #1011, opened 2019-03-11
paramaterized examples

### LINKED ISSUE #816: Parameterized examples
Most of the examples have hard coded settings (image, network, etc). If a user wants to try the example, then the example has to be modified first.  Alternatively we can parameterize the examples, make them more like templates, and then users will just apply their settings from command line or a json file.

abraverm:
fixes #816


p3ck:
```
o distro tree matches Recipe: <distroRequires><and><distro_name op=\"=\" value=\"centos7-\"/
```
name is being passed to the topology but centos7- is not a valid distro name.

abraverm:
[merge]

PR #1015, opened 2019-03-13
Prevent dicts in template from being treated as strings

14rcole:
We load templates as an OrderedDict, which Ansible cannot parse in its template rendering.  Until Ansible can, which should be shortly after it stops supporting python 2.6, we can revert this change

### DIFF
--- a/linchpin/utils/dataparser.py
+++ b/linchpin/utils/dataparser.py
@@ -77,7 +77,10 @@ class DataParser(object):
             A dictionary of variables to be rendered againt the template
         """
 
-        c = self.parse_json_yaml(context, ordered=ordered)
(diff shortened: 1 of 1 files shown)


samvarankashyap:
@14rcole would this create any issues with inventory logic, given the recent changes in how we handle layout and inventory ??

14rcole:
@samvarankashyap As long as the layout is rendered correctly it shouldn't impact the inventory and if it isn't rendered correctly, the inventory will be incorrect no matter what.  What do you mean by recent changes though?

samvarankashyap:
@14rcole: There were changes with respect to in adding resource_types to outputs. However, I believe it should not effect.

14rcole:
[merge]

PR #1017, opened 2019-03-13
Make libvirt network destroy associated nodes

14rcole:


linchpin/provision/filter_plugins/get_network_domains.py
```diff
@@ -0,0 +1,60 @@
...
+def get_network_domains(network, uri):
+    network_hosts = []
+    #conn = libvirt.openReadOnly(uri)
+    conn = libvirt.open(uri)
+    if conn == None:
+        return network_hosts
+
+    hosts = conn.listDomainsID()
+
+    for host in hosts:
```
14rcole: This loop is n^3.  If there's a better way to do it, I'm open for it but this is what I could gather from the Libvirt documentation.  I assume that linchpin will be used on hosts with few enough nodes that it won't make a difference.

14rcole:
[merge]

PR #1041, opened 2019-03-22
Allow passing cpu mode to libvirt VM

mcornea:
Use host-model as default and make it configurable via cpu_mode parameter.


samvarankashyap:
[merge]

PR #1048, opened 2019-03-27
Move buildah push logic to a separate script

14rcole:


config/Dockerfiles/buildah-push.sh
```diff
@@ -0,0 +1,8 @@
+IMAGE=docker.io/${DOCKER_HUB_USERNAME}/${DOCKER_HUB_IMAGE}
+TAG=$1
+TAGGED_IMAGE=${IMAGE}:${TAG}
+
+buildah bud -t ${IMAGE} .
+buildah push --authfile ${docker_creds} ${IMAGE}
```
p3ck: Your irc question was probably related to ${docker_creds}.  My question is where is this defined even in the old version?  I don't see that it would have any value.  

DOCKER_HUB_USERNAME and DOCKER_HUB_IMAGE are defined in env. so they will be available.  but $IMAGE and $TAG won't be unless you assign them to env.IMAGE and env.TAG before calling the script.

johnbieren: IMAGE and TAG are defined in the shell script so should be fine. I don't see docker_creds anywhere though, so don't see this script working

config/Dockerfiles/JenkinsfileRelease
```diff
@@ -205,12 +205,7 @@ EOF
                                 script: "python setup.py --version",
                                 returnStdout: true
                                 ).trim()
-                            image = "docker.io/" + env.DOCKER_HUB_USERNAME + "/" + env.DOCKER_HUB_IMAGE
-                            tagged_img = image + ":" + version
-                            sh "buildah bud -t " + image + " ."
-                            sh "buildah push --authfile " + $docker-creds + " " + image
-                            sh "buildah tag " + image + " " + tagged_image
-                            sh "buildah push --authfile " + $docker-creds + " " + image
+                            pipelineUtils.executeInContainer(currentStage, "fedora29", "./config/Dockerfiles/buildah-push.sh "+version)
```
johnbieren: Have you tested that ```"./config/Dockerfiles/buildah-push.sh "+version)``` works? I have never used syntax like that... if it doesn't end up working ```./config/Dockerfiles/buildah-push.sh ${version}")``` would probably work

config/Dockerfiles/buildah-push.sh
14rcole: docker_creds is a credentials file I added to jenkins.  I thought I read that you access those files as environment variables but I could be wrong.

@p3ck in the old version it's $docker-creds but bash doesn't allow "-" in variable names

config/Dockerfiles/JenkinsfileRelease
14rcole: I know that the former is valid Groovy code for string concatenation but I haven't tested the code in this Jenkinsfile

johnbieren: Yes, agreed that it is valid groovy code for string concatenation, but I've never used string concatenation passed as an argument to executeInContainer, so I definitely think it is worth a quick test before it is merged

config/Dockerfiles/buildah-push.sh
johnbieren: I don't think that happens for free unless you do something like the prepareCredentials function does https://github.com/CentOS-PaaS-SIG/ci-pipeline/blob/master/src/org/centos/pipeline/PipelineUtils.groovy#L506

config/Dockerfiles/JenkinsfileRelease
14rcole: Is there a way to mock run a Jenkinsfile locally?

johnbieren: Locally? Not sure. You could just make a test job on the Jenkins master and strip out the majority of the code (leave in library checkout, pod creation type stuff) and just run a test executeInContainer

14rcole: I started to do that and then found [this line](https://github.com/CentOS-PaaS-SIG/linchpin/blob/develop/config/Dockerfiles/JenkinsfileRelease#L186), so it looks like it works

14rcole:
[merge]

PR #1055, opened 2019-04-01
Updated resource_names with prefix ci-lp-cp-

samvarankashyap:
$subject 

Update resource names to easily identify the provisioned resources.

### DIFF
--- a/docs/source/examples/workspaces/openstack/PinFile
+++ b/docs/source/examples/workspaces/openstack/PinFile
@@ -6,7 +6,7 @@ os-server-new:
       - resource_group_name: os-server-new
         resource_group_type: openstack
         resource_definitions:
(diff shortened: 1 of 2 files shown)


14rcole:
This may be better suited to a different PR, but do we want to change the default values for credentials?   `clouds.yaml` is a pretty generic name for an OpenStack creds file, but if we change it to `continuous-productization.yaml` we won't have to supply template data to our integration tests

samvarankashyap:
@14cole : 
We can just upload the credentials to jenkins with the ci-rhos profile and clouds.yaml filename. Hence we need not change. 
Does that sound good ?

14rcole:
@samvarankashyap That works for me

samvarankashyap:
[merge]

PR #1063, opened 2019-04-03
Remove fip_pool parameter from examples

samvarankashyap:
$subject 
Since floating ips are no longer used in psi infra
changes to make tests pass.


samvarankashyap:
[merge]

PR #1065, opened 2019-04-03
Automate the updating of github release notes

14rcole:


requirements.txt
```diff
@@ -19,3 +19,4 @@ configparser>=3.5.0
 openshift>=0.6
 pyasn1>=0.1.9
 pycrypto>=2.6.1
+pygithub>=1.43.5
```
samvarankashyap: @14rcole: Instead of having pygithub a hard requirement for linchpin. Shouldn't we include it in Dockerfiles? Since containers are responsible for the release. Though, I am okay with adding pygithub as a requirement of linchpin. It bothers me a little because it is unrelated to the functionality of linchpin.

config/Dockerfiles/release-notes.py
```diff
@@ -0,0 +1,114 @@
...
+
+    for title, items in body.items():
+        if len(items) == 0:
+            continue
+        body_str += title + "\n"
+        for item in items:
+            body_str += "* {0} #{1} by {2}\n".format(item.title, item.number,
+                                                   item.user.login)
+        body_str += "\n"
+    print body_str
```
samvarankashyap: Unnecessary print statement. If needed,  Please use python3 syntax

config/Dockerfiles/release-notes.py
```diff
@@ -0,0 +1,114 @@
+#!/usr/bin/python
+
+from github import Github
+import sys
+
+REPO="CentOS-PaaS-SIG/linchpin"
```
samvarankashyap: Is there any reason for hardcoding repo name Instead of passing it as runtime parameter?

config/Dockerfiles/JenkinsfileRelease
```diff
@@ -216,6 +216,17 @@ EOF
                             sh "git push origin " + env.ghprbActualCommit + ":" + MASTER_BRANCH
                         }
                     }
+                    stage('create GitHub release') {
+                        dir(reponame) {
+                            string version = sh (
+                                script python setup.py --version",
```
samvarankashyap: I think this would be syntax error given an extra " in line 222

config/Dockerfiles/release-notes.py
14rcole: We can have fewer runtime parameters this way.  More importantly, it's unlikely that the repo will ever change so there's really no harm in including as a constant, unlike the version and the key which are variable and secret respectively

requirements.txt
14rcole: That's a good point, I'll change that

14rcole:
[test]

samvarankashyap:
[test]

samvarankashyap:
[test]

14rcole:
@samvarankashyap gtg?

14rcole:
[merge]

Issue #1037, opened 2019-03-20
RFE: Make name seperator configurable for Openstack instances


samvarankashyap:
**Is your feature request related to a problem? Please describe.**
Currently, the name separator is hardcoded when the names are generated. 

**Describe the solution you'd like**
We would like to make it as a configurable parameter from linchpin.conf

samvarankashyap:
fixed by #1072

PR #1072, opened 2019-04-09
Adding delimiter for user defined names in OpenStack

skatlapa:
Fixes issue - https://github.com/CentOS-PaaS-SIG/linchpin/issues/1037

### DIFF
--- a/linchpin/provision/roles/openstack/tasks/provision_os_object.yml
+++ b/linchpin/provision/roles/openstack/tasks/provision_os_object.yml
@@ -16,7 +16,7 @@
 - name: "provision/deprovision swift containers when count is defined"
   os_object:
     auth: "{{ auth_var | omit_filter(omit) }}"
-    container: "{{ container.1 }}default_delimiter{{ container.4 }}"
+    container: "{{ container.1 }}{{ default_delimiter }}{{ container.4 }}"
     container_access: "{{ container.2 }}"
     wait: yes
     state: "{{ container.3 }}"
@@ -40,7 +40,7 @@
 - name: "Async:: provision/deprovision swift containers when count specified"
   os_object:
     auth: "{{ auth_var | omit_filter(omit) }}"
-    container: "{{ container.1 }}default_delimiter{{ container.4 }}"
+    container: "{{ container.1 }}{{ default_delimiter }}{{ container.4 }}"
     container_access: "{{ container.2 }}"
     wait: yes
     state: "{{ container.3 }}"
--- a/linchpin/provision/roles/openstack/tasks/provision_os_object.yml
+++ b/linchpin/provision/roles/openstack/tasks/provision_os_object.yml
@@ -16,7 +16,7 @@
 - name: "provision/deprovision swift containers when count is defined"
   os_object:
     auth: "{{ auth_var | omit_filter(omit) }}"
-    container: "{{ container.1 }}_{{ container.4 }}"
+    container: "{{ container.1 }}default_delimiter{{ container.4 }}"
     container_access: "{{ container.2 }}"
     wait: yes
     state: "{{ container.3 }}"
@@ -40,7 +40,7 @@
 - name: "Async:: provision/deprovision swift containers when count specified"
   os_object:
     auth: "{{ auth_var | omit_filter(omit) }}"
-    container: "{{ container.1 }}_{{ container.4 }}"
+    container: "{{ container.1 }}default_delimiter{{ container.4 }}"
     container_access: "{{ container.2 }}"
     wait: yes
     state: "{{ container.3 }}"
--- a/linchpin/provision/roles/openstack/tasks/provision_os_server.yml
+++ b/linchpin/provision/roles/openstack/tasks/provision_os_server.yml
@@ -49,7 +49,7 @@
 
 - name: generate res_def_names
   set_fact:
-    res_def_names: "{{ res_def_names  + [os_resource_name + 'default_delimiter' + inst_c] }}"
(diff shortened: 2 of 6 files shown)


skatlapa:
I will still have to fix the virtualenv stuff, so I'll remove WIP once I test the deployment there. Just pointing here for you guys to take a quick look, lmk if it looks good.

linchpin/provision/roles/openstack/tasks/provision_res_defs.yml
```diff
@@ -16,7 +16,7 @@
 
 - name: "Create name using uhash value"
   set_fact:
-    os_resource_name: "{{ os_resource_name + '-' + uhash + '-' }}"
+    os_resource_name: "{{ os_resource_name + 'default_delimiter' + uhash + 'default_delimiter' }}"
```
samvarankashyap: I don't think you need quotes for default_delimiter variable

linchpin/provision/roles/openstack/tasks/provision_os_server.yml
```diff
@@ -49,7 +49,7 @@
 
 - name: generate res_def_names
   set_fact:
-    res_def_names: "{{ res_def_names  + [os_resource_name + '-' + inst_c] }}"
+    res_def_names: "{{ res_def_names  + [os_resource_name + 'default_delimiter' + inst_c] }}"
```
samvarankashyap: I don't think you need quotes for default_delimiter variable

linchpin/provision/roles/openstack/tasks/provision_res_defs.yml
skatlapa: yeah, we don't, it's string by default but putting in ' ' is the same, points us that its a custom var. Let's take it out if it's not necessary.

14rcole:
Is there a way to override this in the PinFile?  There is for one of the providers, not sure if its this or libvirt?

skatlapa:
This allows to set the delimiter from linchpin.conf

14rcole:
https://github.com/CentOS-PaaS-SIG/linchpin/pull/986

samvarankashyap:
@14rcole Having delimiter in linchpin.conf would be one place stop for all providers in the future.
However, having delimiter in Pinfile would help and be easier for the end user to override it via command line using template data. I believe we can break it out into a seperate issue since it involves changing the playbooks loop structures for every provider.

skatlapa:
@14rcole thanks for referencing libvirt fix, I did not know about that, so this should work there too. @samvarankashyap Yeah, so I guess we can have this change be universal then. RE Pinfile I think it's a good idea to address in another issue.
Let's have @mcornea take a look as well

linchpin/provision/roles/openstack/tasks/provision_os_object.yml
```diff
@@ -16,7 +16,7 @@
 - name: "provision/deprovision swift containers when count is defined"
   os_object:
     auth: "{{ auth_var | omit_filter(omit) }}"
-    container: "{{ container.1 }}_{{ container.4 }}"
+    container: "{{ container.1 }}default_delimiter{{ container.4 }}"
```
samvarankashyap: @skatlapa According to jinja template syntax the default_delimiter should be in {{  }}
else it will be taken as a string. 
Have you tested this piece?

linchpin/provision/roles/openstack/tasks/provision_os_object.yml
```diff
@@ -40,7 +40,7 @@
 - name: "Async:: provision/deprovision swift containers when count specified"
   os_object:
     auth: "{{ auth_var | omit_filter(omit) }}"
-    container: "{{ container.1 }}_{{ container.4 }}"
+    container: "{{ container.1 }}default_delimiter{{ container.4 }}"
```
samvarankashyap: @skatlapa According to jinja template syntax the default_delimiter should be in {{  }}
else it will be taken as a string. 
Have you tested this piece?

linchpin/provision/roles/openstack/tasks/provision_os_volume.yml
```diff
@@ -18,7 +18,7 @@
     auth: "{{ auth_var | omit_filter(omit) }}"
     state: "{{ vol.3 }}"
     size: "{{ vol.2 }}"
-    display_name: "{{ vol.1 }}_{{ vol.4 }}"
+    display_name: "{{ vol.1 }}default_delimiter{{ vol.4 }}"
```
samvarankashyap: @skatlapa According to jinja template syntax the default_delimiter should be in {{  }}
else it will be taken as a string. 
Have you tested this piece?

linchpin/provision/roles/openstack/tasks/provision_os_volume.yml
```diff
@@ -41,7 +41,7 @@
...
```
samvarankashyap: @skatlapa According to jinja template syntax the default_delimiter should be in {{  }}
else it will be taken as a string. 
Have you tested this piece?

linchpin/provision/roles/openstack/tasks/provision_os_object.yml
skatlapa: I see, I'll check. Its tested only in CI, will confirm locally as well and give an ack.

skatlapa:
It actually worked with the current changes, but let's keep in {{ }}, as currently, it might be taking the string hardcoded, which we will have to check inside those VMs.

samvarankashyap:
[merge]

Issue #990, opened 2019-02-27
Linchpin to support provisioning of VMware VMs 


ohochman:
**Is your feature request related to a problem? Please describe.**

In order to use Linchpin as a cross-team solution: OSP, OCP, CNV, Rhhi.next -  we should have Linchpin to support the provision for all the different teams virtualization platforms.  
VMware is currently missing for the CNV team.  

**Describe the solution you'd like**
VMware provisioning support.

samvarankashyap:
@ohochman: What would be the minimum requirements to be supported in linchpin ?
like vm, network , storage, floating ip ?

RazTamir:
> @ohochman: What would be the minimum requirements to be supported in linchpin ?
> like vm, network , storage, floating ip ?

Hi @samvarankashyap, 
We'll need the same supported requirements we have for AWS or Libvirt with Linchpin.

samvarankashyap:
fixed by #1084

PR #1035, opened 2019-03-20
Esxi on libvirt

abraverm:

PR #1084, opened 2019-04-14
VMware guest support

abraverm:


abraverm:
[test]

abraverm:
fixes  #990

samvarankashyap:
@abraverm Do you have the testing infrastructure for VMware?
Is it possible to add integration tests to VMware provider?

abraverm:
I used the [ESXi on libvirt][https://github.com/CentOS-PaaS-SIG/linchpin/pull/1035] to develop this provider (locally). The integration testing will require deploying ESXi and configuring it on libvirt prior of testing the provider. Its possible but I fear it won't be stable enough... I hope when we get enough traction with this module, we will able to get stable deployment internally and then integration testing would be simple.

14rcole:
This looks great.  If there's anything here to test besides the simple up and destroy that's covered by `general/01_general` in our tests, can you add the integration tests and simply set `providers.include: none`?  That way we'll have the tests available if we need to run them locally

abraverm:
CI is generally failing, not related to the PR

@14rcole, shouldn't it be `providers.exclude: vmware` ?

14rcole:
@abraverm I didn't know if you had anything else you might need to test.  You don't technically need to exclude vmware from `01_general` because the tests won't be run until we add vmware (here)[https://github.com/CentOS-PaaS-SIG/linchpin/blob/develop/config/Dockerfiles/JenkinsfileContainer#L29-L116] but it's not a problem either

abraverm:
awesome, I didn't know that. thanks!

abraverm:
[merge]

PR #1085, opened 2019-04-15
Fix task in dev setup that downloads minishift

14rcole:
Sed task grabbing too much, and was preventing ansible from following links


14rcole:
[merge]

PR #1089, opened 2019-04-16
WIP: test unit test fix

skatlapa:
This patch fails any failures in playbook and also updates jenkins plugin-stage-view to 2.10 which currently is used by cvp and ci-pipeline

### DIFF
--- a/config/s2i/jenkins/master-downstream/plugins.txt
+++ b/config/s2i/jenkins/master-downstream/plugins.txt
@@ -90,7 +90,7 @@ pipeline-model-extensions:1.2.7
 pipeline-rest-api:2.9
 pipeline-stage-step:2.3
 pipeline-stage-tags-metadata:1.2.7
-pipeline-stage-view:2.9
+pipeline-stage-view:2.10
 pipeline-utility-steps:2.0.1
 plain-credentials:1.4
 pubsub-light:1.12
--- a/playbooks/tests/unit_tests.yml
+++ b/playbooks/tests/unit_tests.yml
@@ -1,6 +1,7 @@
 ---
 - name: Unit tests for Linchpin
   hosts: localhost
+  any_errors_fatal: true
   gather_facts: False
   tasks:
           - name: Install Test Reqs


config/s2i/jenkins/master-downstream/plugins.txt
```diff
@@ -90,7 +90,7 @@ pipeline-model-extensions:1.2.7
 pipeline-rest-api:2.9
 pipeline-stage-step:2.3
 pipeline-stage-tags-metadata:1.2.7
-pipeline-stage-view:2.9
+pipeline-stage-view:2.10
```
johnbieren: This is not even the plugins.txt for the Jenkins master that the unit tests run on. Furthermore, the master the unit tests *are* running on has pipeline-stage-view 2.10. This is not the fix

skatlapa: where are the unit tests being run?

johnbieren: https://jenkins-continuous-infra.apps.ci.centos.org/view/Linchpin/job/ci-linchpin-unittests/

skatlapa: I mean the s2i for it :), but nvm we are good I guess, thanks!

skatlapa:
Being addressed here -https://github.com/openshift/contra-hdsl/issues/87

config/s2i/jenkins/master-downstream/plugins.txt
johnbieren: Well, the unit tests are run on the upstream master, like I posted, so if you go to the s2i for the upstream master and go to jobs, you find https://github.com/CentOS-PaaS-SIG/linchpin/blob/develop/config/s2i/jenkins/master-upstream/configuration/jobs/ci-linchpin-unittests/config.xml

skatlapa: ahh, got it, thnx!

PR #1091, opened 2019-04-16
Install GCC in unit test container

14rcole:


samvarankashyap:
@14rcole : Please fix the indentation. 
Furthermore, If gcc is not available there is a high chance that python-devel might not be available too. 
Could you add that too ?

p3ck:
we may have to re-think the way the unittests are running.

14rcole:
@p3ck is there any particular reason that we use the `continuous-infra/jenkins` container to run the unit tests?  Do we really need jenkins for anything?

samvarankashyap:
@p3ck Is there any way we could run it inside a custom container rather than a Jenkins container ??

p3ck:
> @p3ck Is there any way we could run it inside a custom container rather than a Jenkins container ??

HDSL uses a tool called linchpin to do it's provisioning.  If that supports provisioning containers then yes. ;)

14rcole:
@samvarankashyap The container is defined here, we can just change that
https://github.com/CentOS-PaaS-SIG/linchpin/blob/develop/config/Dockerfiles/JenkinsfileUnitTests#L5

*but* it appears to be pulling from an openshift registry somewhere so we'll have to see what images are available in that registry

samvarankashyap:
[test]

playbooks/tests/unit_tests.yml
```diff
@@ -3,7 +3,16 @@
...
-          - name: Install Test Reqs
+          - name: Install yum Test Reqs
+            yum:
+                    name: "{{ packages }}"
+                    state: present
+            vars:
+                    packages:
+                          - gcc
+                          - python-devel
+            become: yes
```
samvarankashyap: @14rcole I dont think you need sudo here if we are running as root user in the container

config/Dockerfiles/JenkinsfileUnitTests
```diff
@@ -1,9 +1,9 @@
 // Define the openshift pod name, docker repo url, namespace, and service acct
 // for the DSL pod template.
 dslPodName = "contraDsl-${UUID.randomUUID()}"
-dockerRepoURL = env.DOCKER_REPO_URL ?: '172.30.254.79:5000'
-openshiftNamespace = env.OPENSHIFT_NAMESPACE ?: 'continuous-infra'
-openshiftServiceAccount = env.OPENSHIFT_SERVICE_ACCOUNT ?: 'jenkins'
+dockerRepoURL = env.DOCKER_REPO_URL ?: ''
+openshiftNamespace = env.OPENSHIFT_NAMESPACE ?: 'docker.io'
+openshiftServiceAccount = env.OPENSHIFT_SERVICE_ACCOUNT ?: 'fedora:29'
```
samvarankashyap: @14rcole Is there a service account named fedora:29 ???

samvarankashyap: @14rcole Further there is no namespace named docker.io , it should be continous-infra and docker repo url would be docker.io i guess

14rcole: The environment variables are poorly named, but we end up doing a pull of `docker://dockerRepoURL/openshiftNamespace/openshiftServiceAccount` e.g. `docker://172.35.15.12/continuous-infra/jenkins:latest`

14rcole:
[test]

14rcole:
[test]

14rcole:
[test]

14rcole:
The solution ended up being to remove pycrypto.  It's the only package that requires a c compiler but we don't seem to use it anywhere

samvarankashyap:
@14rcole pycrypto used to be requirement for ansible 2.0 to 2.3 it was replaced by pycryptodome.
I don't think we need it anymore.

14rcole:
[test]

14rcole:
[test]

14rcole:
[test]

14rcole:
[test]

14rcole:
[test]

samvarankashyap:
@14rcole you may merge it

14rcole:
[merge]

PR #1119, opened 2019-04-29
Fix bug preventing releases from being created

14rcole:
We were passing "1.7.4" to the script, while the release was tagged as "v1.7.4"

### DIFF
--- a/config/Dockerfiles/JenkinsfileRelease
+++ b/config/Dockerfiles/JenkinsfileRelease
@@ -217,7 +217,7 @@ EOF
                     }
                     stage('create GitHub release') {
                         dir(repoName) {
-                            string: version = sh (
+                            string version = "v" + sh (
                                 script: "python setup.py --version",
                                 returnStdout: true
                             ).trim()


config/Dockerfiles/JenkinsfileRelease
```diff
@@ -216,9 +216,9 @@ EOF
                         }
                     }
                     stage('create GitHub release') {
-                        dir(repoName) {
-                            string: version = sh (
-                                script: "python setup.py --version",
+                        dir(reponame) {
+                            string version = "v" + sh (
+                                script python setup.py --version,
```
samvarankashyap: it should be 
```
script: "python setup.py --version"
```

config/Dockerfiles/JenkinsfileRelease
```diff
@@ -216,9 +216,9 @@ EOF
                         }
                     }
                     stage('create GitHub release') {
-                        dir(repoName) {
-                            string: version = sh (
-                                script: "python setup.py --version",
+                        dir(reponame) {
```
samvarankashyap: should be repoName 
else it would create errors

config/Dockerfiles/JenkinsfileRelease
samvarankashyap: @14rcole The following should work 
```
dir(reponame) {
                            string version= sh (
                                script: "python setup.py --version",
                                returnStdout: true
                                ).trim()
                            version = "v"+version
                            withCredentials([string(credentialsId: 'github_api_token', variable: 'TOKEN')]) {
                                pipelineUtils.executeInContainer(currentStage, "fedora29", "python ./config/Dockerfiles/release-notes.py "+ TOKEN + " " + version)
                            }
              }
```

samvarankashyap:
[merge]

PR #1127, opened 2019-05-03
Fix release

skatlapa:


config/Dockerfiles/release-notes.py
```diff
@@ -63,18 +63,15 @@ def get_remaining_changes(pulls):
...
+                continue
+                changes.append(pulls[i])
+                pulls.pop(i)
                 break
-        if not skip:
-            changes.append(pulls[i])
-            pulls.pop(i)
     return changes
 
-
```
14rcole: Nit: we should be leaving two blank lines between functions

config/Dockerfiles/release-notes.py
```diff
@@ -63,18 +63,15 @@ def get_remaining_changes(pulls):
     changes = []
     # iterate over pulls backward so that list can be removed in place
     for i in xrange(len(pulls) - 1, -1, -1):
-        skip = False
         labels = pulls[i].get_labels()
         for label in labels:
             if label.name == "release":
-                skip = True
+                continue
```
14rcole: won't this skip the inner loop?  Also, will the two lines below this ever run?

14rcole: I see this is outdated, you can ignore it

skatlapa: I just changed while you were commenting :)

samvarankashyap:
@skatlapa could you squash the changes into single commit, please.

skatlapa:
@14rcole btw.. would we have more than one label that the other functions don't catch? If yes, we could remove break

14rcole:
@skatlapa yes, but I don't see how that enables us to remove break

skatlapa:
@14rcole break would only catch one flag that isn't 'release', if we need to catch other flags too then we need to remove break. Also, I think we should probably adjust the function to not to return empty stack(changes) or return 'None'.

skatlapa:
Let's merge after this is successful - https://jenkins-continuous-infra.apps.ci.centos.org/view/Linchpin/job/ci-linchpin/1433/ But looks like Travis build failed, looking more

skatlapa:
[test]

skatlapa:
@samvarankashyap @14rcole do you know why Travis failed? Jenkins is good and codewise everything looks good.

14rcole:
@skatlapa it looks like travis passed

skatlapa:
@14rcole there was an unused var that would've to cause some linting/pyflake error, removed, and rest lgtm. I'll merge once the Jenkins build completes

skatlapa:
[test]

skatlapa:
[merge]

Issue #1134, opened 2019-05-16
Linchpin execution fails with Error executing hook: 'can't set attribute


mcornea:
```shell
(rhhi-venv-TodCbu3) [root@sealusa35 linchpin_workspace]# linchpin --pinfile parsedPinFile --template-data @hooks/ansible/rhhi-setup/extravars.yaml -v destroy
/home/rhhi-ci/jenkins/workspace/rhhi.next-virt-customized/rhhi-venv-TodCbu3/lib/python2.7/site-packages/linchpin/utils/dataparser.py:99: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
  data = yaml.load(data)
-------
start hook ansible:rhhi-setup
Error executing hook: 'can't set attribute'
```

The issue is fixed after downgrading ansible:

```shell
(rhhi-venv-TodCbu3) [root@sealusa35 linchpin_workspace]# pip install ansible==2.7.1
DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7.
Collecting ansible==2.7.1
Requirement already satisfied: setuptools in /home/rhhi-ci/jenkins/workspace/rhhi.next-virt-customized/rhhi-venv-TodCbu3/lib/python2.7/site-packages (from ansible==2.7.1) (41.0.1)
Requirement already satisfied: PyYAML in /home/rhhi-ci/jenkins/workspace/rhhi.next-virt-customized/rhhi-venv-TodCbu3/lib/python2.7/site-packages (from ansible==2.7.1) (5.1)
Requirement already satisfied: jinja2 in /home/rhhi-ci/jenkins/workspace/rhhi.next-virt-customized/rhhi-venv-TodCbu3/lib/python2.7/site-packages (from ansible==2.7.1) (2.10.1)
Collecting paramiko (from ansible==2.7.1)
  Using cached https://files.pythonhosted.org/packages/cf/ae/94e70d49044ccc234bfdba20114fa947d7ba6eb68a2e452d89b920e62227/paramiko-2.4.2-py2.py3-none-any.whl
[...]

PR #1154, opened 2019-06-07
Update linchpin to be compatible with newer version of ansible

14rcole:
As it turns out, Ansible best practice is to symlink `filter_plugins` into every `roles` dir, so we'll just have to deal with a bit more messiness now


linchpin/ansible_runner.py
```diff
@@ -50,7 +52,7 @@ def suppress_stdout():
 
 def ansible_runner_2x(playbook_path,
                       extra_vars,
-                      options=None,
```
samvarankashyap: @14rcole Should this be removed in 2.x ? runner too ?

linchpin/ansible_runner.py
```diff
@@ -156,12 +181,19 @@ def ansible_runner(playbook_path,
...
                       sftp_extra_args=None,
                       scp_extra_args=None,
+                      start_at_task=None,
                       verbosity=verbosity,
                       check=False,
                       diff=False
                       )
 
-    if ansible24:
+    if ansible_version >= 2.8:
```
samvarankashyap: @14rcole : Are you sure its >=2.8? I remember you saying there was a problem with 2.7.10 version?

requirements.txt
```diff
@@ -17,7 +17,7 @@ urllib3>=1.23,<1.25
 PyYAML>=5.1
 jinja2>=2.10
 configparser>=3.5.0
-openshift>=0.8.6
+openshift>=0.8.6,<0.8.8
```
samvarankashyap: @14rcole Any reason for the changing openshift version ? 
what was the error/bug?

linchpin/ansible_runner.py
14rcole: The problem with the 2.7.10 version and the problem with 2.8 are the same.  However, I had trouble reproducing the issue with 2.7.10 and I could consistently reproduce it with 2.8

requirements.txt
14rcole: There's another package that requires urllib <1.25, and openshift 0.8.8 requires urllib>=1.25.  I forget exactly which package.  I'll figure it out and make a note of it so we know when we can update openshift

requirements.txt
```diff
@@ -5,7 +5,7 @@ boto3>=1.9.96
 apache-libcloud>=0.20.1
 click>=7.0
 yamlordereddictloader>=0.4.0
-ansible>=2.7.1,<2.7.10
```
samvarankashyap: @14rcole 
Assuming With this current fix linchpin will work with all the versions >=2.7.1 
The requirements should be ansible>=2.7.1 since the adoption of ansible 2.8 is still low.

linchpin/ansible_runner.py
14rcole: No, you're right.  That's leftover from troubleshooting

requirements.txt
14rcole: Update: it looks like we require urllib<1.25.  So I can look into updating that as well at some point

14rcole:
Fixes #1154

14rcole:
[test]

14rcole:
This fixes #1134, but is dependent on #1161

samvarankashyap:
Waiting on tests to pass

samvarankashyap:
@14rcole Any update on this PR ?

14rcole:
@samvarankashyap yes, it doesn't work on CentOS because of the version of Python they use.  I'm working to fix it now in addition to #1158

14rcole:
[test]

14rcole:
[test]

14rcole:
@samvarankashyap random failure on gcp followed by random failure on aws.  They've both passed in the last 6 hours.  Good to merge?

14rcole:
[test]

14rcole:
[merge]

PR #1170, opened 2019-06-27
Cancel beaker job when there is a timeout while fetching the resources

### LINKED ISSUE #1153: Beaker job not cancelled if up is killed
**Describe the bug**
If you do a linchpin up to provision a machine in Beaker and that up command is killed, the Beaker job is left running. However, you will get no inventory back to work with it (which is probably expected). The bug is that even with the rundb, you cannot tear it down. Calling linchpin destroy with the resulting rundb will do nothing, so when using automation, the beaker job is just left alive until someone goes and manually kills it.

**To Reproduce**
1. Create a Jenkins job with say a 1h timeout
2. Create a pinfile with a beaker target that has a timeout >1h
3. Run the jenkins job that does a linchpin up
4. Let the job timeout
5. Even if you have it in a try catch with linchpin destroy, the beaker job is not cancelled

**Expected behavior**
The beaker job is cancelled

samvarankashyap:
fixes #1153 

Summary:
- added fetch_beaker_ids filter
- added unit test to fetch_beaker_ids
- Updated playbooks to handle timed out beaker resources

### DIFF
--- a/linchpin/FilterUtils/FilterUtils.py
+++ b/linchpin/FilterUtils/FilterUtils.py
@@ -260,3 +260,13 @@ def transform_os_server_output(res_def_out):
         res_def["openstack"].append(ele.get("openstack", {}))
         res_def["servers"].append(ele.get("server", {}))
     return res_def
+
+
+def fetch_beaker_job_ids(topo_out):
+    output = []
+    for entry in topo_out:
+        entry_dict = {}
+        entry_dict["ids"] = []
+        entry_dict["ids"].append("J:" + entry["id"])
+        output.append(entry_dict)
+    return output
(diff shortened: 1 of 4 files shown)


johnbieren:
Looks fine to me, but I don't know the code in this PR as well as the others

14rcole:
@samvarankashyap when the cancel message is sent, does ansible wait for beaker to be finished cancelling the job or does it just send the message and continue?  If it's the former, is there any way to make this nonblocking?  It would seem silly to hit a timeout then have the user wait another 5-10 minutes because the job had  to be cancelled

samvarankashyap:
> @samvarankashyap when the cancel message is sent, does ansible wait for beaker to be finished cancelling the job or does it just send the message and continue? If it's the former, is there any way to make this nonblocking? It would seem silly to hit a timeout then have the user wait another 5-10 minutes because the job had to be canceled.

@14rcole As far as my testing. Cancel message are instantaneous. I didn't face any blocking wait while the job is canceled. 
According to beaker module, we relay the beaker message and return.

samvarankashyap:
[merge]

PR #1175, opened 2019-07-02
Add option to disable uhash for specific targets via the command line

### LINKED ISSUE #797: Disable uhash for specific provider over command line.
Currently uhash is enable throughout the topology 
There is a recent request to disable uhash for only certain providers at commandline/conf level. 
forexample : 
If a topology contains aws, openstack, beaker 
user should be able to disable uhash for aws and beaker over commandline/conf 
```
linchpin -vvv up --disable-uhash aws beaker 
```

14rcole:
Resolves #797

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,6 @@ install:
   - pip install .[tests]
 
 script:
-  - flake8 --exclude=\.eggs,tests,docs --ignore=E124,E303,W504 --max-line-length 80 .
+  - flake8 --exclude=\.eggs,tests,docs,config/Dockerfiles --ignore=E124,E303,W504 --max-line-length 80 .
   - $(which python) setup.py test
 #  - $(which python) $(which nosetests) -s linchpin/tests/*
--- a/docs/source/advanced/layouts.rst
+++ b/docs/source/advanced/layouts.rst
@@ -9,8 +9,11 @@ New in version 1.5.2
 
 When an :term:`layout` is provided in the PinFile, LinchPin automatically generates
 a static inventory for Ansible. The inventory filename is dynamically generated based
-upon a few factors. However, the value can be overridden simply by adding the
-``inventory_file`` option.
+upon the name of the target and the uhash.  However, the value can be overridden
+simply by adding the ``inventory_file`` option.  The uhash can be disabled for all targets
+by setting the ``enable_uhash`` flag to False in ``linchpin.conf`` or for a subset of
+targets by using the ``--disable-uhash`` flag when running ``linchpin up`` and providing
+a comma-separated list of targets
 
 .. code-block:: bash
 
--- a/docs/source/includes/provisioning.rst
+++ b/docs/source/includes/provisioning.rst
@@ -8,7 +8,8 @@ Once a :term:`PinFile`, :term:`topology`, and optional :term:`layout` are in pla
 
     Target              Run ID  uHash       Exit Code
     -------------------------------------------------
-    dummy_cluster       70      0446                0
+    dummy-new           83      a18e9a	            0
+    dummy-topo          70      044695              0
 
 As you can see, the generated inventory file has the right data. This can be used in many ways, which will be covered elsewhere in the documentation.
 
@@ -29,4 +30,12 @@ To verify resources with the dummy cluster, check ``/tmp/dummy.hosts``
     web-0446-0.example.net
     test-0446-0.example.net
 
+A subset of the hosts in a PinFile can be provisioned by listing each of them at the end of the command
 
+.. code-block:: bash
+
+    $ linchin -vv up dummy-new
+
+    Target              Run ID  uHash       Exit Code
+    -------------------------------------------------
+    dummy-new           83      a18e9a	            0
--- a/linchpin/__init__.py
+++ b/linchpin/__init__.py
@@ -473,16 +473,19 @@ class LinchpinAPI(object):
                                            'vault_pass',
                                            default=''))
 
-        for target in list(provision_data.keys()):
+        targets = [x.lower() for x in list(provision_data.keys())]
+        disable_uhash_targets = self.ctx.get_evar('disable_uhash_targets')
+        enable_uhash = self.ctx.get_evar('enable_uhash')
+
+        for target in targets:
             if not isinstance(provision_data[target], dict):
                 raise LinchpinError("Target '{0}' does not"
                                     " exist.".format(target))
 
-        targets = [x.lower() for x in list(provision_data.keys())]
         if 'linchpin' in targets:
             raise LinchpinError("Target 'linchpin' is not allowed.")
 
-        for target in list(provision_data.keys()):
+        for target in targets:
             if target == 'cfgs':
                 continue
 
@@ -615,6 +618,12 @@ class LinchpinAPI(object):
                 self.hook_state = '{0}{1}'.format('pre', action)
 
             # FIXME need to add rundb data for hooks results
+
+            # Enable/disable uhash based on provided flag(s)
+            if disable_uhash_targets and target in disable_uhash_targets:
+                self.ctx.set_evar('use_uhash', False)
+            else:
+                self.ctx.set_evar('use_uhash', enable_uhash)
 
             # invoke the appropriate action
             return_code, results[target]['task_results'] = (
--- a/linchpin/ansible_runner.py
+++ b/linchpin/ansible_runner.py
@@ -89,8 +89,8 @@ def ansible_runner_24x(playbook_path,
     pbex = PlaybookExecutor([playbook_path],
                             inventory,
                             variable_manager,
-                            options,
                             loader,
+                            options,
                             passwords)
     return pbex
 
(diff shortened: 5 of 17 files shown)


jaypoulz:
I've tested this out in a local env, and I'm not getting it to work. What happens is that dummy-topo-2 is never brought up, only 1 and 3.
```
linchpin --config linchpin.config --pinfile workspaces/dummy/PinFile up dummy-topo --disable-uhash dummy-topo-2 dummy-topo-3
```
linchpin.config
```
[evars]
enable_uhash = True
```
topologies
```
dummy-topo:
  topology:
    topology_name: "dummy_cluster" # topology name
    resource_groups:
        resource_group_type: "dummy"
        resource_definitions:
          - name: "{{ distro | default('') }}web-1"
            role: "dummy_node"
            count: 3
          - name: "{{ distro | default('') }}test-1"
            role: "dummy_node"
            count: 1
  layout:
    inventory_layout:
      vars:
        hostname: __IP__
      hosts:
        example-node:
          count: 3
          host_groups:
            - example
        test-node:
          count: 1
          host_groups:
            - test
dummy-topo-2:
  topology:
    topology_name: "dummy_cluster" # topology name
    resource_groups:
        resource_group_type: "dummy"
        resource_definitions:
          - name: "{{ distro | default('') }}web-2"
            role: "dummy_node"
            count: 3
          - name: "{{ distro | default('') }}test-2"
            role: "dummy_node"
            count: 1
  layout:
    inventory_layout:
      vars:
        hostname: __IP__
      hosts:
        example-node:
          count: 3
          host_groups:
            - example
        test-node:
          count: 1
          host_groups:
            - test

dummy-topo-3:
  topology:
    topology_name: "dummy_cluster" # topology name
    resource_groups:
        resource_group_type: "dummy"
        resource_definitions:
          - name: "{{ distro | default('') }}web-3"
            role: "dummy_node"
            count: 3
          - name: "{{ distro | default('') }}test-3"
            role: "dummy_node"
            count: 1
  layout:
    inventory_layout:
      vars:
        hostname: __IP__
      hosts:
        example-node:
          count: 3
          host_groups:
            - example
        test-node:
          count: 1
          host_groups:
            - test
```

14rcole:
They're both being provisioned.  But the inventory file is named after the topology, not the target.  So the first target is being provisioned and writing to `dummy_cluster.inventory` and then the second target is being provisioned and then overwriting it.

samvarankashyap:
@14rcole @jaypoulz 
I think we should change the inventory to be named after target since targets are unique and topologies can be reused when we use file-based references. 
or we can add target as a prefix to the inventory we generate making it unique. 
@jaypoulz anythoughts?

14rcole:
You can also overwrite the name of the inventory file in the layout

jaypoulz:
> They're both being provisioned. But the inventory file is named after the topology, not the target. So the first target is being provisioned and writing to `dummy_cluster.inventory` and then the second target is being provisioned and then overwriting it.

It is true that the names conflict, but that is not the bug that is present. If the command were working as intended, we would end up with 3 inventory files. Both inventories for topos 1 and 3 would have uhashes in the inventory file name, and topology 2 would be output as  dummy_cluster.inventory. It's not being overridden - it's never being provisioned.

This is the output of the command above - topo 2 isn't mentioned at all.
```
 [WARNING]: Unable to parse /home/jpoulin/Projects/scratch/linchpin/localhost
as an inventory source

 [WARNING]: No inventory was parsed, only implicit localhost is available

Action 'up' on Target 'dummy-topo' is complete
Action 'up' on Target 'dummy-topo-3' is complete

ID: 11
Action: up

dummy-topo-3        	     9	7a50ff	        0
dummy-topo          	    12	44b45c	        0
```

jaypoulz:
Also, I mentioned this before, but specifying a single target with `--disable-uhash` provisions all targets in the provided PinFile, which the same behavior as if you never provided a target.

jaypoulz:
@samvarankashyap I do think it makes sense to use the targets as the inventory names; however, it would be a breaking change so it would have to handled cautiously. I think being able to override the name is probably sufficient, which is already a feature. It prevent the confusion of overriding your inventory if you have the same topology name, but the real issue there is that we can't turn on uhash by default until a linchpin 2.x.x is released.

jaypoulz:
I updated the topology names to dummy_cluster-1, -2, and -3 respectively.
The command I provided above results in two inventory files.
```
dummy_cluster-1-5544f4.inventory  dummy_cluster-3-c586e7.inventory
```

14rcole:
@jaypoulz hmm, I was able to provision a whole PinFile and only disable the uhash for some targets - that would be the expected  behavior, right?

When I run `linchpin up --disable-uhash dummy-new` on the PinFile in `docs/source/examples/workspaces/dummy` I get the this:
```
(linchpin) [ryan@localhost dummy]$ ls inventories/
dummy-new-96d8d1.inventory.inventory  dummy_topo.inventory
```

I changed the name of the topologies just to make it clear which is which.  I don't know why one of your resources isn't being provisioned.

14rcole:
I see what the problem is above.  You have to provide the `--disable-uhash` flag once for each item which you want to disable.  Otherwise the syntax here is ambiguous:

`linchpin --config linchpin.config --pinfile workspaces/dummy/PinFile up dummy-topo --disable-uhash dummy-topo-2 dummy-topo-3`

It could mean "provision dummy-topo-3 and disable uhash for dummy-topo-2" or "provision everything but disable the uhash for dummy-topo-2 and dummy-topo-3".  Instead, try this:

`linchpin --config linchpin.config --pinfile workspaces/dummy/PinFile up dummy-topo --disable-uhash dummy-topo-2 --disable-uhash dummy-topo-3`

I know its clunky but it gets around the syntactic ambiguity.  Also, that seems to be the prefered way for the click module

jaypoulz:
> @jaypoulz hmm, I was able to provision a whole PinFile and only disable the uhash for some targets - that would be the expected behavior, right?

Ah. This is the source of the confusion. That is not the expected behavior. The expected behavoir would by to provision only the target specified with uhash disabled.

> It could mean "provision dummy-topo-3 and disable uhash for dummy-topo-2" or "provision everything but disable the uhash for dummy-topo-2 and dummy-topo-3". Instead, try this:
I figured the flag applied to the target next to it. I expected this to function like:
```
linchpin --config linchpin.config --pinfile workspaces/dummy/PinFile up --enable-uhash dummy-topo --disable-uhash dummy-topo-2 --enable-uhash dummy-topo-3
```
In this version, 1 and 3 would have uhashes enabled, and 2 would not. I figured that omitting the overriden disable would leave it on by default.

samvarankashyap:
@14rcole you need to add documentation around --disable-uhash flag

jaypoulz:
Thought about this some more over lunch.

Targets after the command (up, destroy) filter out the specific targets from the PinFile that you want to perform the action on.

For that reason, if you did want to disable configuration for a single target (or list of targets) but run the command against all targets in the PinFile, the syntax should look like this:
```linchpin --disable-uhash filteredtarget1 filteredtarget2 up [optional target(s)]```
```linchpin [-o/--options] command [optional target(s)]```

**Comment by 14rcole** [comment]
@jaypoulz My only concern about putting the flag before the command is that the flag is specific to `linchpin up`.

Also, because of the way that the click module functions, this command:
```
linchpin --disable-uhash filteredtarget1 filteredtarget2 up
```
Will be interpreted as "ignore the uhash for the targets filteredtarget1, filteredtarget2, and up" instead of "run all targets and ignore the uhash for filteredtarget1 and filteredtarget2"

**Comment by jaypoulz** [comment]
> @jaypoulz My only concern about putting the flag before the command is that the flag is specific to `linchpin up`.

Yes, I understand, I was just explaining that my expectation for the hierarchy of CLI args is somewhat informed by how the cucumber rubygem and other CLI tools implement this sort of thing.

> Also, because of the way that the click module functions, this command:
> 
> ```
> linchpin --disable-uhash filteredtarget1 filteredtarget2 up
> ```
> Will be interpreted as "ignore the uhash for the targets filteredtarget1, filteredtarget2, and up" instead of "run all targets and ignore the uhash for filteredtarget1 and filteredtarget2"

There are ways to make that syntax unambiguous -- e.g. ``--disable-uhash=filteredtarget1,filteredtarget2 up``, but like I said, I would prefer target configuration flags that configure the target that follows immediately after, or the list of targets that follow after.

I don't think it matters which you choose (1 target, or all following), as long as it's clear from the docs.
If you choose to have the configuration flag apply to all following targets, I suggest that a `--default-uhash` is added so I can unset the flag somewhere mid-list.

**Comment by samvarankashyap** [comment]
@14rcole I think comma separated ars for disable-uhash seems feasible.

**Comment by 14rcole** [comment]
@jaypoulz I can't do the target configuration flags like you suggested because click doesn't support having multiple flags with variadic arguments.  However, I made it so that `--disable-uhash` supports a comma-separated list of targets so that you only have to supply the flag once.  You will have to list the targets you want to provision separately though (including those for which the uhash is disabled)

**Comment by jaypoulz** [comment]
OK so just to be clear, this is my current understanding of how this feature is expected to work.

### All Targets

To run all targets but disable uhash for specific ones, run like this:
```
linchpin up --disable-uhash targetA,targetC
```

Expected behavior would be that A B and C are provisioned, but A and C have uhash disabled.

### Specified Targets

To run specific targets, but disable uhash for specific ones, run like this:
```
linchpin up --disable-uhash targetA,targetC targetA targetB targetC
```
Expected bevahoir would be that A B and C are provisioned, but A and C have uhash disabled.

14rcole:
Yes, that's correct

14rcole:
[test]

14rcole:
[merge]

PR #1184, opened 2019-07-09
Fix bugs causing unit tests to fail in python 3

14rcole:
There were two failing tests: The first failed because the hosts within a host group were listed in different orders.  This was resolved by switching the inventory formatters to OrderedDicts instead of dicts.  The second was the test that verified the order of the generated inventory file as a whole, which failed because the host groups were consistently being printed in different orders.  Seeing as this test is not necessary and the failure seems to be the result of the configparser module, this test was removed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ language: python
 
 python:
   - "2.7"
-  - "3.6"
+  - "3.7"
 
 cache: pip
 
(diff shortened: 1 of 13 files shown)


14rcole:
[test]

samvarankashyap:
@14rcole needs rebase

14rcole:
[merge]

Issue #1194, opened 2019-07-11
Container releases are broken in release job of linchpin


samvarankashyap:
**Describe the bug**
When merging release PR it was found that the container releases was broken

error trace:
```
11:28:42  Running in /workDir/workspace/cd-linchpin-release/linchpin
[Pipeline] {
[Pipeline] sh
11:28:42  + python setup.py --version
[Pipeline] withCredentials
11:28:50  Masking only exact matches of $DOCKER_CREDS
[Pipeline] {
[Pipeline] sh
11:28:51  + mkdir -p test-pypi-verify
[Pipeline] withEnv
[Pipeline] {
[Pipeline] container
[Pipeline] {
[Pipeline] sh
11:28:52  + ./config/Dockerfiles/buildah-push.sh 1.7.6.1
11:28:52  + IMAGE=docker.io/contrainfra/linchpin
11:28:52  + TAG=1.7.6.1
11:28:52  + TAGGED_IMAGE=docker.io/contrainfra/linchpin:1.7.6.1
11:28:52  + buildah bud -t docker.io/contrainfra/linchpin config/Dockerfiles/linchpin
11:28:52  failed to mount overlay for metacopy check: invalid argument
11:28:52  + buildah push --authfile **** docker.io/contrainfra/linchpin
11:28:52  failed to mount overlay for metacopy check: invalid argument
11:28:52  + buildah tag docker.io/contrainfra/linchpin docker.io/contrainfra/linchpin:1.7.6.1
11:28:52  failed to mount overlay for metacopy check: invalid argument
11:28:52  + buildah push --authfile **** docker.io/contrainfra/linchpin
11:28:52  failed to mount overlay for metacopy check: invalid argument
[Pipeline] }
[Pipeline] // container
[Pipeline] }
[Pipeline] // withEnv
[Pipeline] sh
11:28:53  + '[' -d logs ']'
11:28:53  + echo 'No logs for executeInContainer(). Ignoring this.'
```

14rcole:
Based on https://github.com/containers/libpod/issues/3571, it looks like the issue here is because of the host kernel.  Is there a way I can make changes to this script and test it without triggering a new release?

samvarankashyap:
@14rcole you can comment out the part where we push it to pypi and trigger the release PR On develop branch. 
I see thats the only way now .

14rcole:
Fixed by #1250

PR #1196, opened 2019-07-11
disable container releases until it is fixed

samvarankashyap:
Disable container releases until its fixed.
releated to #1194


samvarankashyap:
[merge]

PR #1231, opened 2019-07-24
Freeze GitPython version

14rcole:
GitPython 2.1.12 is only compatible with Python 3.  This freezes the version at 2.1.11 until Python 2 support is dropped


samvarankashyap:
@14rcole Shouldn't this be <=2.1.11

14rcole:
@samvarankashyap It could be, but we first started using gitpython with version 2.1.11 so I don't know if there's anything incompatible in previous versions

14rcole:
[merge]

PR #1249, opened 2019-07-31
Enable passing environment vars and run ansible_runner as subprocess

### LINKED ISSUE #1125: Ansible options overwrite with Linchpin API
[Ansible configurations are currently hard coded](https://github.com/CentOS-PaaS-SIG/linchpin/blob/696d7a08b6a58c1b147432203fddc6d5c55ce0e7/linchpin/ansible_runner.py#L122-L162) and the only way to overwrite them is with `ansible.cfg` file. Linchpin API should allow overwrite Ansible configurations without using `ansible.cfg` but programmatically. This is needed for custom configurations such as `log_path`.

samvarankashyap:
Fixes #1125 

This is a preview feature which enables running of ansible_runner as subprocess shell
instead of running ansible API. 

example usage:
```
linchpin -vvvvv up --use-shell --env-vars TESTENV envvalue
```

### DIFF
--- a/docs/source/includes/provisioning.rst
+++ b/docs/source/includes/provisioning.rst
@@ -39,3 +39,14 @@ A subset of the hosts in a PinFile can be provisioned by listing each of them at
     Target              Run ID  uHash       Exit Code
     -------------------------------------------------
     dummy-new           83      a18e9a	            0
+
+
+
+Preview Feature:
+
+linchpin up and destroy includes --use-shell parameter which makes linchpin run as a subprocess rather than ansible api call
+usefull when we would like to overwrite environment varibles
+
+.. code-block:: bash
+
+    $ linchpin -vvvv up dummy-new --use-shell --env-vars TESTENV testenv value
--- a/config/Dockerfiles/tests.d/dummy/10_dummy-use-shell
+++ b/config/Dockerfiles/tests.d/dummy/10_dummy-use-shell
@@ -14,7 +14,7 @@ TARGET="dummy_target"
 
 function clean_up {
     set +e
-    linchpin -w . -p "${PINFILE}" -v destroy "${TARGET}"
+    linchpin -w . -p "${PINFILE}" -v destroy "${TARGET}" --use-shell
     D_RC=0
     D_RC=(${?} -o ${D_RC})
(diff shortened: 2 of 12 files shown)


docs/source/includes/provisioning.rst
```diff
@@ -39,3 +39,14 @@ A subset of the hosts in a PinFile can be provisioned by listing each of them at
     Target              Run ID  uHash       Exit Code
     -------------------------------------------------
     dummy-new           83      a18e9a	            0
+
+
+
+Preview Feature:
+
+linchpin up and destroy includes --use-shell parameter which makes linchpin run as a subprocess rather than ansible api call
```
14rcole: Given what we discussed yesterday about modularization in the linchpin devel meeting, should we make this a flag?  If we make it the default and move to ansible galaxy later, it won't be a breaking change.  This implementation would.

samvarankashyap: @14rcole Currently --use-shell is a flag in click module and is only enabled when used. If not used in command line it would take the ansible API route by default.
I think I might have misquoted in the documentation. 
If --use-shell is used linchpin runs ansible as a subprocess. 
Even if we move towards ansible-galaxy am not sure why would it break.

samvarankashyap:
[test]

docs/source/includes/provisioning.rst
14rcole: If we move toward ansible galaxy, wouldn't be have to use a subprocess?  if so, using a subprocess should be the default

samvarankashyap:
[test]

samvarankashyap:
Failing on JSON inventory. 
Trying to troubleshoot the cause.

docs/source/includes/provisioning.rst
samvarankashyap: I totally agree with you. Subprocess should be default, However there are many features we might have to make sure we implement which might not be in time for 1.8 release. 
That's the reason we are pushing out the preview with a flag. 
Enduser can default it to subprocess through linchpin.conf if they would like to.

samvarankashyap:
[test]

samvarankashyap:
@14rcole Needs review

samvarankashyap:
[merge]

PR #1255, opened 2019-08-05
Fix missing system packages in RPM setup

abraverm:


abraverm:
@samvarankashyap, also we need to configure a webhook that will trigger copr building.

abraverm:
The fix was verified to work.

Issue #1101, opened 2019-04-22
RFE: Collect Beaker boot logs


abraverm:
Subcase of #868 for Beaker - create a post hook that will collect [logs of a recipes](https://beaker-project.org/docs/server-api/http.html#get--recipes-(int-id)-logs-(path-path))

PR #1289, opened 2019-08-21
Beaker log collector hook

abraverm:
Implements #1101


abraverm:
Agree, and I think we should start pulling all these comments into a document :)

Maybe we can add the built-in hooks as "addons" to future galaxy roles..

samvarankashyap:
[merge]

Issue #1290, opened 2019-08-21
Libvirt provider adds the private key to ~/.ssh/authorized_keys instead of public one


mcornea:
**Describe the bug**
Libvirt provider adds the private key to ~/.ssh/authorized_keys instead of public one:

https://github.com/CentOS-PaaS-SIG/linchpin/blob/develop/linchpin/provision/roles/libvirt/tasks/provision_libvirt_node.yml#L72-L98

PR #1291, opened 2019-08-21
Append public key to ~/.ssh/authorized_keys in libvirt provider

mcornea:
Currently the private key gets appended to ~/.ssh/authorized_keys which is wrong. This change adjusts the tasks so they append the public key to ~/.ssh/authorized_keys.

Fixes: https://github.com/CentOS-PaaS-SIG/linchpin/issues/1290


samvarankashyap:
[merge]

PR #1294, opened 2019-08-22
Modify release notes script so it can run in nightly build

14rcole:


samvarankashyap:
@14rcole we can add it to release if you would like to.

14rcole:
@samvarankashyap might as well, it won't change anything
[merge]

PR #1295, opened 2019-08-22
Troubleshooting beaker credentials inside psi env

samvarankashyap:
$subject


samvarankashyap:
Found out that the credential files in Jenkins environment should match the provider name in order to make them work. 
closing the PR

PR #1298, opened 2019-08-22
Make cfgs run when only certain targets are specified

14rcole:


samvarankashyap:
@14rcole PSI failure is due to a jenkins restart. 
You can re-trigger the tests from psi openshift console.

johnbieren:
@mcornea fyi

14rcole:
@johnbieren If `cfgs` are in the PinFile but not listed in the targets, they will be parsed and applied to the layout when the inventory is generated.  Variables in `cfgs` will be applied to all targets in the layout with the proper resource group type

johnbieren:
Ah, okay so the key under cfgs: has to match the resource group type. Great :+1:

14rcole:
[merge]

PR #1312, opened 2019-09-05
Fix Openstack Network managment

abraverm:
- Openstack teardown of network resources wasn't in specific order which failed when there is dependency (net -> sub -> router).
- Network resources names were generated (uhash) which makes it impossible to link dependent resources (net -> sub -> router)f


abraverm:
[merge]

PR #1316, opened 2019-09-05
Enable public image azure

JunqiZhang0:
Enabled public image for Linchpin Azure
For current Azure support, we only support private image which means you can only create vm based on the user owned images. After enable the public image, user could create vm based on any online image

### DIFF
--- a/docs/source/azure.rst
+++ b/docs/source/azure.rst
@@ -24,12 +24,29 @@ definition, the following options are available.
 | role             | true       | string        | N/A               |                 |
 +------------------+------------+---------------+-------------------+-----------------+
 | vm_name          | true       | string        | name              |It can't include |
-|                  |            |               |                   | _ and other     |
+|                  |            |               |                   | '_' and other   |
 |                  |            |               |                   |special char     |
 +------------------+------------+---------------+-------------------+-----------------+
-| image            | true       | string        | image             |                 |
+| private_image    | false      | string        | image             |This para takes  |
+|                  |            |               |                   | private images  |
+|                  |            |               |                   |                 |
++------------------+------------+---------------+-------------------+-----------------+
+| resource_group   | true       | string        | resource_group    |                 |
++------------------+------------+---------------+-------------------+-----------------+
+| vm_size          | true       | string        | vm_size           |                 |
++------------------+------------+---------------+-------------------+-----------------+
+| public_image     | false      | dict          | image             |This para takes  |
+|                  |            |               |                   | public images   |
+|                  |            |               |                   |                 |
++------------------+------------+---------------+-------------------+-----------------+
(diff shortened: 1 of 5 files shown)


JunqiZhang0:
[test]

JunqiZhang0:
@abraverm Thanks for reviewing Alex
1. I had the documentation inside the Pinfile, I'll add more into azure.rst
2. I'll add the integration test
3. It wouldn't fail if both are declared, because it will take the private image as default. I documented this inside Pinfile, I'll added into the azure.rst as well
4. The test failed magically because of jenkins, it passed before

JunqiZhang0:
[test]

abraverm:
"3. It wouldn't fail if both are declared, because it will take the private image as default. " - I understand that, I think its better if it failed from user perspective: explicit declaration is easier to maintain than documented logic that can change. Its not new idea to document logic, but it was never liked by users IMHO.

JunqiZhang0:
@samvarankashyap both private and public are not required, if none is input, linchpin would take the defaulted one

JunqiZhang0:
@samvarankashyap we can't do your way because, private image is a string but public image is a dict, it's a little complex

docs/source/azure.rst
```diff
@@ -24,12 +24,29 @@ definition, the following options are available.
 | role             | true       | string        | N/A               |                 |
 +------------------+------------+---------------+-------------------+-----------------+
 | vm_name          | true       | string        | name              |It can't include |
-|                  |            |               |                   | _ and other     |
+|                  |            |               |                   | '_' and other   |
 |                  |            |               |                   |special char     |
 +------------------+------------+---------------+-------------------+-----------------+
-| image            | true       | string        | image             |                 |
+| private_image    | false      | string        | image             |This para takes  |
```
abraverm: is 'para' means 'parameter'?  the comment is quite obvious, I think you can mention the priority of private_image here.

docs/source/azure.rst
```diff
@@ -24,12 +24,29 @@ definition, the following options are available.
...
+|                  |            |               |                   |                 |
++------------------+------------+---------------+-------------------+-----------------+
+|virtual_network_name| true     | string        |virtual_network_name|                 |
++------------------+------------+---------------+-------------------+-----------------+
+| vm_username      | true       | string        | image             |                 |
++------------------+------------+---------------+-------------------+-----------------+
+| vm_password      | true       | string        | image             |                 |
 +------------------+------------+---------------+-------------------+-----------------+
 
+#If you declare both public and private image, only the private will be taken
```
abraverm: '#' at the beginning of the line doesn't have any special meaning in ReST syntax, can you change it to a bullet point?

JunqiZhang0:
[merge]

PR #1321, opened 2019-09-10
Openstack check resources hook

### LINKED ISSUE #1164: RFE: breaking up #857 POC on Openstack resource dependency
As a part of #857 
This ticket is for implementing cloud resource dependency check for openstack based resources. 
Which includes quota check, images, networks, tenant, credentials, ssh keys, files

abraverm:
Fixes  #1164

It is limited to what is available with Ansible facts and only checks resources that have dependency such as os_server. If more dependencies need to be tested, such as keypair for os_server, then it would be handled by developing a new ansible module and\or by using shell commands as workaround.

The hook could be extended to support other providers in the future.


abraverm:
@samvarankashyap , I thought about it, but then I remembered @14rcole  comment about having too much builtin hooks and I do plan to support other providers too (next would be beaker)

abraverm:
[test]

abraverm:
[test]

PR #1332, opened 2019-09-16
fix openstack teardown bug remove missing filter

samvarankashyap:
Remove missing filter which is causing the failure in openstack instances deletion


samvarankashyap:
[merge]

PR #1342, opened 2019-09-23
Fix schema for hooks

14rcole:
Fix error in hooks schema causing vault_password_file to not be read
Signed-off-by: 14rcole <rycole@redhat.com>


14rcole:
[merge]

PR #1348, opened 2019-09-25
Ignore credentials in workspaces

14rcole:
This just prevents people from accidentally uploading credentials when they submit a PR


14rcole:
[merge]

PR #1350, opened 2019-09-26
Python3 fix for --use-shell flag

samvarankashyap:
Linchpin  --use-shell feature does not work on python3.7 environments
The following commit acts as a fix for the same.


samvarankashyap:
[merge]

PR #1353, opened 2019-09-27
Drop support for ansible 2.6.x or below

samvarankashyap:
Since in requirements.txt we support only ansible versions>=2.7.1. 
The following PR cleans up ansible runner code dropping support for ansible versions
2.6.x or below


samvarankashyap:
[merge]

PR #1355, opened 2019-09-27
Added Azure doc and change name "Provediers" to Examples

### LINKED ISSUE #1199: Reference to examples in documentation
There are many examples but only few are referenced in the documentation.
It would be helpful if there would be a documentation page with all the examples and their descriptions and maybe tags (provider\topic). It also would be helpful if relevant pages (i.e providers) will have reference to examples. I believe this can be autogenerated with sphinx and ReST, maybe it will require additional plugin and updating the examples.

JunqiZhang0:
Add the examples by changing name of Providers, the name "Examples for Proveders" make more sense for user to find the examples fix #1199

### DIFF
--- a/docs/source/azure.rst
+++ b/docs/source/azure.rst
@@ -9,7 +9,7 @@ azure_vm
 Azure VM Instances can be provisioned using this resource.
 
 * Example <workspaces/azure/Pinfile>`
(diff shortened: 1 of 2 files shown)


JunqiZhang0:
[merge]

PR #1401, opened 2019-10-15
Added example resource into resource group integration test

### LINKED ISSUE #1375: Add an example to resource group[Azure]
Currently the example of creating a resource group is just creating resource group, could add some resource like a vm or network to test it works correctly as a resource group

JunqiZhang0:
Fix #1375


JunqiZhang0:
[merge]

Issue #1413, opened 2019-10-19
Linchpin destory fails with UnboundLocalError: local variable 'prev_id' referenced before assignment


Dannyb48:
**Describe the bug**
Performing some carbon/linchpin int testing on python 27 test client using python 3. I was able to successfully provision a beaker resource but when I went to cleanup it up with destroy it fails with

`local variable 'prev_id' referenced before assignment
 File "/home/cloud-user/cbn/lib/python3.6/site-packages/carbon/provisioners/ext/linchpin_wrapper_plugin.py", line 381, in delete
    self.linchpin_api.do_action(self.pinfile, action='destroy', tx_id=txid)
  File "/home/cloud-user/cbn/lib/python3.6/site-packages/linchpin/__init__.py", line 629, in do_action
    self.hooks.rundb = (rundb, rundb_id, prev_id)
UnboundLocalError: local variable 'prev_id' referenced before assignment
`

In carbon we don't do anything with hooks. 

Here is what the pinfile dict we generated looks like
```yaml
carbon:
  layout:
    inventory_layout:
      hosts:
        node:
          count: 1
          host_groups:
          - hosts
          - client
      vars:
        ansible_ssh_private_key_file: keys/demo
        ansible_user: root
        hostname: db2-carbon-beaker-node
  topology:
    resource_groups:
    - resource_definitions:
      - job_group: dummy
        recipesets:
        - arch: x86_64
          count: 1
          distro: RHEL-7.6
          hostrequires:
          - force: redacted host
          ids:
          - '3849923'
          name: db2-carbon-beaker-node
          ssh_key_file:
          - demo.pub
          variant: Server
        role: bkr_server
        ssh_keys_path: /home/cloud-user/cbn/e2e-acceptance-tests/keys
        whiteboard: py3-beaker-testing
      resource_group_name: carbon
      resource_group_type: beaker
    topology_name: carbon
```

**To Reproduce**
1. Install latest version of Linchpin from develop
2. Provision a resource 
3. Attempt to destroy using the transaction id 

**Expected behavior**
We are able to successfully destroy/cleanup the resources that were provisioned.

samvarankashyap:
fixed by #1414

PR #1414, opened 2019-10-21
bugfix: fix destroy issue #1413

samvarankashyap:
Fixes Issue #1413 
Caused due to improper indentation. 
I still wonder why it was not caught earlier.


samvarankashyap:
[merge]

PR #1417, opened 2019-10-22
Enable public key login

### LINKED ISSUE #1381: Enable the key login for Azure
Right now azure only supports password login, it needs the public key login

JunqiZhang0:
fix #1381


JunqiZhang0:
@abraverm I just followed the ansible parameter, and actually I strongly agree with you, but Ansible doesn't allow me to use username and password with the public key filled

abraverm:
More parameters will harm usability and increases the overall complexity, thus we should abvoid adding parameters and think of more elegant implementation, maybe Ansible advanced features can help.

14rcole:
Why can't the key just be used in `credentials`?

linchpin/provision/roles/azure/tasks/provision_azure_vm.yml
```diff
@@ -73,6 +86,7 @@
     admin_username: "{{ res_def['vm_username'] | default(linchpinUsername) }}"
     admin_password: "{{ res_def['vm_password'] | default(linchpinPassword) }}"
     resource_group: "{{ res_def['resource_group'] | default(omit) }}"
+    ssh_public_keys: "{{ssh_public_keys}}"
```
abraverm: how about  using `defaut(omit)` and also remove lines 75-79, will it work?

JunqiZhang0: No, it has to be NULL, it will not work if I use omit, Or I can write a filter which gives a NULL, but that way it will be inside py file,  this way it's clearer.

abraverm: then how about `default(NULL)`?

JunqiZhang0:
[test]

linchpin/provision/roles/azure/tasks/provision_azure_vm.yml
JunqiZhang0: I can't set a NULL directly to the default of the public key, because, even the public key is null, it's not gonna working. Ansible needs the field "ssh_public_keys" (contains "public key") to be NULL. Only set "public key" to null, the "ssh_public_keys" will be LIST("path/to/key",NULL) which is not NULL

abraverm:
why not one commit?

JunqiZhang0:
[MERGE]

Issue #1425, opened 2019-10-23
linchpin libvirt provisioning fails during 'define node template' task with UndefinedVariable: 'int object' has no attribute 'memory'" error


mcornea:
**Describe the bug**
linchpin libvirt provisioning fails during 'define node template' task with UndefinedVariable: 'int object' has no attribute 'memory'" error:

```shell

TASK [libvirt : define node template] ********************************************************************************************************************************************************************************************************
[...]

PR #1429, opened 2019-10-24
Adjust indices in libvirt node template to match changes in Ansible

14rcole:
This fixes an issue that was posted on Google chat.  It looks like at some point an argument was removed from a task in linchpin, but the indices in the template remained the same


samvarankashyap:
@14rcole 
this should also fix #1425 
Good job :+1:

14rcole:
[merge]

PR #1432, opened 2019-10-25
Introduce libvirt network IPv6 support

mcornea:
- extends the libvirt network template to allow passing ipv6 required tags
- switch the wait for ip address task to use iproute instead of arp. excludes link local(fe80) addresses from check
- extends user-data template to allow passing commands from the topology definition. this is used to enable ipv6 config inside the VMs at boot time: 
```yaml
        cloud_config:
          run_commands:
            - nmcli con modify 'System eth0' ipv6.method auto
```


samvarankashyap:
[merge]

PR #1433, opened 2019-10-25
Run sync instead of sleep during libvirt provisioning

mcornea:
Instead of wasting 2 minutes waiting let's run sync to write the memory data to disk, which should avoid failures caused by delayed write locks.


samvarankashyap:
[merge]

PR #1434, opened 2019-10-25
Allow roles to write to topology_outputs directly

14rcole:
This allows role developers to avoid the hard-coding in gather_resources while developing external roles


14rcole:
@samvarankashyap If we want galaxy roles to be a preview feature in 1.9, this needs to be merged before the release

samvarankashyap:
[merge]

PR #1436, opened 2019-10-28
Coala init

abraverm:
Centralized static analysis management tool for Linchpin.
To use Coala, install 'test' requirements or with pip `pip install coala-bears` and run `coala`. 
By default Coala will run in interactive mode and ask what to do with the issues found, this could be avoided by running with --ci option - `coala --ci`. Also, one can specifiy specific tests (Bears) to run and on which files to do it. General settings are stored in `.coafile`.

### DIFF
--- a/.coafile
+++ b/.coafile
@@ -1,0 +1,47 @@
+[python]
+files = linchpin/**.py
+language = Python
+max_line_length = 80
+use_spaces = True
+
+[python.looks]
+# Patches may conflict with autopep8 so putting them in own section so they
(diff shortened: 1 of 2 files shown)


14rcole:
@abraverm I know it's been a while, but would I be correct in saying that coala covers the same features as flake8 plus a little bit more?  If so, should lwe remove flake8 from our unit tests and replace it with coala?

abraverm:
Actually after using Coala for some time now, I found that you can install it with Pip but it has such strict requirements and so many of them that it can criple a project or the usage of Coala. This change is abandoned

PR #1441, opened 2019-10-28
Fix the ansible config paths since they would not be available

samvarankashyap:
$subject

PR #1443, opened 2019-10-29
Use the ansible galaxy python API to search for role paths

14rcole:
This is the less hacky (but potentially less stable) way of finding the role paths that Ansible uses


14rcole:
@samvarankashyap You can close #1441 

@Dannyb48 let me know if this resolves the role issue you were having

samvarankashyap:
[merge]

Dannyb48:
@14rcole This worked thanks for the quick turnaround.

PR #1448, opened 2019-10-29
Patch for str to array error in beaker output

samvarankashyap:
Fixes: 
Fixes beaker provision failure.
This is just a patch.


14rcole:
This still won't solve the "role = AnsibleUndefined" error, will it?

samvarankashyap:
@14rcole No. It won't I guess.

PR #1449, opened 2019-10-29
Fix issues preventing beaker resources from being generated properly

14rcole:
This will fix the last issue blocking release


14rcole:
@Dannyb48 good catch regarding the "AnsibleUndefined" output, that was actually the key to the whole thing

Dannyb48:
@14rcole This resolved the problem! Awesome work!!

samvarankashyap:
[merge]

PR #1462, opened 2019-11-02
Update the azure requirements

JunqiZhang0:
Updated the azure requirements and reverted one commit to pass the unittest


linchpin/provision/library/rundb.py
```diff
@@ -5,7 +5,296 @@
 import ast
 import json
 from ansible.module_utils.basic import AnsibleModule
-from ..module_utils.linchpin_rundb import TinyRunDB, BaseDB
```
14rcole: It looks like you may need to rebase

PR #1463, opened 2019-11-05
Updated the azure dependencies to compatible with Ansible 2.9.0

JunqiZhang0:
Updated the azure dependencies to compatible with Ansible 2.9.0


requirements-azure.txt
```diff
@@ -1,36 +1,42 @@
-packaging
-requests[security]
+PyYAML
```
abraverm: PyYAML is ansible requirement

requirements-azure.txt
```diff
@@ -1,36 +1,42 @@
...
-azure-mgmt-rdbms==1.4.1
 azure-mgmt-servicebus==0.5.3
 azure-mgmt-sql==0.10.0
 azure-mgmt-storage==3.1.0
 azure-mgmt-trafficmanager==0.50.0
 azure-mgmt-web==0.41.0
 azure-nspkg==2.0.0
 azure-storage==0.35.1
+cryptography
+jinja2
```
abraverm: Those are also ansible requirements:
https://github.com/ansible/ansible/blob/devel/requirements.txt

JunqiZhang0: Let me try get rid of these

abraverm:
looks like `pip install ansible[azure]` works too, why not use that instead ?

JunqiZhang0:
> looks like `pip install ansible[azure]` works too, why not use that instead ?

We tried this before, it only installs ansible for you, we need azure dependencies

abraverm:
> > looks like `pip install ansible[azure]` works too, why not use that instead ?
> 
> We tried this before, it only installs ansible for you, we need azure dependencies

It worked for me today... maybe you need to install ansible first?

14rcole:
@junqizhang Will this PR resolve the issue I'm seeing in my tests?
```
msg": "Failed to import the required Python library (ansible[azure] (azure >= 2.0.0)) on lp-centos7-ced21096-ea72-4f0f-b4c8-da3cc89767d5-s7mj0-wzs57's Python /usr/bin/python2. Please read module documentation and install in the appropriate location. If the required library is installed, but Ansible is using the wrong Python interpreter, please consult the documentation on ansible_python_interpreter"
```

JunqiZhang0:
@14rcole yes

samvarankashyap:
[merge]

samvarankashyap:
[merge]

PR #1471, opened 2019-11-08
Add support for mongodb driver in rundb

14rcole:
This PR accomplishes a few things:

1. Add's support for MongoDB to the rundb
2. Decouples the rundb code so that other drivers can be added more easily
3. Moves tinydb-specific logic back into tinydb.  Maybe this will allow us to support parallelism in tinydb?  Also, we may be able to do that with just file locks - tinydb is just a json document after all


linchpin/provision/library/rundb.py
```diff
@@ -1,15 +1,18 @@
...
 from ansible.module_utils.basic import AnsibleModule
-from linchpin.linchpin_rundb import TinyRunDB, BaseDB
+
+from linchpin.rundb.basedb import BaseDB
+from linchpin.rundb.tinydb import TinyRunDB
+from linchpin.rundb.mongodb import MongoDB
 
 DB_DRIVERS = {
     "TinyRunDB": TinyRunDB,
+    "MongoDB": MongoDB
```
samvarankashyap: @14rcole If we make MongoDB default. We have to run MongoDB in a container for our linchpin integration tests.

14rcole: I don't think we should make it the default, I think we should just open up the option for teams like Carbon

samvarankashyap:
@14rcole One of the reasons the concurrency might be failing is because of the CACHE size 
at
https://github.com/CentOS-PaaS-SIG/linchpin/blob/develop/linchpin/linchpin_rundb.py#L138 
Could you increase the it to 4096 or more? It might fix the concurrency issue with Rundb

samvarankashyap:
[test]

14rcole:
@samvarankashyap I can change it, but the concurrency issue with tinydb is here:
https://github.com/CentOS-PaaS-SIG/linchpin/blob/develop/linchpin/linchpin_rundb.py#L170

14rcole:
@samvarankashyap our unit tests are currently run in python3, right?

samvarankashyap:
@14rcole Yes. 
They are running in a python3 virtualenv

14rcole:
[test]

14rcole:
[merge]

PR #1491, opened 2019-11-21
Remove fedora29 from the container integration tests

samvarankashyap:
Dropping support for fedora 29 
removing references to fedora29


samvarankashyap:
[merge]

PR #1495, opened 2019-11-25
Mock linchpin dummy module for easier testing 

samvarankashyap:
This PR is a proof of concept to use ansible action_plugins to replace integration testing 
within linchpin.

### DIFF
--- a/config/Dockerfiles/tests.d/dummy/11_mock-dummy-module
+++ b/config/Dockerfiles/tests.d/dummy/11_mock-dummy-module
@@ -1,0 +1,34 @@
+#!/bin/bash -xe
+
+# Verify --no-hooks and --ignore-failed-hooks on dummy provider
+# distros.exclude: none
+# providers.include: dummy
+# providers.exclude: none
+
+DISTRO=${1}
+PROVIDER=${2}
+
+PINFILE="PinFile"
+WORKSPACE_PATH="docs/source/examples/workspaces/mock_dummy_module"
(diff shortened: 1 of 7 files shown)


samvarankashyap:
[test]

samvarankashyap:
@14rcole Needs review
With the above structure, we can easily adapt to all other providers. 
which helps us cut down the testing time from hours to minutes. 
The best part is we can do the same with downstream providers like beaker and openstack. 
Making all the tests in one place.
We can also leverage external ci systems like Travis, Github actions or circle CI to do the same.

config/Dockerfiles/tests.d/dummy/11_mock-dummy-module
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash -xe
```
14rcole: Same comments as in the openstack PR.  Also, shouldn't we be mocking all of the tests, not just doing one mocked test?

samvarankashyap: @14rcole This is just a POC. 
for dummy provider. 
We should be eventually transitioning to mocking all the providers.
Since the downstream Jenkins is not working Openstack and beaker are of utmost priority. 

@14rcole #1509 is PR for all openstack tests and am working on the beaker.

samvarankashyap:
Acted as POC. 
Not needed anymore

PR #1507, opened 2019-12-02
bugfix: Update private image path in examples

JunqiZhang0:
create a new private image and fix the bug caused by it


samvarankashyap:
[merge]

PR #1506, opened 2019-12-02
fixes: #1503 Linchpin clean installation with Python 3 requires pymongo

samvarankashyap:
Fixes #1503

PR #1509, opened 2019-12-02
Add Mock tests to linchpin openstack & beaker providers

samvarankashyap:
Summary:
- Made changes to existing playbooks to fix a few bugs
- Added action_plugins to run mock tests 
- Added openstack-mock-workspace for testing 
- Added test for running openstack mock workspace tests


docs/source/examples/workspaces/openstack-mock-workspace/PinFile
```diff
@@ -0,0 +1,159 @@
+---
```
14rcole: Why do we need a new workspace for this?

samvarankashyap: @14rcole I just wanted to keep the mocked workspaces different from the actual ones. 
Since the linchpin.conf is different and mock data is only applicable to specific examples. 
for the time being.

samvarankashyap:
Failed due to bug in develop 
#1506 needs to be merged before this PR is tested

docs/source/examples/workspaces/openstack-mock-workspace/PinFile
14rcole: I thought that we were going to mock all of the providers?

Regardless, the workspaces need to be provisionable in both a mocked and non-mocked way.  Otherwise we can't effectively verify that the data being sent back from a given provider doesn't change.  Why don't we have `linchpin.conf` and `linchpin-mock.conf` in `docs/source/examples/workspaces` then have the integration tests look for a flag that, when set, uses the mock config/data instead of the normal data.

samvarankashyap: @14rcole Yes we will be mocking all providers. 
It takes time to get all the sample data from all the providers. 
Having two linchpin.conf makes sense. 
I think we have option in linchpin to specify the linchpin.conf through the command line. 
we can make use of it. 
I will try to implement the same and maintain one workspace for both.

samvarankashyap:
[test]

samvarankashyap:
[test]

samvarankashyap:
[merge]

PR #1514, opened 2019-12-04
Improved Deep clean Feature

### LINKED ISSUE #1423: Improve the deep clean function of Azure
The deep clean at this moment is just clean the extra disk, it should clean all the resources attached. And the ansible built-in function (remove_on_absent) field should be removed. (Set it as default)We should have a list of all the resources attached and then clean all of them.

JunqiZhang0:
Fix #1423


samvarankashyap:
[merge]

PR #1527, opened 2019-12-05
Add support for vault-password-file in ansible api

### LINKED ISSUE #1351: Add `vault-password-file` support to hooks via ansible API
This already works with `--use-shell`, providing a temporary workaround for #1254.  Now  we need to add that to the API as well

14rcole:
Fixes #1351


14rcole:
[merge]

PR #1544, opened 2019-12-11
Fix install_selinux_venv.sh for RHEL8 environments

mcornea:
On RHEL8 _selinux.so file is not present. This change adjusts the symlinks in install_selinux_venv.sh to match python version present on RHEL8.


samvarankashyap:
[merge]

Issue #1543, opened 2019-12-11
Reduce Cyclomatic Complexity in do_action function


14rcole:
**Is your feature request related to a problem? Please describe.**
Sub-issue of #1435 

**Describe the solution you'd like**
The complexity of this function should be reduced from an E to a C (and that of all functions created while refactoring)

**Describe alternatives you've considered**
None

14rcole:
This was resolved in #1554

PR #1554, opened 2020-01-06
Decrease cyclomatic complexity of do_action function

14rcole:
Also add radon tests to linchpin unit tests, but do not require a certain score in order to pass


14rcole:
resolves #1543

14rcole:
@abraverm `do_action` and all of the helper functions I created for it are (just barely) down to a B!  We should absolutely tackle the rest and put a failure condition on the radon tests, but this was by far our worst function

14rcole:
[merge]

PR #1556, opened 2020-01-07
Make aws module OASIS compatible

14rcole:


linchpin/FilterUtils/FilterUtils.py
```diff
@@ -301,3 +301,13 @@ def convert_to_json(input_str):
...
     else:
         return input_str
+
+
+def path_relative_to(path, base_path):
+    """
+    If `path` is not an OS filesystem absolute or relative path then assume
+    it is relative to `base_path`.
+    """
+    if path.startswith(('/', './', '../', '~/')):
```
abraverm: Why change OS standards that known by all?

linchpin/provision/roles/aws/.travis.yml
```diff
@@ -0,0 +1,28 @@
...
+  - pip install ansible ansible-lint yamllint flake8 molecule docker
+  - ansible --version
+  - ansible-lint --version
+  - yamllint --version
+  - flake8 --version
+  - molecule --version
+
+script:
+  # every scenario with a converge playbook should be linted
+  - molecule lint -s openstack
```
abraverm: aws?

linchpin/provision/roles/aws/Jenkinsfile
```diff
@@ -0,0 +1,8 @@
+library('oasis-pipeline')
+
+oasisMultistreamMoleculePipeline {
+  upstream_git_url = 'https://github.com/oasis-roles/aws.git'
+  molecule_role_name = 'aws'
+  molecule_scenarios = ['openstack']
```
abraverm: aws?

linchpin/provision/roles/aws/library/auth_driver.py
```diff
@@ -0,0 +1,146 @@
+#!/usr/bin/env python
```
abraverm: we usually use symbolic link to library, change of policy?

linchpin/provision/roles/aws/meta/main.yml
```diff
@@ -0,0 +1,15 @@
+galaxy_info:
+  author: Author Name
```
abraverm: fix metadata?

linchpin/provision/roles/aws/molecule/openstack/create.yml
```diff
@@ -0,0 +1,9 @@
+- name: Create
+  hosts: localhost
+  connection: local
+  gather_facts: false
+  roles:
+    - role: oasis_roles.molecule_openstack_ci
+      molecule_openstack_ci_state: prsent
+      molecule_openstack_ci_cloud: |-
+        {{ lookup('env', 'OS_CLOUD') | default('default', true) }}
```
abraverm: Openstack stuff

linchpin/provision/roles/aws/molecule/openstack/destroy.yml
```diff
@@ -0,0 +1,9 @@
+- name: Destroy
+  hosts: localhost
+  connection: local
+  gather_facts: false
+  roles:
+    - role: oasis_roles.molecule_openstack_ci
+      molecule_openstack_ci_state: absent
+      molecule_openstack_ci_cloud: |-
+        {{ lookup('env', 'OS_CLOUD') | default('default', true) }}
```
abraverm: Openstack stuff

linchpin/provision/roles/aws/molecule/openstack/molecule.yml
```diff
@@ -0,0 +1,36 @@
+dependency:
+  name: galaxy
+  options:
+    role-file: molecule/shared/requirements.yml
+driver:
+  name: openstack
```
abraverm: Openstack stuff

linchpin/provision/roles/aws/molecule/shared/prepare.yml
```diff
@@ -0,0 +1,15 @@
...
+  vars:
+    rhsm_username: "{{ lookup('env', 'OASIS_RHSM_USERNAME') }}"
+    rhsm_password: "{{ lookup('env', 'OASIS_RHSM_PASSWORD') }}"
+    rhsm_server_hostname: "{{ lookup('env', 'OASIS_RHSM_SERVER_HOSTNAME') }}"
+    rhsm_pool_ids: "{{ lookup('env', 'OASIS_RHSM_POOL_IDS') }}"
+    rhsm_repositories:
+      enable:
+        - rhel-7-server-rpms
+        - rhel-7-server-extras-rpms
+        - rhsl-7-server-optional-rpms
```
abraverm: Why do we need RHSM?

linchpin/provision/roles/aws/molecule/shared/requirements.yml
```diff
@@ -0,0 +1,3 @@
+- oasis_roles.rhsm
+- oasis_roles.molecule_openstack_ci
```
abraverm: RHSM? Openstack stuff

linchpin/provision/roles/aws/tasks/teardown_res_defs.yml
```diff
@@ -10,7 +14,10 @@
     loop_var: res_item
 
 - name: "teardown of s3 resource def of current group"
-  include: teardown_aws_s3.yml res_def={{ res_item.0 }} res_grp_name={{ res_item.1 }}
+  include_tasks: teardown_aws_s3.yml
+  vars:
+          res_def: "{{ res_item.0 }}"
```
abraverm: bad indentation?

linchpin/FilterUtils/FilterUtils.py
14rcole: What do you mean?  Also, I should mention that this filter was the last one in a separate file, so I just moved the code in here with the rest of the filters.

linchpin/provision/roles/aws/.travis.yml
14rcole: I think this means "test this role from an openstack host".  It doesn't really matter for us yet since we're not using the molecule tests yet.  I can delete it though since we'll probably only use the docker/podman provider for molecule anyway

linchpin/provision/roles/aws/Jenkinsfile
14rcole: Same as above

linchpin/provision/roles/aws/library/auth_driver.py
14rcole: What do you mean?

linchpin/provision/roles/aws/molecule/shared/prepare.yml
14rcole: This is  a part of the OASIS defaults.  I don't quite know what its used for.  I can try deleting it and seeing what happens since the docker testing provider uses centos 7 instead of RHEL

linchpin/FilterUtils/FilterUtils.py
abraverm: given a path which starts ('/', './', '../', '~/') then join it with `bath_path` path - this logic doesn't make sense because:
`/` - is absolute path
`./` - is current folder (relative)
`../` - is parent folder (relative)
`~/` - is home folder (relative)
Why change that behavior

linchpin/provision/roles/aws/.travis.yml
abraverm: I think it runs linting against "openstack" folder?

linchpin/provision/roles/aws/Jenkinsfile
abraverm: Not sure what is a molecule scenario...

linchpin/provision/roles/aws/library/auth_driver.py
abraverm: well usually inside the roles we have a symbolic link to the library, in this case it would be
linchpin/provision/roles/aws/library -> ../../library

linchpin/provision/roles/aws/molecule/shared/prepare.yml
14rcole: The short answer is that we don't right now.  However, I do have a couple of PRs coming up to add RHEL 7 and RHEL 8 support.  I can remove the lines referencing the rhel-related roles for oasis and then add them back in those PRs if you'd prefer.  Or I can just leave them in here to avoid breaking anything

linchpin/provision/roles/aws/library/auth_driver.py
14rcole: I can put them symbolic link back in.  The reason I did this is to separate out the files in library that are actually necessary for each provider so that we can more easily separate each provider into repos when that becomes an option.  I can also link the individual files instead of the entire library directory if you think that's a more elegant solution.  That way we know the true dependencies of each role

14rcole: Ideally, this would be the situation:
* Each role has its own GitHub repo so that we can store the roles on Galaxy
* `library` and `filter_plugins`/`FilterUtils` are their own separate repos which are included in the roles through the use of a submodule

But that's far off at best

abraverm: I think symbolic link expresses that ideal and will be a reminder when the role will be moved. The part I'm worried of is that some changes will be specific for AWS and it will be hard to reunite as time go by. But I would like to track this issue for future work of OASIS and Linchpin, maybe some of the library and plugins could be moved to specific roles (for example Beaker) and other parts could become generic solutions (authentication).

linchpin/provision/roles/aws/molecule/shared/prepare.yml
abraverm: I'm a coding puritan but it doen't hurt and it has already an enabling condition.

linchpin/FilterUtils/FilterUtils.py
14rcole: I looked at the code from which `path_relative_to` is called from and it essentially is meant to make a non-absolute path a relative path".  So if there's a path that looks like this: `provision/roles/aws` it appends the base path to it so it becomes `/path/to/linchpin/linchpin/provision/roles/aws`.  It's only used in the docker provider.  If the path starts with any of those then its just returned as-is.

14rcole:
Resolves #1357

linchpin/FilterUtils/FilterUtils.py
abraverm: If it is only used by docker provider, why is it in the general FilterUtils and why is it part of AWS change?

14rcole:
@samvarankashyap any notes before this gets merged?

14rcole:
[merge]

PR #1562, opened 2020-01-14
Auth debug variable

abraverm:
Debug mode is required for CI investigation (for linchpin project or user tests) and it could be needed by CI engineers (infra owners) or CI users (Quality engineers), but not everyone should able to see authentication credentials in that case. The change adds a new variable to Linchpin config to separate between general debug mode and credentials specific case.

### DIFF
--- a/docs/source/examples/workspace/linchpin.conf
+++ b/docs/source/examples/workspace/linchpin.conf
@@ -113,9 +113,13 @@ distill_on_error = False
 #async_timeout = 1000
 
 # enables debug mode, which turns on debugging outputs for certain commands.
+#debug_mode = False
+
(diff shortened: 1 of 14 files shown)


abraverm:
I searched for 'auth_driver' in the roles.. so, that's what we have...

JunqiZhang0:
@abraverm Shouldn't we change all the debug_mode to your new variable to be consistent?

abraverm:
@JunqiZhang0, what's the point of changing all `debug_mode` to `auth_debug`? the purpose of the PR is to allow running Linchpin in debug mode without exposing credentials.

JunqiZhang0:
[merge]

PR #1563, opened 2020-01-14
Make profile a variable in Azure example

abraverm:
This is a preparation towards testing different authentication methods (AD)


abraverm:
[merge]

PR #1564, opened 2020-01-14
Pass credentials path to tests instead of azure subscription

abraverm:
All tests were getting azure subscription id  or nothing, this PR changes it to credentials path as it has higher chance to be used in other tests. Also it will make it possible to test different subscription id if it will require. Last, the PR fixed bash scripts linting issues and makes it more readable.


abraverm:
[test]

abraverm:
[test]

abraverm:
[merge]

PR #1568, opened 2020-01-15
Fix unit test failures in linchpin API related to progress bar

14rcole:


samvarankashyap:
@14rcole Any update? 
Can we merge this?

14rcole:
@samvarankashyap My only questions is how to we re-enable the unit tests?  It doesn't look like they're running now and they really need to be enabled for the release

14rcole:
[merge]

PR #1571, opened 2020-01-16
Fix progress bar bug for linchpinAPI

### LINKED ISSUE #1566: Cannot run unit tests
**Describe the bug**
Unit tests are failing because the linchpin API does not support the progress monitor

**To Reproduce**
1. Go to the linchpin base dir
2. run `python setup.py test`

**Expected behavior**
Tests pass

samvarankashyap:
@14rcole This PR fixes #1566

### DIFF
--- a/linchpin/__init__.py
+++ b/linchpin/__init__.py
@@ -109,6 +109,10 @@ class LinchpinAPI(object):
         self.set_evar('from_api', True)
         self.workspace = self.get_evar('workspace')
         self.disable_pbar = 'True'
+        if not hasattr(self.ctx, 'no_monitor'):
+            self.ctx.no_monitor = False
+        self.get_evar('no_monitor', self.ctx.no_monitor)
+
 
     def setup_pbar(self):
         if (eval(self.get_cfg('progress_bar', 'no_progress')) or
@@ -1026,6 +1030,7 @@ class LinchpinAPI(object):
             self.set_evar('state', 'absent')
 
         inventory_src = '{0}/localhost'.format(self.workspace)
+        self.setup_pbar()
         self.pbar.set_description_str("Linchpin(%s)" % action)
         self.pbar.total = len(resources) + 1
         self.pbar.update()


linchpin/__init__.py
```diff
@@ -109,6 +109,7 @@ def __init__(self, ctx):
         self.set_evar('from_api', True)
         self.workspace = self.get_evar('workspace')
         self.disable_pbar = 'True'
+        self.ctx.no_monitor = False
```
14rcole: You'll need to check if ctx has already set monitor before setting it to false

linchpin/__init__.py
```diff
@@ -1026,6 +1027,8 @@ def _invoke_playbooks(self, resources={}, action='up', console=True,
             self.set_evar('state', 'absent')
 
         inventory_src = '{0}/localhost'.format(self.workspace)
+        if not hasattr(self, 'pbar'):
```
14rcole: Why don't we just only use `setup_pbar()` within the `LinchpinAPI` class instead of having `linchpin/cli/__init__.py` also run the code?

linchpin/__init__.py
14rcole: Also, you'll have to set the evar `no_monitor` because it needs to be available in the ZMQ CallbackPlugin

14rcole:
After making these comments I remembered I had also submitted #1568 yesterday for the same issue, but I have to make some changes there as well so we can take whichever PR you want

samvarankashyap:
@14rcole Let me know when your changes are ready

linchpin/__init__.py
samvarankashyap: @14rcole I didn't get you. 
Currently self.setup_pbar is only used in linchpin/__init__ and shell/__init__.py 
I dont see it being run in cli implementation. 
Could you clarify on it.

14rcole: Sorry, I meant `shell/__init__.py`.  Basically, there's no reason for a class outside of `linchpin/__init__.py` to have to setup the progress bar, so if we just run `setup_pbar()` on line 1034 here we can remove the if statement on line 1033

samvarankashyap:
@14rcole PR updated needs review

samvarankashyap:
@14rcole shell/__init__ needs to implement progress bar because the progress bar is only visible in shell context. It is not visible or usable in case of API calls.

14rcole:
Closing since #1568 was merged

PR #1576, opened 2020-01-17
Added res_type to Azure Network Interface

JunqiZhang0:
Added resource type to the output for Azure Network Interface. It will appear in azure_nic.output


JunqiZhang0:
[merge]

Issue #1218, opened 2019-07-17
Manage Azure storage accounts


abraverm:
Create, update or delete a storage account.

https://docs.ansible.com/ansible/latest/modules/azure_rm_storageaccount_module.html#azure-rm-storageaccount-module

PR #1591, opened 2020-01-26
Azure storage account

abraverm:


abraverm:
Fixes #1218

abraverm:
test

abraverm:
[test]

linchpin/hooks/built_ins/azure_boot_log/boot_log.yml
```diff
@@ -12,11 +12,15 @@
       vault_pass: "{{ vault_pass }}"
     register: auth_var_out
     delegate_to: localhost
+    no_log: "{{ not auth_debug }}"
 
   - name: "set auth_var"
     set_fact:
       auth_var: "{{ auth_var_out['output'][credentials_profile] }}"
 
+  - debug:
```
14rcole: Is the debug necessary?

Also, one of the linting errors you'll see a lot of in ansible-lint is that all tasks must be named.  So if this is necessary, please name it

abraverm:
[test]

linchpin/hooks/built_ins/azure_boot_log/boot_log.yml
abraverm: no, I was trying to figure out why CI was failing.

abraverm:
[test]

linchpin/hooks/built_ins/azure_boot_log/boot_log.yml
```diff
@@ -12,6 +12,7 @@
       vault_pass: "{{ vault_pass }}"
     register: auth_var_out
     delegate_to: localhost
+    no_log: "{{ not auth_debug }}"
```
14rcole: This isn't really specific to this PR, but why do we have an `auth_debug` and a `debug` variable?  They seem to do the same thing

abraverm: Because when debugging `auth_debug` module it exposes the credentials, but we need debug enabled in CI tests for other modules.

14rcole: The debug variable was originally created to hide credentials.  If there's tasks that don't expose credentials in which debug is set, we should remove them in another PR and consolidate to one variable.  We don't need to worry about that for this PR though

abraverm:
[merge]

PR #1599, opened 2020-02-04
AWS EC2 resource check

### LINKED ISSUE #1585: AWS resouces check
Validate that required resources exist with a pre-hook before running Linchpin up.

abraverm:
fixes #1585 

With the built-in pre-up hook, the user can check if required resources exists for creating ec2 instance:
 - security group
 - image
 - region


abraverm:
aws-simple example was updated where the pre-up hook was added

abraverm:
[merge]

Issue #1600, opened 2020-02-05
Test `01_azure_vm` is failing


14rcole:
Azure VM test fails at the book log postup hook.  Output from the hook is below.

samvarankashyap:
@14rcole Looks like syntax problem which also occurs when the parameters are incorrect.

samvarankashyap:
@abraverm Any clue?

abraverm:
1. delete\edit the log it contains  credentials
2. that's the issue I mentioned in google chat - there is some problem with cfgs, it tries to render __LOG__  but it just ends up '__log__' string.

PR #1601, opened 2020-02-05
Temporarily disable postup hook for azure_vm

14rcole:
The hook is causing all PRs that test azure to fail.  Disabling until #1600 is resolved


14rcole:
[merge]

PR #1602, opened 2020-02-05
load strings from topology output to fix AWS EC2 destroy

samvarankashyap:
AWS EC2 destroy is currently fails due to the rundb output being a string .
This PR uses the convert_to_json filter to load the topology out from string if it is a string else it returns the same.


maneeshmehra:
@samvarankashyap: Is there an issue that this PR tracks and fixes ?

samvarankashyap:
fixes #1604

samvarankashyap:
[merge]

Issue #1606, opened 2020-02-11
Custom roles do not add info to topology_outputs


14rcole:
The aggregation of the `topology_outputs_<whatever>` variables that store the data from provisioning each resource definition is hard-coded in the `gather_resources` role.  Any outputs that are not aggregated here will not have their inventories generated properly, which means that no custom ansible roles will be able to generate inventories.  This will need to be fixed

samvarankashyap:
fixed by #1607

PR #1607, opened 2020-02-11
Optimize topology_outputs aggregation

14rcole:
This automates the aggregation of topology outputs by collecting data from every variable beginning with `topology_outputs`.

If we want to standardize the names of the variables to `topology_outputs_<resource group>_<role>` then we can make it even easier to filter out false positives.  This may also make it easier to do some of the molecule testing, although the current PR could also be sufficient


14rcole:
Fixes #1606

14rcole:
[merge]

PR #1608, opened 2020-02-12
Fix bug preventing topology outputs from multiple sources from being collected

14rcole:
I noticed this while I was working on something else.  #1607 didn't aggregate outputs into `topology_outputs`, it overwrote the outputs every time.  This fixes that bug


14rcole:
[merge]

PR #1610, opened 2020-02-13
Add file permissions to keyfiles created

### LINKED ISSUE #1588: Generated cloud ssh key files have permissions that are too open for use later by Ansible
**Describe the bug**
After Linchpin is used to provision aws_keys or os_keys and you try to use that key later with Ansible, it fails with `key too open` errors. 

We've been working around this by changing the permissions in our code but I think this logic is better suited to be done by Linchpin since it is the one creating it. 

Need to update the filter `write_to_file` to maybe take a boolean param so that it knows it should change the permissions on the file if it is a key file. 

**To Reproduce**
1. Create a PinFile that creates the key file and a cloud resource
2. Create a layout that utilizes the key file as an `ansible_ssh_private_key_file`
2. Run `linchpin -vvvv up`
3. Run `ansible-playbook` command to run a test playbook that uses the inventory and key file. 

**Expected behavior**
Ansible playbook commands should work right after.

samvarankashyap:
$subject
fixes #1588

### DIFF
--- a/linchpin/FilterUtils/FilterUtils.py
+++ b/linchpin/FilterUtils/FilterUtils.py
@@ -287,13 +287,17 @@ def get_os_server_names(topo_output):
     return names
 
 
-def write_to_file(data, path, filename, extn='.output'):
+def write_to_file(data, path, filename, extn='.output', mode=0o600):
     
     filename = filename.replace(' ', '_').lower()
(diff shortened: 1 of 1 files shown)


samvarankashyap:
[merge]

PR #1623, opened 2020-02-20
Test libvirt only on Fedora 31

14rcole:


14rcole:
Now that we've mocked the libvirt tests and we are moving to github actions, this isn't really relevant

PR #1627, opened 2020-02-21
Add delegated driver support to molecule tests

14rcole:
Adds support for the delegated driver in molecule tests that that we don't have to use nested containers in github actions

To run the delegated driver locally, you may have to set `ansible_python_interpreter` to the python interpreter within your virtual environment.  This is set under `ansible_connection_options` in `molecule/delegated/molecule.yml`

### DIFF
--- a/linchpin/provision/roles/aws/molecule/delegated/create.yml
+++ b/linchpin/provision/roles/aws/molecule/delegated/create.yml
@@ -1,0 +1,6 @@
+- name: Create
+  hosts: all
+  tasks:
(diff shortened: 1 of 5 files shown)


14rcole:
[merge]

14rcole:
@samvarankashyap You have to specify the scenario for the molecule tests, which here is either `docker` or `delegated`.  So you would run

```
molecule test -s delegated
```
That will run the entire test suite.  If you just want to run the integration tests:
```
molecule converge -s delegated
```

Issue #1630, opened 2020-02-25
1.10.0 cloud-config doesn't work on libvirt host


johnbieren:
**Describe the bug**
When provisioning VMs with linchpin 1.10.0, execution completes but ssh'ing to the nodes does not work if you are using a cloud-config setup

**To Reproduce**
1. Provision VM on rhel8.1 with libvirt provider that uses cloud-config
2. Try to ssh to provisioned VMs
3. Get connection refused

**Expected behavior**
ssh works

**Desktop (please complete the following information):**
 - linchpin 1.10.0
- RHEL 8.1

**Additional context**
Works on linchpin 1.9.2

samvarankashyap:
@johnbieren Could you share the Pinfile too ? 
Thanks

samvarankashyap:
fixed by #1636

PR #1636, opened 2020-02-26
update logic for cloud-init and variable names for resolving ambiguity

samvarankashyap:
$subject
#1630 

@johnbieren Please check if this patch works for you.

### DIFF
--- a/linchpin/provision/roles/libvirt/tasks/provision_libvirt_node.yml
+++ b/linchpin/provision/roles/libvirt/tasks/provision_libvirt_node.yml
@@ -23,11 +23,11 @@
 - name: set cloud config default
   set_fact:
     cloud_config: "{{ res_def['cloud_config'] }}"
-    cloud_config_used: true
+    cloud_init_used: true
 
 - name: "check cloud_config is used or not"
   set_fact:
-    cloud_config_used: "{{ true if (cloud_config['users'] == [] and cloud_config['run_commands'] == []) else false }}"
+    cloud_init_used: "{{ true if (cloud_config['users'] != [] and cloud_config['run_commands'] is not defined ) else false }}"
 
 - name: "set empty local_image_path"
   set_fact:
@@ -166,7 +166,7 @@
     loop_var: definition
   remote_user: "{{ res_def['remote_user'] }}"
   delegate_to: "{{ uri_hostname }}"
-  when: node_exists['failed'] is defined and uri_hostname != 'localhost' and virt_type == "cloud-init" and cloud_config_used
+  when: node_exists['failed'] is defined and uri_hostname != 'localhost' and virt_type == "cloud-init" and cloud_init_used
 
 - name: "Create directories"
   file:
@@ -178,7 +178,7 @@
     - ["{{ res_def['name_separator'] }}"]
   loop_control:
     loop_var: definition
-  when: node_exists['failed'] is defined and uri_hostname == 'localhost' and virt_type == "cloud-init" and cloud_config_used
+  when: node_exists['failed'] is defined and uri_hostname == 'localhost' and virt_type == "cloud-init" and cloud_init_used
 
 - name: "Prepare cloud-config/user-data-local"
   template:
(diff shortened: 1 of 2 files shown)


linchpin/provision/roles/libvirt/tasks/provision_libvirt_node.yml
```diff
@@ -23,11 +23,11 @@
 - name: set cloud config default
   set_fact:
     cloud_config: "{{ res_def['cloud_config'] }}"
-    cloud_config_used: true
+    cloud_init_used: true
 
 - name: "check cloud_config is used or not"
   set_fact:
-    cloud_config_used: "{{ true if (cloud_config['users'] == [] and cloud_config['run_commands'] == []) else false }}"
+    cloud_init_used: "{{ true if (cloud_config['users'] != [] and cloud_config['run_commands'] == []) else false }}"
```
p3ck: This line still looks wrong to me.  Shouldn't you correct the check on run_commands and is and still the correct logic?  Wouldn't it be correct if either you had users or run_commands defined?

Has this passed your unit tests for cloud-init??

Why change the variable name when it has nothing to do with the logic fix?  it makes a one line patch much bigger than it needs to.

samvarankashyap: @p3ck 
There are two ways to one can use cloud_config in linchpin
cloud_config variable that contains virt-type as one of the parameters that determine whether we use cloud-init or virt-customize. It defaults to cloud-init. 
when using cloud-init does not have an option for run_commands since it is provided by virt-customize. 
the above condition checks if run_commands empty or not. 
Thinking of the above maybe we should check if run_commands parameter is defined or not.

I am afraid we do not have any unit tests for cloud-init but I tested it with libvirt-custom-xml target
from the libvirt workspace. 

cloud_config_used variable is used to only check for the case when is cloud-init is used. 
That's the reason I renamed  cloud_config used to cloud_init used.

14rcole: The logic seems to work, but there may be a more elegant way of solving the issue.  In order to determine if you  run cloud-init or virt-customize, do you just check whether `run_commands` exists?

I do see SK's point about changing the variable name.  It makes the PR longer but definitely makes the code easier to understand

johnbieren: "I am afraid we do not have any unit tests for cloud-init but I tested it with libvirt-custom-xml target
from the libvirt workspace." <- Is adding unit tests for it and CI testing for libvirt overall on the near term roadmap? I know of at least three teams that rely on libvirt in linchpin working and this latest release bit us

p3ck:
It's no wonder I'm confused.  Looking at the code there is virt_type which can be virt-customize or cloud-init, but then there is more logic in the latest version with cloud_config_used which is now called cloud_init_used.

johnbieren:
No, this patch does not work

johnbieren:
```
"msg": "AnsibleUndefinedVariable: 'cloud_config_used' is undefined"}
linchpin-workspace]# grep -r cloud_config_used ../
../lib/python3.6/site-packages/linchpin/provision/roles/libvirt/templates/libvirt_node.xml.j2:{% if virt_type == 'cloud-init' and cloud_config_used %}
```

linchpin/provision/roles/libvirt/tasks/provision_libvirt_node.yml
samvarankashyap: @johnbieren We will try to re-enable to ci for libvirt on one distro, 
however, I am not sure how easy or difficult the task would be we faced multiple (networking) problems with libvirt provisioning with existing ci. Which made us disable libvirt testing. 
I am currently, working on #1542 for libvirt that simulates one of the scenarios of libvirt. 
provisioning which might help partially with testing.

johnbieren:
ssh now works with the latest version of the patch

linchpin/provision/roles/libvirt/tasks/provision_libvirt_node.yml
johnbieren: I guess I don't follow. I have been using libvirt from linchpin for over a year now in my projects and with each linchpin release it either works exactly as I expect or doesn't work at all. There are no inconsistencies. What networking problems come up? I have not hit any networking problems ever using linchpin when I am on a working release

14rcole:
@samvarankashyap Can you rebase this against the latest version of develop?

samvarankashyap:
@14rcole rebased

samvarankashyap:
[test]

samvarankashyap:
[merge]

PR #1639, opened 2020-02-27
Fix validation return codes

### LINKED ISSUE #1578: LinchpinAPI do_validation is always returning rc 0 even though validation fails on layout
**Describe the bug**
When calling the LinchpinAPI `do_validation` function, it returns a tuple containing the return code, and the results of the validation. The` do_validation` always returns 'successful' because the rc is 0 but it's not until you run `do_action`, where it does another validation and throws a ValidationException error on the layout fields.

Looks like `do_validation` calls validator function `validate_pretty`, where as `do_action` is calling `validate`. In `validate_pretty` the only time the return code is changed to non-zero is if topology validation fails but not if `layout` or `cfgs` validation fails. `validate_pretty` should be returning non-zero for all schema key failures.

**To Reproduce**
Unfortunately this won't be reproduceable through the CLI because the shell `__ini__.py` doesn't even use the return code to detect failure, it uses the actual results data, and does a string check for certain key indicators.

1. Create a pinfile with improper layout schema
2. Open python console
3. Import `json`, `LinchpinAPI`, `LinchpinContext` modules
4. Use json module to load the pinfile data
5. Intialize a `LinchpinAPI` with a `LinchpinContext`
6. Provide the pinfile dict to the `LinchpinAPI.do_validation()`
7. Inspect that rc is 0 but results shows error in layout 

**Expected behavior**
If layout is specified in the pinfile and `do_validation` fails due to improper layout schema it should return a non-zero rc code.

Dannyb48:
Fixes #1578 where validate_pretty was not returning a non zero return code if the layout or cfg of the schema was failed to validate.

### DIFF
--- a/linchpin/validator/__init__.py
+++ b/linchpin/validator/__init__.py
@@ -113,6 +113,7 @@ class Validator(object):
             except SchemaError as e:
                 error = self._format_error(err_prefix, e)
                 results['layout'] = error + '\n'
+                return_code += 1
             else:
                 results['layout'] = "valid"
 
@@ -123,6 +124,7 @@ class Validator(object):
             except SchemaError as e:
                 error = self._format_error(err_prefix, e)
                 results['cfgs'] = error + '\n'
+                return_code += 1
             else:
                 results['cfgs'] = "valid"
 


14rcole:
@Dannyb48 Can you rebase this against the latest version of develop?

Dannyb48:
@14rcole done

samvarankashyap:
[merge]

PR #1641, opened 2020-02-27
Fix linchpin setup command

14rcole:
Fix bugs preventing linchpin setup from working properly.  These include not only bugs resulting from the move to Ansible Galaxy, but bugs that were present in linchpin setup already.  This also required some changes to the functionality of linchpin.

1. **The `linchpin setup` command no longer works with RHEL/CentOS 7**.  We removed support for yum.  However, commands such as `linchpin setup beaker` which only rely on pip will still work under RHEL/CentOS 7
2. Add support for a `--use-shell` flag.   This flag works exactly the same as the `--use-shell` flag in `linchpin up` and `linchpin destroy`.  In the future, this flag should be moved to the greater linchpin command.
2. The `--ask-sudo-pass` flag actually results in the user being asked for a sudo password.  Previously it simply included a task that linked libraries such as `libselinux-python` from the host environment to the virtual environment
3. The `--ask-sudo-pass` only works if `--use-shell` is set.  This is because of limitations caused by the way the playbook executor is called when linchpin does not call ansible through the shell
4. The `--use_venv` flag should be set if you are in a virtual environment.  It does two things.  The first is that it enables any linking tasks for libraries that need to exist in the virtual environment but can only be installed through pip.  The second is that it uses the `bin` folder of the virtual environment for ansible galaxy commands


14rcole:
I'll be adding a second commit to this that makes the setup for the containers use
```
linchpin setup <provider>
```
instead of 
```
pip install .[<provider>]
```
So that `linchpin setup` is actually tested in the future

linchpin/provision/roles/setup/tasks/docker_deps.yml
```diff
@@ -4,10 +4,15 @@
      current_user: "{{ lookup('env','USER') }}"
 
   - name: Gather facts
-    setup:
+     setup:
```
JunqiZhang0: ?? Should it be empty here? I tried to run 'linchpin setup' it failed here

linchpin/__init__.py
```diff
@@ -956,14 +956,14 @@ def _find_n_run_pb(self, pb_name, inv_src, console=True):
         self._get_role(pb_name)
```
JunqiZhang0: I guess this function 'get_role' should be changed? Because if pb_name is not the name of 'role' it would crash

JunqiZhang0: For example, the pb_name would be 'setup' if you run 'linchpin setup', it will raise an error

14rcole: setup is the name of the role, I changed it in this PR

14rcole:
[merge]

PR #1644, opened 2020-02-28
Make CentOS 7 integration tests run on Python 3

14rcole:


samvarankashyap:
@14rcole 
failed on  centos 7 installaion 
```
Traceback (ost recent call last):
  ​File "/usr/local/bin/linchpin", line 8, in <module>
    ​sys.xit(runcli())
  ​File "/usr/local/li/python3.6/site-packages/click/core.py", line 764, in __call__
    ​return self.mai(*args, **kwargs)
  ​Fil "/usr/local/lib/python3.6/site-packages/click/core.py", line 696, in main
    ​_verify_python3_env()
  ​ile "/usr/local/lib/python3.6/site-packages/click/_unicodefun.py", line 124, in _verify_python3_env
    ​' mitigation steps.' + extra
RuntimeError: Click will abort further execution because Python 3 was configured to use ASCII as encoding for the environment. Consult https://click.palletsprojects.com/en/7.x/python3/ for mitigation steps.

This system lists a couple of UTF-8 supporting locales that
you can pick from.  The following suitable locales were
discovered: en_U.utf8
```
add the following to your dockerfile 
```
ENV LC_ALL="en_US"
ENV LANG="en_US"
```

config/Dockerfiles/centos7/Dockerfile
```diff
@@ -3,25 +3,27 @@ LABEL maintainer "https://github.com/CentOS-PaaS-SIG/linchpin"
 LABEL description "This container will verify linchpin works under Centos7"
 
 ENV HOME=/root
+ENV LC_ALL=en_US.utf8
+ENV LANG=en_US.utf8
```
samvarankashyap: @14rcole For some reason it still fails. 
I faced the same error while setting up GitHub actions 
however i was able to fix it by setting variables as follows in shell script
```
VERSION_ID=$(cat /etc/*release | grep ^VERSION_ID | tr -d 'VERSION_ID="')

echo $VERSION_ID

if [ $VERSION_ID = "7" ]
then
    echo "This is centos7";
    export LC_ALL="en_US";
    export LANG="en_US";
fi
```
https://github.com/CentOS-PaaS-SIG/linchpin/blob/develop/.github/runmocktests.sh#L13

samvarankashyap:
@14rcole 
Two errors observed on centos7:
1. 
```
TASK [Write to a file] *********************************************************
fatal: [localhost]: FAILED! => {"changed": false, "checksum": "a91a42a661d9a7b88fb1e138574d226993b02813", "msg": "Aborting, target uses selinux but python bindings (libselinux-python) aren't installed!"}

localhost                  : ok=2    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0   
```
Avoid the libselinux by either using the write_to_file filter in linchpin or 
removing the write to file task inside the hook.

2. 
```
start hook ansible:ex_hook
[Errno 17] File exists: '/workDir/workspace/ci-linchpin/linchpin/docs/source/examples/workspaces/dummy-external-hook/hooks/ansible/ex_hook'
```
Avoid the above error by re-naming the external hook or 
deleting the folder, from localpath before running the hook.

14rcole:
@samvarankashyap Why would I delete the folder? It includes the playbook for the hook

samvarankashyap:
@14rcole 
Sorry, when I said "delete the folder" I meant to remove it via playbook task. 
since its an external hook would it be downloaded ? from the external link

14rcole:
You're right, I'll see if that helps.  However, I thought we had built in functionality to ignore that error.  Since users may run a pinfile multiple times

14rcole:
[test]

samvarankashyap:
@14rcole One more change for centos7 
https://github.com/CentOS-PaaS-SIG/linchpin/blob/develop/docs/source/examples/workspaces/dummy-multi-file-hook/hooks/ansible/ex_hook/ex2.yaml#L7 
Copy, template modules do not work on the centos7 python3 environment due to libselinux issues
it is better if we avoid it with filters. 
You can delete the above file or replace the copy module with something else. 
I see that is the only failure for this PR. 
Once, Updated we can merge it. 
Good work @14rcole

samvarankashyap:
@14rcole Alternatively, 
you could have just deleted this line 
https://github.com/CentOS-PaaS-SIG/linchpin/blob/develop/docs/source/examples/workspaces/dummy-multi-file-hook/PinFile#L12 
and deleted ex2.yml file.

14rcole:
[merge]

PR #1645, opened 2020-03-03
Add all OpenStack resource definitions to molecule converge tests

14rcole:


linchpin/provision/roles/openstack/tasks/provision_res_defs.yml
```diff
@@ -22,6 +22,6 @@
   include_tasks: "provision_{{ res_def['role'] }}.yml"
   when: res_def['role'] != 'os_server'
```
samvarankashyap: @14rcole Adding this when condition never runs os_server provisioning. please remove the when condition and comments following this task.

14rcole:
[merge]

PR #1661, opened 2020-03-05
Fix bug with aws ec2 mock & Breakdown GitHub actions into multiple steps 

samvarankashyap:
In order to easily verify the logs we have to breakdown GitHub actions into multiple steps 
Further, a couple of tests were failing for aws-ec2 due to a bug on ec2.py action_plugin 
and also the mock data was missing. 
The following PR fixes the above situation.

### DIFF
--- a/.github/run_aws_mocktests.sh
+++ b/.github/run_aws_mocktests.sh
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+echo "Hello this is first attempt to run linchpin provisioning in github actions"
+
+
+VERSION_ID=$(cat /etc/*release | grep ^VERSION_ID | tr -d 'VERSION_ID="')
+
+echo $VERSION_ID
+
+if [ $VERSION_ID = "7" ]
+then
+    echo "This is centos7";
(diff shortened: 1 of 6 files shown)


.github/run_aws_mocktests.sh
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env bash
```
14rcole: Why run the regular integration tests on GitHub actions?  Can we just leave that technical debt behind and move to the molecule tests?  We're still running the integration tests on GitHub for now so if something isn't caught by the molecule tests but is caught by the integration tests, we'll find out

samvarankashyap:
@14rcole I agree with previous comments, We need to have a working example running molecule tests. 
So that we can update all roles to molecule tests. 
The reason I created this PR is, 
there is a bug in aws mock tests which was not reporting failure caused due to zmq. 
Further, I wanted existing steps to be easily accessible instead of scrolling through 10k lines of test logs. 
It is not adding anything new. Just making existing things easily accessible.

samvarankashyap:
[merge]

PR #1671, opened 2020-03-09
Added azure mock tests to github actions

JunqiZhang0:
Added azure mock tests to GitHub actions


.github/run_azure_mocktests.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+echo "Hello this is first attempt to run linchpin provisioning in github actions"
```
14rcole: You can probably remove this line

PR #1695, opened 2020-03-17
Update AppImage building process

abraverm:


rpm/linchpin.json
```diff
@@ -12,7 +12,7 @@
         "license": "GPLv3+",
         "name": "linchpin",
         "summary": "Ansible based multicloud orchestrator",
-        "release": "2%{?dist}",
+        "release": "1",
```
JunqiZhang0: 1?

abraverm: linchpin.json is an input for rpmvenv tool, 'release' is required and the default is '1'. The 'release' means 'RPM' build version, it has no meaning for linchpin version itself. It is needed when there is a change in the RPM building processes (i.e the building script), for example for different OS versions (fedora vs centos, etc). The current building process is uniformal, but might change later so that's why I removed `{?dist}` and `2` didn't mean anything. I left `1` because I believe CI should be explicit as possible.

PR #1700, opened 2020-03-20
Enabled azure molecule tests

JunqiZhang0:
Enabled azure molecule tests to test ansible roles


.github/workflows/molecule-tests.yml
```diff
@@ -45,3 +46,13 @@ jobs:
           cd ./linchpin/provision/roles/openstack
           molecule converge -s delegated
           molecule destroy -s delegated
+      - name: "Execute Azure molecule tests"
```
14rcole: These same tests need to be run in `.github/workflows/molecule-tests-centos7.yml`

JunqiZhang0: Thanks!

JunqiZhang0:
[merge]

PR #1701, opened 2020-03-20
Add support for molecule converge tests to dummy and run tests on gthub actions

14rcole:


abraverm:
Where can I see the coverage results (percentage)?

14rcole:
@samvarankashyap PTAL

14rcole:
@abraverm If you go to one of the stages that says `multicontainer Testing on container matrix`, there's a step right after the unit tests that says `Report on code coverage`.  This will also fail if the coverage drops below 70%.  Hopefully we can increase that as we add more tests

14rcole:
The code coverage within github actions is 75%, but when I run it locally I get 70%.  I don't know why there's a difference

14rcole:
nvm, on github actions it was counting the unit tests as covered code.  That's fixed now

14rcole:
[merge]

PR #1710, opened 2020-03-27
Fix ansible hooks hanging

### LINKED ISSUE #1709: Carbon running ansible hooks hangs due to zmq action plugin
**Describe the bug**
Was testing out running post Ansible hooks with Carbon and I noticed that it would just hang after the playbook completed. I had to CTRL+Z and I would notice the following messages

`^CException KeyboardInterrupt in <bound method Context.__del__ of <zmq.sugar.context.Context object at 0x7f7054941710>> ignored`  

This has to do with the zmq plugin. I think a determined that even though I've set the `no_monitor` in the context in the API, it does not seem to be propagating to the ansible_hook_manager.

I have a potential fix for this that I will submitting that allows Carbon to be able to run ansible hooks and still be backwards compatible.

Dannyb48:
* Fix #1709 where no_monitor was not being propagated to the
  ansible action manager to disable the zmq multiprocess/progress bar code

* Fix an issue with the dummy action plugin causes hangs as well

### DIFF
--- a/linchpin/hooks/__init__.py
+++ b/linchpin/hooks/__init__.py
@@ -136,6 +136,8 @@ class LinchpinHooks(object):
         uhash = self.api.get_evar('uhash')
         uhash_enabled = self.api.get_evar('enable_uhash', False)
         ext = self.api.get_cfg('extensions', 'inventory')
+        # some applications need no_monitor disabled for ansible hooks
+        no_monitor = self.api.get_evar('no_monitor')
 
         inv_file = '{0}/{1}/{2}{3}'.format(workspace,
                                            inv_folder,
@@ -151,6 +153,7 @@ class LinchpinHooks(object):
         self.api.target_data['extra_vars'] = {}
         self.api.target_data['extra_vars']['inventory_dir'] = inv_folder
         self.api.target_data['extra_vars']['inventory_file'] = inv_file
+        self.api.target_data['extra_vars']['no_monitor'] = no_monitor
 
 
     def prepare_inv_params(self):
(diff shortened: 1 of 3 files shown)


Dannyb48:
Not sure why the appImage portion is failing. I did a side by side comparision of the workflow logs of 1.10 commit that successfully built the 1.10 appimage and this workflow log and nothing really sticks out to me. Other than I can see at the end of building the appimage it copies to the correct place on that volume fuse1 but I see two packages the 1.10 and the 2.0 appimage executables. 

```
 mv Linchpin-2.0.0.glibc2.17-x86_64.AppImage ../out/
+ ls -lh ../out/Linchpin-1.10.0.glibc2.17-x86_64.AppImage ../out/Linchpin-2.0.0.glibc2.17-x86_64.AppImage
-rwxr-xr-x 1 1001 115 64M Mar 18 15:50 ../out/Linchpin-1.10.0.glibc2.17-x86_64.AppImage
-rwxr-xr-x 1 1001 115 65M Mar 27 16:52 ../out/Linchpin-2.0.0.glibc2.17-x86_64.AppImage
```

And then it looks like it successfully copies these artifacts from the docker container to the local out directory
```
docker cp appimage-build-defae1a16cda942c:/workspace/out/ out

```

Only thing I can think of is that it did not successfully copy to `appimage/out` or that the runner location has changed?
`/home/runner/work/linchpin/linchpin/appimage/out/`

Is it actually?
`/home/runner/work/linchpin/appimage/out/`

Unfortunately, Alex is out for 6 weeks not sure if we can approve this without the appimage build being successful?

Dannyb48:
[merge]

PR #1712, opened 2020-03-30
Fix bug in which provisioning AWS fails when count is not set

### LINKED ISSUE #1711: AWS EC2 provisioning fails when `count` is not set
This bug was reported over google chat

14rcole:
Fixes #1711


14rcole:
[merge]

samvarankashyap:
[merge]

PR #1717, opened 2020-04-01
Fix bug preventing no_monitor from being set in linchpin.conf

### LINKED ISSUE #1716: no_monitor option cannot be set in linchpin.conf
When `no_monitor = True` is set in the linchpin.conf but not in the CLI, the flag is not set when ansible runs

14rcole:
Fixes #1716


samvarankashyap:
[merge]

PR #1757, opened 2020-09-22
hotfix release 2.0.4

JunqiZhang0:


Dannyb48:
@JunqiZhang0 So This flake8_polyfill is a radon [issue](https://github.com/rubik/radon/issues/199). Looks like when radon is installed they automatically register a flake [entry point](https://github.com/rubik/radon/blob/master/setup.py#L35) that gets utilized by flake8. Looks like radon 4.3.2 made a change where the flake8 dependency was moved to an extras_requirement so it's not getting installed by default. 

The options to resolve this fix might be to restrict the radon requirement in linchpin setup.py to `radon==4.2.0` or add an extra step in the [linting.yml](https://github.com/CentOS-PaaS-SIG/linchpin/blob/edcc91442779dedf465dcff2c81ad99409cc5b42/.github/workflows/linting.yml#L20) github action to pip install the required dependency i.e. `pip install radon['flake8']`

Hopefully one of those two actions should resolve this benign failure.

JunqiZhang0:
@Dannyb48 Will take a look, thanks!

JunqiZhang0:
need to trigger a hotfix release, should fix the CI later

PR #1758, opened 2020-10-02
fixed the ansible version

JunqiZhang0:


Dannyb48:
@JunqiZhang0 It looks like ansible 2.10 still has `InventoryManger`. I was able to import that in a local virtualenv. I think the flaw is in this check

https://github.com/CentOS-PaaS-SIG/linchpin/blob/develop/linchpin/ansible_runner.py#L15

If I run the same command to extract the ansible version I get `2.1` 
```
>>> ansible.__version__[0:3]
'2.1'
```
Hence why it's dropping into the else block and trying to import `Inventory` and NOT `InventoryManager`

I think a better way to get the version would be to do something like `ansible.__version__[:-2]`. This would be a better option rather than pinning the version to 2.9.13

JunqiZhang0:
@Dannyb48 Thanks for the information, I'll have a try with it.

JunqiZhang0:
Submitted another better fix, if that one #1759 gets merged, we can close this one

PR #1764, opened 2020-11-02
Increase wait retry for shutdown state

yprokule:
50 seconds for VM to shutdown not always enough so it's increased to 5 minutes.

Task `MSVMISD: Make sure vm is shutdown` is dropped cause:
* > state and command are mutually exclusive except when command=list_vms.
*  > there may be some lag for state requests like `shutdown' since these refer only to VM states.


JunqiZhang0:
Hi, @yprokule , I'm the new maintainer of Linchpin now, I'll review it soon and merge it

JunqiZhang0 (approved):
LGTM

